PR #19, opened 2021-09-01
Split up detection rules to prevent false positives for Github sawfish campaign

LeShadow:
This PR will essentially make sure that there must be a mention of the word github before it triggers the rule.


jkamdjou (approved):

jkamdjou:
thanks @LeShadow!

PR #40, opened 2021-11-02
Update sawfish from being too generic

jkamdjou:


ianthiel (approved):

detection-rules/impersonation_github_sawfish.yml
```diff
@@ -12,8 +12,8 @@ source: |
     or headers.return_path.domain.root_domain != 'github.com'
   )
   and 2 of (
-    iregex_search(body.plain.raw, '.*account activity.*', '.*github.*', '.*your activity.*', '.*suspicious api call.*'),
-    iregex_search(body.html.raw, '.*account activity.*', '.*github.*', '.*your activity.*', '.*suspicious api call.*'),
+    iregex_search(body.plain.raw, '.*account activity.*', '.*your activity.*', '.*suspicious api call.*'),
```
rw-access, line 15: btw, these are all additional ways to write the same thing:
```suggestion
    ilike(body.plain.raw, '*account activity*', '*your activity*', '*suspicious api call*'),
```

```suggestion
    iregex_search(body.plain.raw, 'account activity', 'your activity', 'suspicious api call'),
```

```suggestion
    iregex_match(body.plain.raw, '.*account activity.*', '.*your activity.*', '.*suspicious api call.*'),
```

rw-access (approved):

detection-rules/impersonation_github_sawfish.yml
```diff
@@ -12,8 +12,8 @@ source: |
     or headers.return_path.domain.root_domain != 'github.com'
   )
   and 2 of (
-    iregex_search(body.plain.raw, '.*account activity.*', '.*github.*', '.*your activity.*', '.*suspicious api call.*'),
```
rw-access, line 15: we could also add these to the "and 1 of ...` test below

PR #41, opened 2021-11-02
Update Ledger impersonation

jkamdjou:


detection-rules/impersonation_ledger.yml
```diff
@@ -7,21 +7,18 @@ type: "rule"
...
   and (
-    (
       sender.email.domain.root_domain == 'ledger.com'
       and headers.return_path.domain.root_domain not in ('ledger.com', 'amazonses.com')
-    ) or (
-      ilike(sender.email.email, '*ledger.com*')
-      and sender.email.domain.root_domain != 'ledger.com'
-    ) or (
+  )
+  or (
```
rw-access, line 13: is this `or` at the right level?
it means that everything below here doesn't need to match `type.inbound`

jkamdjou: you are absolutely right, will update

rw-access (approved):

PR #43, opened 2021-11-02
Use domain as FTS instead of email for custom domains

jkamdjou:


detection-rules/attachment_any_html_unsolicited.yml
```diff
@@ -9,6 +9,15 @@ type: "rule"
 source: |
   type.inbound
   and any(attachments, .file_extension in ('htm', 'html'))
-  and sender.email.email not in $recipient_emails
+  and (
+          (
+              sender.email.domain.root_domain in $free_email_providers
```
rw-access, line 14: confirming i understand the logic (may be worth commenting, but I don't know how often rules in here are commented)
```
if this comes from a free email provider:
  flag if i've never seen this email before

if this doesn't come from a free email provider:
  flag if i've never even seen this domain before
```
makes a lot of sense

rw-access (approved):

detection-rules/attachment_any_html_unsolicited.yml
jkamdjou: that's exactly right. added a comment!

rw-access (approved):

ianthiel (approved):

PR #57, opened 2021-11-24
Add binexplode rules

cameron-dunn-sublime:


ianthiel (approved):
Beautiful 💯

jkamdjou:
let's hold on the merge, there are some analysis API implications that we should think through

jkamdjou (approved):

PR #59, opened 2022-01-05
Create customer_service_phone_fraud_from_freemail detection rule

vector-sec:


detection-rules/customer_service_phone_fraud_from_freemail
```diff
@@ -0,0 +1,18 @@
+name: "Invoice or receipt from freemail sender with fraudulent customer service number"
+description: |
+  An invoice or receipt from a freemail sender which instructs the receipient to call a fraudulent customer service number
+type: "rule"
```
jkamdjou, line 4: if you'd like attribution when this becomes part of the rule feed, add the below here (we'll document this better in the future):

```suggestion
type: "rule"
author: "Your Name"
```

jkamdjou (commented):
Thank you for the awesome PR @vector-sec! Added one comment for attribution (if you'd like it) and one small suggestion

jkamdjou (approved):

PR #60, opened 2022-01-06
Fix `or` -> `and` in Github Sawfish rule

rw-access:
I'm not sure where this `or` is supposed to be. I stumbled on this rule and noticed that it looks incorrect.


jkamdjou:
this is a list of known malicious domains, so i think the intent is for it to flag under any condition

PR #63, opened 2022-01-15
Update FTS checks

jkamdjou:


ianthiel (approved):
Nice, should we make this update for all of the impersonation rules?

jkamdjou:
> Nice, should we make this update for all of the impersonation rules?

i think so yeah

PR #66, opened 2022-01-19
Add initial set of queries

jkamdjou:
Will update several queries once `distinct()` is merged, but we can get the ball rolling now


queries/authentication/dkim_any_fail.yml
```diff
@@ -0,0 +1,7 @@
+name: "DKIM: Fail"
+type: "query"
+source: |
+	any(headers.hops, ilike(.authentication_results.dkim, "*fail*"))
```
ianthiel, line 4: this might error on the leading tab here?

queries/attachments/links_free_file_host.yml
```diff
@@ -0,0 +1,12 @@
+name: "Attachment with links to free file sharing service"
+type: "query"
+source: |
+  filter(
+    map(attachments,
+        map(beta.binexplode(.),
+            map(filter(.scan.url.urls, .domain.root_domain in $free_file_hosts), .url)
```
ianthiel, line 7: it looks like we're extracting `.url` here, should this be pulling out `.file_name` instead?

queries/attachments/links_free_subdomain_host.yml
```diff
@@ -0,0 +1,12 @@
+name: "Attachment with links to free subdomain host"
+type: "query"
+source: |
+  filter(
+    map(attachments,
+        map(beta.binexplode(.),
+            map(filter(.scan.url.urls, .domain.root_domain in $free_subdomain_hosts), .url)
```
ianthiel, line 7: same here, should we be extracting `.file_name` ultimately?

queries/attachments/links_url_shortener.yml
```diff
@@ -0,0 +1,12 @@
+name: "Attachment with URL shorteners"
+type: "query"
+source: |
+  filter(
+    map(attachments,
+        map(beta.binexplode(.),
+            map(filter(.scan.url.urls, .domain.root_domain in $url_shorteners), .url)
```
ianthiel, line 7: Same here, the name suggests we should be pulling out `.file_name`

queries/attachments/auto_executing_macros.yml
```diff
@@ -0,0 +1,15 @@
+name: "Attachment with auto-executing VBA macros"
```
ianthiel, line 1: Nit: these should all be plural, so "Attachments with auto-executing VBA macros"

queries/headers/suspicious_recipient_pattern.yml
```diff
@@ -0,0 +1,11 @@
+name: "Suspicious Recipient pattern"
```
ianthiel, line 1: Let's consider updating this to be more descriptive until Descriptions are shown in the UI, e.g. "'To' matches 'From' but not 'Return-Path'"

ianthiel (approved):

queries/headers/all_recipients_bccd.yml
```diff
@@ -0,0 +1,8 @@
+name: "All recipients BCCd"
+type: "query"
+source: |
+  type.inbound
+  and length(recipients.to) == 0
+  and length(recipients.cc) == 0
+  and length(recipients.bcc) == 1
```
rw-access, line 7: `==` or `>=`?
or does it show up as a single "Undisclosed recipients"?

queries/headers/compauth_verdicts.yml
```diff
@@ -0,0 +1,8 @@
+name: "Compauth verdicts"
+type: "query"
+source: |
+  map(filter(headers.hops, .authentication_results.compauth.verdict is not null),
+    .authentication_results.compauth.verdict)
```
rw-access, line 5: slightly more compact, but up to you
```suggestion
  filter(map(headers.hops, .authentication_results.compauth.verdict, . is not null))
```

queries/headers/all_recipients_bccd.yml
jkamdjou: believe it's a single one

ianthiel (approved):

PR #71, opened 2022-01-21
Delete duplicate rule

jkamdjou:


ianthiel:
Should we delete [this one](https://github.com/sublime-security/sublime-rules/blob/master/detection-rules/file_sharing_link_suspicious_subject.yml) instead (assuming that's the dupe)? It appears to have some weaker FP checking.

jkamdjou:
> Should we delete [this one](https://github.com/sublime-security/sublime-rules/blob/master/detection-rules/file_sharing_link_suspicious_subject.yml) instead (assuming that's the dupe)? It appears to have some weaker FP checking.

i think deleting this one is ok, it's also more optimized

ianthiel (approved):

PR #86, opened 2022-02-16
Create headers_php_mailer_with_common_phishing_attachments.yml

vector-sec:


jkamdjou (approved):
We've been running this the past few days - this is an incredibly effective rule. It's caught hundreds of attacks with what looks to be 0 FPs so far. Nice work

PR #93, opened 2022-03-07
Add rule: ukr[.]net impersonation

jkamdjou:
Sandbox link: https://sandbox.sublimesecurity.com?id=7a916afe-e654-46f1-b65c-4a00c425a452

(updated)

### DIFF
--- a/detection-rules/impersonation_ukr_net.yml
+++ b/detection-rules/impersonation_ukr_net.yml
@@ -1,0 +1,48 @@
+name: "Brand impersonation: ukr[.]net"
+description: |
+  Impersonation of ukr[.]net.
+
+  Originally reported by CERT-UA on 07 March, 2022, phishing emails impersonate
+  ukr[.]net to steal user credentials. "Compromised mailboxes are used by the
+  Russian Federation's special services to conduct cyber attacks on citizens of Ukraine."
+references:
+  - "https://www.facebook.com/UACERT/posts/317482093744389"
+  - "https://www.facebook.com/UACERT/posts/317539153738683"
+type: "rule"
+source: |
+  type.inbound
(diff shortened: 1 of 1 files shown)


jkamdjou:
added sandbox link

detection-rules/impersonation_ukr_net.yml
```diff
@@ -0,0 +1,50 @@
...
+  - "https://www.facebook.com/UACERT/posts/317482093744389"
+  - "https://www.facebook.com/UACERT/posts/317539153738683"
+  - "https://twitter.com/thehackersnews/status/1500824885957857280?s=21"
+  - "https://thehackernews.com/2022/03/ukrainian-cert-warns-citizens-of.html"
+type: "rule"
+source: |
+  type.inbound
+  and ilike(sender.display_name, "ukr*net")
+  and sender.email.domain.root_domain != "ukr.net"
+  or (
```
rw-access, line 18: I think when `and` and `or` are at the indentation here, it makes it too easy to misread how the logic connects. I think explicit parentheses here would make it visually represent how things are grouped, and shows that this is intentional and not an accident. Same thing confused me back in https://github.com/sublime-security/sublime-rules/pull/60
```suggestion
  (
      type.inbound
      and ilike(sender.display_name, "ukr*net")
      and sender.email.domain.root_domain != "ukr.net"
  ) or (
```

to confirm that I'm understanding the rule right, it's two different pieces of logic:
* inbound with a ukr net name (more general purpose)
* or it matches a list of known IOCs.

jkamdjou: @rw-access updated!

ianthiel (approved):

PR #97, opened 2022-03-10
Use body.html.inner_text when possible (requires v0.24.5+)

rw-access:
Update rules to use the new `body.html.inner_text` field when they are searching for strings inside of HTML. Some guidance is on the docs [here](https://docs.sublimesecurity.com/docs/how-to-detect-keywords-or-phrases-in-the-body)

This field is new as of 0.24.5

### DIFF
--- a/queries/content/zero_width_spaces.yml
+++ b/queries/content/zero_width_spaces.yml
@@ -1,9 +1,9 @@
-name: "Zero-width spaces"
+name: "Zero-width spaces in a URL"
 references:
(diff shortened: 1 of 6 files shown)


queries/content/zero_width_spaces.yml
```diff
@@ -1,9 +1,9 @@
-name: "Zero-width spaces"
+name: "Zero-width spaces in a URL"
```
rw-access, line 1: the rule didn't quite seem to match what the link mentions which is more finely scoped. the original characters are actually pretty normal for HTML from my understanding

jkamdjou (approved):

PR #99, opened 2022-03-13
Added rule for RDP connection files

ajpc500:
This includes RDP connection files as attachments, or embedded in archive files.


jkamdjou (approved):
awesome work @ajpc500! solid

PR #100, opened 2022-03-17
Added excel web query file attachment rule

JamesCooteUK:


jkamdjou (approved):
🙌 thank you @JamesCooteUK!

PR #102, opened 2022-03-28
Rule: Potential Sandbox Evasion in Office File

ajpc500:
Could expand on this over time with other known sandbox checks, but thought it was a decent start!


jkamdjou (approved):
this is 🔥. looks great!

PR #105, opened 2022-04-06
Rule: Punycode Characters in Links

ajpc500:


detection-rules/link_contains_punycode_characters.yml
```diff
@@ -0,0 +1,15 @@
+name: "Link Contains Punycode Characters"
+description: |
+  The message contains Punycode characters to hide the true URL destination, or contains non-printable ASCII content.
```
rw-access, line 3: Since it's a domain thing:
```suggestion
name: "Link to a Domain with Punycode Characters"
description: |
  The body contains a link to a domain with Punycode characters to hide the true URL destination, or contains non-printable ASCII content.
```

ajpc500:
Nice - much clearer! 👍🏻

detection-rules/link_contains_punycode_characters.yml
```diff
@@ -0,0 +1,15 @@
...
+references:
+  - "https://www.bleepingcomputer.com/news/security/hackers-abuse-lookalike-domains-and-favicons-for-credit-card-theft/"
+type: "rule"
+authors:
+  - twitter: "ajpc500"
+source: |
+  type.inbound
+  and any(body.links,  
+    regex_search(.href_url.domain.domain, '\Axn--', '[^\x00-\x7F]')
+  )
```
jkamdjou, line 13:
```suggestion
  ) 
  and (
    // include automated emails
    sender.display_name == "WordPress"
    or sender.email.local_part == "wordpress"

    // Has my organization ever sent an email to this sender?
    or (
        sender.email.domain.root_domain in $free_email_providers
        and sender.email.email not in $recipient_emails
    )
    or (
        sender.email.domain.root_domain not in $free_email_providers
        and sender.email.domain.domain not in $recipient_domains
    )
  )
```

jkamdjou (commented):
we're seeing some FPs over the last 30 days, what do you think of this FP reduction logic? primarily from [common snippets](https://docs.sublimesecurity.com/docs/common-snippets)

in our testing, that looks to work pretty well to reduce the noise and still hit all the attacks, but curious how well it works for you

detection-rules/link_contains_punycode_characters.yml
ajpc500: Yep, happy with that! We've seen a few too, largely from other regions e.g. Japan. This would filter those and still hit the TPs we've seen 👍🏻

jkamdjou: awesome 🤘

jkamdjou (approved):
thank you for this 🙌 :shipit:

PR #107, opened 2022-04-15
Fix duplicate attachment/binexplode loops

rw-access:
Found duplicate loops in the "Please Enable Macros" rule that can be merged together.


ianthiel (approved):

ianthiel:
love me a good red diff 👏

PR #119, opened 2022-05-05
Update description to accurately reflect file types scanned

jkamdjou:
@vector-sec pointed out in https://github.com/sublime-security/sublime-rules/pull/118#discussion_r865991391 that our description here was inaccurate.


ianthiel (approved):

jkamdjou:
Closing this out, and instead we'll add archives to this as in #118.

PR #121, opened 2022-05-06
Expand rule to include archives to match description

jkamdjou:
h/t @vector-sec for pointing this out in #118


ianthiel (approved):
nice 👍

PR #123, opened 2022-05-10
Rename html smuggling "location" file name rule to be consistent with other html smuggling rules

jkamdjou:
- If the user doesn't have this rule installed, it will show up as 1 new rule added to the feed
- If the user has this rule installed, it will show up as 1 rule deleted and 1 added


rw-access:
This is the URI that I see in the platform:
> https://platform/feeds/0214465b-004a-4952-8469-4bd705c52f8e/rules/f541a262-97b6-5b87-aa65-c3210b459833

If you add `id: 'f541a262-97b6-5b87-aa65-c3210b459833'` to the YAML, it'll recognize it as the same rule before and after the change. Otherwise, it'll be the delete+new like you mentioned to to people who have installed the rule, like you mentioned.

rw-access (approved):

PR #124, opened 2022-05-13
Update attachment_filename_with_unicode_rtlo.yml to scan compressed files

vector-sec:


detection-rules/attachment_filename_with_unicode_rtlo.yml
```diff
@@ -1,15 +1,23 @@
...
+  and (
+      any(attachments, iregex_search(.file_name, '\x{202E}', '\x{202D}'))
+      or (
+          any(attachments, .file_extension in~ ("7z","bz2","gz","rar","tar","zip","zipx","iso","img")
+              and any(beta.binexplode(.),
+                  iregex_search(.name, '\x{202E}', '\x{202D}')
+              )
+          )
+      )
+  )
```
rw-access, line 21: you can put these two together. not necessarily for performance, but I tend to think of it more like this:
```suggestion
  and any(attachments,
          iregex_search(.file_name, '\x{202E}', '\x{202D}') or (
              .file_extension in~ ("7z","bz2","gz","rar","tar","zip","zipx","iso",  "img")
              and any(beta.binexplode(.),
                  iregex_search(.name, '\x{202E}', '\x{202D}')
              )
          )
      )
```

jkamdjou (approved):
awesome, great update @vector-sec

PR #125, opened 2022-05-17
New Rule - Attachment: PowerShell Content

ajpc500:


detection-rules/attachment_powershell_content.yml
```diff
@@ -0,0 +1,26 @@
...
+source: |
+  type.inbound and
+  any(attachments,
+      .file_extension in~ (
+          // PowerShell related file extensions
+          "ps1", "ps1xml", "psm1", "psd1", "pssc", "psrc", "cdxml", "ps2", "ps2xml", "psc2",
+          // Archive file formats 
+          "7z","bz2","gz","rar","tar","zip","zipx","iso","img"
+      )
+      and any(beta.binexplode(.),.extension in
```
jkamdjou, line 21:
```suggestion
      and any(beta.binexplode(.), .extension in~
```

we might want to make this case insensitive too

jkamdjou (approved):
🔥

we could take this further in another rule for other commonly abused file types too - .hta, etc.

detection-rules/attachment_powershell_content.yml
ajpc500: Ah yeah, good spot 👍🏻

PR #126, opened 2022-05-25
New Rule - Attachment: PDF file with embedded content

ajpc500:
Hi guys, keen to hear your thoughts on this one.

Rule has been performant so far with just a `type.inbound` but curious on what you see and whether it warrants an 'unsolicited' filter.

I've also added additional file formats (beyond solely doc-based ones) into the embedded file's `flavor` list, but could refine it to exclusively tackle the specific actor's implementation included in the bleepingcomputer link referenced in the rule!

lmk what you think!


jkamdjou (approved):
thanks a lot for this @ajpc500!

> Rule has been performant so far with just a `type.inbound` but curious on what you see and whether it warrants an 'unsolicited' filter.

we're seeing some Hunts that don't complete in large environments because it returns too many files. this means that those users won't see auto-backtest results in the Rule Feed for the L24H. would you be opposed to adding the Unsolicited filter? that seems to do the trick

> I've also added additional file formats (beyond solely doc-based ones) into the embedded file's `flavor` list

love the additional file formats!

ajpc500:
perfect, updated with the `unsolicited sender` filters. Congrats on 100+ open source rules too btw! 🎉👏🏻

ajpc500:
Absolutely, updated rule to remove `mz_file`

PR #128, opened 2022-05-28
HTML smuggling updates: add .file_type check and improve formatting

jkamdjou:
h/t @vector-sec for pointing out that some attackers don't set a file extension. we can flag these by using the improve .file_type property, which uses file headers to determine the file type.


ianthiel (approved):
👍

jkamdjou:
@rw-access what's the correct `id` to use to prevent the file rename of `detection-rules/attachment_html_smuggling_unescape.yml` from showing up as a delete + new change in the Feed?

is it `0b0fed36-735a-50f1-bf10-6673237a4623`?

PR #131, opened 2022-05-31
Update remote template detection: Detect .htm files in addition to .html

jkamdjou:
h/t @vector-sec for pointing out that .htm files could potentially be used as a bypass

### DIFF
--- a/detection-rules/office_remote_html_template.yml
+++ b/detection-rules/office_remote_html_template.yml
@@ -25,7 +25,7 @@ source: |
               // detect files without an extension
               or .file_type in ("docx")
           )
-          and any(beta.oletools(.).relationships, iregex_search(.target, ".*html!"))
+          and any(beta.oletools(.).relationships, iregex_search(.target, ".*html?!"))
       )
       or (
           (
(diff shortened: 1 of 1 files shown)


detection-rules/office_remote_html_template.yml
```diff
@@ -25,7 +25,7 @@ source: |
               // detect files without an extension
               or .file_type in ("docx")
           )
-          and any(beta.oletools(.).relationships, iregex_search(.target, ".*html!"))
+          and any(beta.oletools(.).relationships, iregex_search(.target, ".*html?!"))
```
rw-access, line 28: no need for leading `.*` when doing a search.  doesn't matter for performance because MQL should strip them out under the hood.
```suggestion
          and any(beta.oletools(.).relationships, iregex_search(.target, "html?!"))
```

detection-rules/office_remote_html_template.yml
```diff
@@ -36,7 +36,7 @@ source: |
           )
           and any(beta.binexplode(.),
               .flavors.mime == "text/xml" and
-              any(.scan.strings.strings, iregex_search(., ".*oleObject.*html!.*"))
+              any(.scan.strings.strings, iregex_search(., ".*oleObject.*html?!.*"))
```
rw-access, line 39:
```suggestion
              any(.scan.strings.strings, iregex_search(., "oleObject.*html?!"))
```

rw-access (approved):

detection-rules/office_remote_html_template.yml
jkamdjou: ty!

PR #135, opened 2022-06-10
New rule - Attachment: HTML smuggling with atob and high entropy

jkamdjou:
New coverage, 0 FPs so far


ianthiel (approved):

detection-rules/attachment_html_smuggling_atob.yml
```diff
@@ -0,0 +1,24 @@
+name: "Attachment: HTML smuggling with atob and high entropy"
+description: |
+  Recursively scans files and archives to detect HTML smuggling techniques.
+type: "rule"
+source: |
+  type.inbound
+  and any(attachments, .size <= 15000 and
```
rw-access, line 7: Is this `.size` supposed to apply to all the top-level files or the exploded files?

PR #136, opened 2022-06-10
New rule - Attachment: HTML smuggling with setTimeout

jkamdjou:
New coverage, 0 FPs so far

### DIFF
--- a/detection-rules/attachment_html_smuggling_settimeout.yml
+++ b/detection-rules/attachment_html_smuggling_settimeout.yml
@@ -1,0 +1,21 @@
+name: "Attachment: HTML smuggling with setTimeout"
+description: |
+  Recursively scans files and archives to detect HTML smuggling techniques.
+references:
+  - "https://www.microsoft.com/security/blog/2021/11/11/html-smuggling-surges-highly-evasive-loader-technique-increasingly-used-in-banking-malware-targeted-attacks/"
+type: "rule"
+source: |
+  type.inbound
+  and any(attachments, .size <= 400 and
+      (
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_html_smuggling_settimeout.yml
```diff
@@ -0,0 +1,21 @@
...
+source: |
+  type.inbound
+  and any(attachments, .size <= 400 and
+      (
+          .file_extension in~ ("html", "htm", "7z","bz2","gz","rar","tar","zip","zipx","iso","img") or
+          .file_type == "html"
+      )
+      and any(beta.binexplode(.),
+          "setTimeout" in .scan.javascript.identifiers
+          and any(.scan.strings.strings, ilike(., "*location.href*"))
```
rw-access, line 16: it looks like this doesn't detect some examples in the linked reference. I also don't see any mention of setTimeout on the page. I guess that means the link is loosely related?

jkamdjou: good catch. this link actually has nothing to do with the rule, and was just a copy/paste from other HTML smuggling rules. will remove. or we could keep it if it's helpful to have _something_ about HTML smuggling

jkamdjou: removed from this PR and #135

rw-access: ah yeah. I also like having a link to _something_  like you said. your call on which approach to take. could also add a yaml comment about why the link makes sense

rw-access (approved):

PR #138, opened 2022-06-14
New rule - Sendgrid onmicrosoft.com domain phishing

ajpc500:
Seen a few campaigns doing this in the past couple of months with no FPs, keen to hear what you see!


jkamdjou (approved):
🔥 really solid. same results over here! 🙌

PR #144, opened 2022-06-29
New rule - Disallowed file type within an archive attachment

jkamdjou:


ianthiel (approved):
lgtm, possible alt name: "Attachment: Archive containing disallowed file type"

PR #146, opened 2022-07-01
Adding additional header indicators

arnydo:
Adding additional indicators observed in phishing emails utilizing Sendgrid with '"Voicemail" subjects, etc.


detection-rules/headers_voicemail_sendgrid.yml
```diff
@@ -5,8 +5,12 @@ reference:
...
 type: "rule"
 source: |
-  type.inbound
-  and headers.return_path.domain.domain == 'sendgrid.net'
-  and ilike(subject.subject, '*voicemail*', '*voice message*')
+  type.inbound 
+  and (headers.return_path.domain.domain == 'sendgrid.net'
+    or any(headers.domains, ilike(.domain, "*.outbound-mail.sendgrid.net"))
+    or any(headers.hops, any(.fields, ilike(.name, "Received") and ilike(.value, "*.outbound-mail.sendgrid.net*")))
+    or any(headers.hops, any(.authentication_results.dkim_details, .domain == 'sendgrid.net'))
```
rw-access, line 12: @jkamdjou are these redundant with the `headers.domains` check? I believe that both these locations are copied into `headers.domains`, which is also much faster. If `headers.domains` isn't too permissive, then we would only need to do `any(headers.domains, ilike(.domain, "*.outbound-mail.sendgrid.net"))`.

jkamdjou: `headers.domains` aggregates the domains found in the [`Received` headers only](https://docs.sublimesecurity.com/docs/mdm), which wouldn't include authentication results. but i do think `headers.domains` should probably be sufficient on its own, unless @arnydo is seeing some instances where one is present and not the other

image

jkamdjou (commented):
thanks for this @arnydo, and apologies for the delay!

we've done some testing on our side and we're seeing quite a few false positives in larger environments unfortunately. it turns out a lot of legitimate voicemail services use Sendgrid. they mostly all have custom domains verified so the return_path is not sendgrid, unlike the phishing attacks we've seen that just use the stock onmicrosoft domain.

i think it might be best to limit this change to the addition of keywords here:

`and ilike(subject.subject, '*voicemail*', '*voice message*', 'Voice*')`

and for other attacks that don't use onmicrosoft but still abuse sendgrid, we could build more tailored detections. how does that sound? if you're able to share any of those attacks that don't match this rule, we'd be happy to help build for those and run more tests on our side!

detection-rules/headers_voicemail_sendgrid.yml
arnydo: I am pretty sure it wasn't firing until I added the additional check. Let me verify again and let you know.

jkamdjou:
After discussion with @arnydo, we're going to close this one out for now and re-visit if we see some fresh samples.

PR #148, opened 2022-07-14
Added a simple rule to detect eml attachments

Und3rf10w:
Added a simple rule that grabs any attachment with the extension `eml` and sends it to binexplode. Perhaps there should be an `or` condition instead of an `and` condition for matches in the `binexplode` on line 11. Another `or` statement to find instances using the yara flavor may be beneficial as well.

I couldn't pick a tag that seemed relevant, nor is there an "info" severity, so I omitted both of those.


jkamdjou (commented):
thanks for the PR @Und3rf10w!

> nor is there an "info" severity

yeah, originally we had an "info" severity, but we decided to remove it because the rules that fit this category weren't necessarily _detection_ rules but for other purposes like discovery (or in this case automation).

i think it might make sense for us to create a new folder in this repo for rules like this, such as `discovery-rules`. and then i think just this would suffice:

```
  type.inbound
  and any(attachments,
    // ideally this would be off of a magic number, but that doesn't seem to exist for eml files
    .file_extension in~ ('eml')
  )
```

how does that sound?

Und3rf10w:
Sure, I'll revise accordingly

Und3rf10w:
On this note, why is it recommended to exclude the `binexplode` function? Wouldn't a user be interested in eml files even if they're inside attached zip(s) for example? To my understanding the `binexplode` function would capture that.

jkamdjou:
> On this note, why is it recommended to exclude the `binexplode` function? Wouldn't a user be interested in eml files even if they're inside attached zip(s) for example? To my understanding the `binexplode` function would capture that.

We'd just want to be targeted for that use case, because bin-exploding _all_ files won't be performant at scale (both for live detection rules and Hunt). If we wanted to look for users reporting EMLs, looking directly attached to the email should be sufficient. If we wanted EMLs inside archives for a different use case, we could include a file type check for zips, tars, etc.

rw-access:
And you can always add another rule for the deeper inspection, searching for `.eml` files!

discovery-rules/attachment_any_eml.yml
```diff
@@ -0,0 +1,12 @@
+name: "Attachment: Any eml file"
```
rw-access, line 1:
```suggestion
name: "Attachment: Any EML file"
```

rw-access (approved):

discovery-rules/attachment_any_eml.yml
```diff
@@ -0,0 +1,12 @@
...
+description: |
+  Potential EML file in message. This rule may potentially be combined with a webhook 
+  notification to submit the EML contents to the analysis API
```
jkamdjou, line 4:
```suggestion
  Any EML attachment. This rule can be combined with a webhook action for further analysis
  of attached EML files, eg via the analysis API.
```

jkamdjou: small nit: notification -> action, and slight verbage change since you'll technically have to poll instead of sending the contents directly

jkamdjou (approved):

discovery-rules/attachment_any_eml.yml
Und3rf10w: Changed

discovery-rules/attachment_any_eml.yml
Und3rf10w: Changed

PR #149, opened 2022-07-18
Add Test rule

jkamdjou:
Instead of auto-creating the "Test rule" when setting up your first message source like we do today, it can be auto-installed and auto-activated from the Feed. I think it would be cooler to see a Feed Rule in your Dashboard when you first sign in, opposed to just a regular one.

We'd need to standardize Feed rule IDs across all deployments for that to work.

There is an argument for not having "test rules" in the feed at all so I am a bit torn tbh, lmk if you feel strongly either way @rw-access


rw-access:
I'm leaning towards keeping it out of the feed. One less dependency to worry about, and I think once you delete the rule, you don't need to keep seeing it show up as available in the feed.

PR #153, opened 2022-07-27
New rule: Lookalike sender domain, remove template

jkamdjou:


detection-rules/lookalike_sender_domain.yml
```diff
@@ -0,0 +1,38 @@
...
+      or (
+          sender.email.domain.root_domain not in $free_email_providers
+          and sender.email.domain.domain not in $sender_domains
+      )
+  )
+
+  // unsolicited
+  and (
+      (
+          sender.email.domain.root_domain in $free_email_providers
```
rw-access, line 28: looks like these two things can be merged together

detection-rules/lookalike_sender_domain.yml
```diff
@@ -0,0 +1,38 @@
+name: "Lookalike sender domain"
+description: |
+  Sender's domain is a lookalike of one of your organization's domains.
+type: "rule"
+source: |
+  type.inbound
+
+  and any($org_domains,
+      // this rule can be duplicated in order to tune the threshold
+      iedit_distance(sender.email.domain.domain, .) < 2
```
rw-access, line 10: safe as long as org_domains is fairly small

detection-rules/lookalike_sender_domain.yml
jkamdjou: damn you're right, lots of rules to update later

detection-rules/lookalike_sender_domain.yml
rw-access: also @jkamdjou do you mean to do this?
```suggestion
      iedit_distance(sender.email.domain.domain, .) == 1
```

this way, you handle the case where it's exactly one edit. `< 2` handles includes the `== 0` and the `== 1` case, so you would get both lookalike and exact matches.

rw-access: > is there a way for us to exclude the `== 0` case if we left it at that?

not exactly sure what you mean. `<= 1` or `< 2` means exact match or one edit off. `== 1` means exactly one edit off

jkamdjou: sorry i meant <=2. for now i'm thinking let's just roll with ==1 for expediency

rw-access: OH!. in that case, yeah it can be `<= 2` if that makes more sense

rw-access (approved):

PR #154, opened 2022-07-28
Remove PHP mailer rule + add rule exception

jkamdjou:


detection-rules/impersonation_github.yml
```diff
@@ -11,7 +11,7 @@ source: |
       or ilike(sender.email.email, '*github*')
       or iedit_distance(sender.email.domain.sld, 'github') <= 2
   )
-  and sender.email.domain.root_domain not in ('github.com', 'gitlab.com', 'itthub.net')
+  and sender.email.domain.root_domain not in ('github.com', 'gitlab.com', 'itthub.net', 'githubsupport.com')
```
rw-access, line 14: how confident are you in this? I don't get any results from a domain lookup

rw-access: hmmm, looks like it has MX and TXT records though?
https://www.whatsmydns.net/dns-lookup?query=githubsupport.com&server=google

jkamdjou: i've communicated with them before

image

rw-access: okay, it's zendesk
https://webcache.googleusercontent.com/search?q=cache:QeIMJ5eQ5BIJ:https://github.community/t/emails-from-support-githubsupport-com/171928+&cd=1&hl=en&ct=clnk&gl=us

https://news.ycombinator.com/item?id=23703453

rw-access (approved):

PR #157, opened 2022-08-03
Added amazonpay domains for US and India

indigocarmen:


jkamdjou (approved):
thanks @indigocarmen!

PR #161, opened 2022-08-19
Update Github impersonation: Add exception and reduce distance threshold

jkamdjou:
h/t @indigocarmen

### DIFF
--- a/detection-rules/impersonation_github.yml
+++ b/detection-rules/impersonation_github.yml
@@ -9,9 +9,9 @@ source: |
   and (
       ilike(sender.display_name, '*github*')
       or ilike(sender.email.email, '*github*')
-      or iedit_distance(sender.email.domain.sld, 'github') <= 2
+      or iedit_distance(sender.email.domain.sld, 'github') == 1
   )
-  and sender.email.domain.root_domain not in ('github.com', 'gitlab.com', 'itthub.net', 'githubsupport.com')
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_github.yml
```diff
@@ -9,9 +9,9 @@ source: |
   and (
       ilike(sender.display_name, '*github*')
       or ilike(sender.email.email, '*github*')
-      or iedit_distance(sender.email.domain.sld, 'github') <= 2
+      or iedit_distance(sender.email.domain.sld, 'github') == 1
```
rw-access, line 12: `<= 1` isn't a bad idea, since it will also handle the case of different TLDs like `github.net`.
could go either way

jkamdjou: updated

detection-rules/impersonation_github.yml
```diff
@@ -9,9 +9,9 @@ source: |
   and (
       ilike(sender.display_name, '*github*')
       or ilike(sender.email.email, '*github*')
-      or iedit_distance(sender.email.domain.sld, 'github') <= 2
+      or iedit_distance(sender.email.domain.sld, 'github') <= 1
   )
-  and sender.email.domain.root_domain not in ('github.com', 'gitlab.com', 'itthub.net', 'githubsupport.com')
+  and sender.email.domain.root_domain not in ('github.com', 'gitlab.com', 'itthub.net', 'githubsupport.com', 'gtmhub.com')
```
rw-access, line 14: I don't think this will ever hit because of the earlier thing, but it doesn't hurt

rw-access (approved):

PR #171, opened 2022-09-12
Replace tabs with spaces in attachment_suspicious_macro.yml

hugh-sublime:
## Context

Tabs are invalid in MQL and cause the CI to fail. Simply replaced the tabs with spaces.


rw-access (approved):
LGTM!

It took me a second to realize how this change wasn't caught earlier. Looks like it was
- we weren't up to date with `main` in the CI PR
- #168 fixed a detection rule, but not a discovery rule. that tripped me up for a second. I thought this was a dupe and the other one somehow got reverted

PR #172, opened 2022-09-12
Apply rule validation to all PR types

hugh-sublime:
## Context

Instead of just running the `Rules PR CI` on just branches, we should also run it on every PR as well. Added CI triggers for `pull_request`.

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -2,7 +2,9 @@ name: Rules PR CI
 
 on:
   push:
-    branches: [ "**" ]
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "*" ]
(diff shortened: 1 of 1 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -2,7 +2,9 @@ name: Rules PR CI
 
 on:
   push:
-    branches: [ "**" ]
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "*" ]
```
rw-access, line 7: strangely, you need `**` to match branches with a `/` in them. it's a glob thing, I guess
```suggestion
    branches: [ "**" ]
```

rw-access (approved):
LGTM.
just one nit for an edge case we'll probably never run into

PR #175, opened 2022-09-24
Adding another Spotify domain to their impersonation rule

aidenmitchell:
Ordered a Car Thing from Spotify and the order confirmation came from no-reply@fromspotify.com


ianthiel (approved):
nice 👍

PR #178, opened 2022-10-26
Add severity to all rules

rw-access:


jkamdjou (commented):
Many of these rules already have a `severity` field, so there's 2 now :/

jkamdjou:
> Many of these rules already have a `severity` field, so there's 2 now :/

i can make the updates in like 30 minutes

jkamdjou:
> Many of these rules already have a `severity` field, so there's 2 now :/

i can make the updates in like 45 minutes

jkamdjou (approved):

ianthiel (approved):
🔥

PR #181, opened 2022-11-14
Use new lists: $file_extensions_common_archives and $file_extensions_macros

jkamdjou:
Also in this PR:

- Updated some legacy Unsolicited snippets
- Updated the RDP file rule severity to medium (h/t @ajpc500)
- Renamed 3 files for consistency (below for reference): @rw-access should we add the feed ID so they don't show up as deleted/new rules when viewing the Feed? That would also keep historically flagged data on these rules in tact. If yes, I could use a reminder on how to do that
- Small FP update

```
 rename detection-rules/{attachment_any_html_suspicious_javascript_identifiers.yml => attachment_html_smuggling_rot13.yml}
 rename detection-rules/{attachment_html_smuggling.yml => attachment_html_smuggling_unescape.yml}
 rename detection-rules/{office_remote_html_template.yml => attachment_office_remote_html_template.yml}
```

### DIFF
--- a/detection-rules/impersonation_wells_fargo.yml
+++ b/detection-rules/impersonation_wells_fargo.yml
@@ -9,7 +9,7 @@ severity: "high"
 source: |
   type.inbound
   and (
-      strings.ilike(sender.display_name, '*wells fargo*')
+      sender.display_name =~ 'wells fargo'
       or strings.ilevenshtein(sender.display_name, 'wells fargo') <= 1
       or strings.ilike(sender.email.domain.domain, '*wellsfargo*')
       or strings.ilike(subject.subject, '*wells fargo security*')
--- a/detection-rules/attachment_any_html_in_archive_unsolicited.yml
+++ b/detection-rules/attachment_any_html_in_archive_unsolicited.yml
@@ -10,7 +10,7 @@ severity: "medium"
 source: |
   type.inbound
   and any(attachments,
-      .file_extension in~ ( "7z", "bz2", "gz", "rar", "tar", "zip", "zipx", "iso", "img")
+      .file_extension in~ $file_extensions_common_archives
       and any(beta.binexplode(.), .depth > 0
           and .extension in~ ("html", "htm"))
   )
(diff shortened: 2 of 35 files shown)


ianthiel (approved):
beautiful 👍

rw-access:
to preserve the rule ID, you need to manually specify the ID in the YAML so that it matches whatever it was before. it's generated deterministically from the hash of the path.

also keep in mind that lists only are synced when bora boots and also nightly. since we don't have many guarantees about bora booting I would wait until tomorrow to merge

jkamdjou:
@rw-access adding rule IDs worked perfectly, confirmed the 3 rules show up as `Updated` instead of `Deleted`+`New` in a test branch

detection-rules/impersonation_wells_fargo.yml
```diff
@@ -9,7 +9,7 @@ severity: "high"
 source: |
   type.inbound
   and (
-      strings.ilike(sender.display_name, '*wells fargo*')
+      sender.display_name =~ 'wells fargo'
```
rw-access, line 12: no substrings anymore? sounds fine to me, otherwise you may want `strings.icontains`

jkamdjou: yeah no substring anymore, too many FPs

rw-access (approved):

aidenmitchell:
I believe this query will also need to be updated with the $file_extensions_macros list.
https://github.com/sublime-security/sublime-rules/blob/ed14fdf1208707fa1ecd47bce9a1eec6f0e5843a/queries/attachments/vba_macros.yml#L4

jkamdjou:
> I believe this query will also need to be updated with the $file_extensions_macros list.
> 
> https://github.com/sublime-security/sublime-rules/blob/ed14fdf1208707fa1ecd47bce9a1eec6f0e5843a/queries/attachments/vba_macros.yml#L4

NICE @aidenmitchell thank you

PR #182, opened 2022-11-17
Add embedded base64-encoded executable rule

aidenmitchell:


detection-rules/attachment_any_embedded_b64_pe
```diff
@@ -0,0 +1,13 @@
...
+description: |
+  Attachment contains a base-64 encoded executable.
+type: "rule"
+severity: "high"
+  type.inbound 
+  and any(attachments,
+    any(beta.binexplode(.), 
+      any(.flavors.yara, . == 'base64_pe')) 
+  )
+  and sender.email.email not in $recipient_emails
```
jkamdjou, line 11: based on how great the performance of this rule is in some Hunts we've run, i think we can actually do without any historical context. i have yet to see any FPs! if that changes, i'll let you know

also check out [this unsolicited snippet](https://docs.sublimesecurity.com/docs/common-snippets#unsolicited-has-my-organization-ever-sent-an-email-to-this-sender), which is a bit more flexible

detection-rules/attachment_any_embedded_b64_pe
```diff
@@ -0,0 +1,13 @@
...
+tags:
+  - "Suspicious attachment"
```
jkamdjou, line 13:
```suggestion
  - "Suspicious attachment"
  - "HTML smuggling"
```

jkamdjou (commented):
🙌  thanks @aidenmitchell! some quick feedback and thoughts:

- we'll need a `.yml` extension on the file for it to be picked up in the Rules Feed
- thoughts on calling this rule `Attachment: HTML smuggling with embedded base64-encoded executable` w/ file name `attachment_html_smuggling_embedded_b64_pe.yml` ? since pretty much the only use of b64 encoding a PE in an HTML file i can think of is HTML smuggling, i think we can be explicit

aidenmitchell:
I'm not sure what it's unhappy about, I believe the yaml is valid.

detection-rules/attachment_html_smuggling_embedded_b64_pe.yml
```diff
@@ -0,0 +1,28 @@
+name: "Attachment: HTML smuggling with embedded base64-encoded executable"
+description: |
+  HTML attachmemt contains a base-64 encoded executable.
+type: "rule"
+severity: "high"
+  type.inbound
```
jkamdjou, line 6:
```suggestion
source: |
  type.inbound 
```

detection-rules/attachment_html_smuggling_embedded_b64_pe.yml
```diff
@@ -0,0 +1,28 @@
...
+  and (
+          (
+              sender.email.domain.root_domain in $free_email_providers
+              and sender.email.email not in $recipient_emails
+          )
+          or (
+              sender.email.domain.root_domain not in $free_email_providers
+              and sender.email.domain.domain not in $recipient_domains
+          )
+  )
```
jkamdjou, line 25: sorry my last comment was confusing! i don't think we need this snippet for this particular rule based on how great it performed in our Hunts. but if we did, this snippet in general would be better than the original `and sender.email.email not in $recipient_emails`

aidenmitchell: ah ok, no worries! Do we want to go back to what was there, or just keep this one in?

jkamdjou: i think we can delete this block

jkamdjou: this is what i was thinking [cc8929d](https://github.com/sublime-security/sublime-rules/pull/182/commits/cc8929dd9309ad463e0e415c13f4324b41ea77f6)

jkamdjou (approved):

PR #185, opened 2022-11-22
New rule: Download link to disk image in encrypted zip

ajpc500:


detection-rules/download_link_to_disk_image_in_encrypted_zip.yml
```diff
@@ -0,0 +1,25 @@
...
+  any(body.links, 
+      any(beta.linkanalysis(.).files_downloaded, 
+          any(
+              beta.binexplode(.), (
+                  any(.flavors.yara, . == "encrypted_zip") and
+                  any(.scan.zip.all_paths, any([".img", ".iso", ".vhd"], strings.ends_with(.., .)))
+              )
+          )
+      )
+  )
```
jkamdjou, line 21: thoughts on adding the [FTS snippet](https://docs.sublimesecurity.com/docs/common-snippets#first-time-sender) at the end? since most qbot campaigns are spray and pray, i *think* that shouldn't make a big difference, but it might reduce analysis load + FPs?

jkamdjou (approved):

PR #186, opened 2022-11-22
New Rule: Rootlayer VPS in Headers

ajpc500:


morriscode:
Hey @ajpc500 catching up on the backlog here. Have you been running this rule, do you have any concerns/issues updates?

ajpc500:
Hey! This rule has been 100% TPs for us, so no changes/concerns from me. Only question would be if you wanted to aggregate suspect VPS providers into a list and create a catch-all rule for them instead of this.

jkamdjou (approved):

PR #192, opened 2022-11-22
New Rule: Attachment: Double Base64-encoded Zip File in HTML Smuggling Attachment

ajpc500:


detection-rules/attachment_double_encoded_zip_in_html_smuggle.yml
```diff
@@ -0,0 +1,41 @@
...
+              "*QJFUUBFUUCJERzVUV*",
+              "*USBFVQBFlQCR0cFV*",
+              "*BlUQRFUQRJkQENXRV*"
+          )
+      ))
+  )
+tags:
+  - "Suspicious attachment"
+  - "HTML Smuggling"
+  - "Qakbot"
```
jkamdjou, line 41:
```suggestion
  - "Qakbot"
  - "Malware"
```

jkamdjou (approved):

jkamdjou:
i just ran some Hunts, and can confirm it's picked up several campaigns already with 0 FPs 🔥

PR #193, opened 2022-11-27
New Rule: Executable embedded as an OLE object

aidenmitchell:
Detects executable files hidden inside Office documents.


aidenmitchell:
Do we want to broaden this rule to include more file extensions, such as .pdf?

aidenmitchell:
also worth noting, the `application/x-dbt` detection _should_ cover older Word documents with embedded batch files, but that data is based on a quick test I did, so it might be FP prone. If necessary, we can remove that MIME type, and do some more investigation.

jkamdjou:
> also worth noting, the `application/x-dbt` detection _should_ cover older Word documents with embedded batch files, but that data is based on a quick test I did, so it might be FP prone. If necessary, we can remove that MIME type, and do some more investigation.

yeah looks like you were right, we're seeing a bunch of FPs on that

image

jkamdjou:
looks like we get some FPs with `application/x-dosexec` :/

image

aidenmitchell:
I'll do some investigating into that, I'm not too sure why a document would flag that type without containing an executable. Will update soon.

aidenmitchell:
Any FPs on the `any(.scan.strings.strings, strings.ilike(., "*OLE Package*"))` part?

I did some more testing, the `text/x-msdos-batch` and `"OLE Package"` detections seem to be doing all of the heavy lifting anyway.

morriscode:
Rule was updated to use new $file_extensions_executables list, but I think that's a 1x daily deployment. Will revisit tomorrow for backtesting.

morriscode:
This fired 512 times in a hunt, all look to be benign messages.

jkamdjou:
should we close this?

PR #194, opened 2022-11-29
Update Punycode rule to use the punycode MDM field

rw-access:
I think this rule was created before we had the MDM field, updating it now to take advantage of it.

Does this break anything for you, @ajpc500?


ajpc500:
All good for me! 😄

ianthiel (approved):

PR #196, opened 2022-12-03
Fixing typo in rule

aidenmitchell:
My bad! Fixing typo in my "Attachment: HTML smuggling with embedded base64-encoded executable" rule!


aidenmitchell:
Nevermind, I'll be amending this rule in a new PR anyways, and will include it there.

PR #197, opened 2022-12-03
Adding JS base64 detection and fixing typo

aidenmitchell:
Editing my `Attachment: HTML smuggling with embedded base64-encoded executable` rule to include base64 JS identifiers.


jkamdjou:
hey @aidenmitchell - catching up here. it looks like this might FP quite a bit on legit base64 embedded in Javascript (eg for "secure message" mailers). we're going to go ahead and close this, but feel free to comment if you'd like to discuss further!

PR #202, opened 2022-12-06
Add file_type check to any_html_attachment rules to be more resilient

jkamdjou:
This will flag attacks with .shtml extensions, or attachments missing an extension.


ianthiel (approved):
elegant 👍

PR #204, opened 2022-12-07
New rule: Attachment: HTML smuggling with raw array buffer

jkamdjou:


detection-rules/attachment_html_smuggling_raw_buffer.yml
```diff
@@ -0,0 +1,29 @@
...
+        (
+            .file_extension in~ ("html", "htm") or
+            .file_extension in~ $file_extensions_common_archives or
+            .file_type == "html"
+        )
+        and any(beta.binexplode(.),
+                any(.scan.strings.strings,
+                    // arrayBuffer, eg repetition:
+                    // [0xa0,0x8e,0xd0,0x60,...]
+                    regex.icontains(., "(0x[a-zA-Z0-9]{2},){100}")
```
rw-access, line 20: optional spaces can't hurt
```suggestion
                    regex.icontains(., '(0x[a-zA-Z0-9]{2},\s*){100}')
```

detection-rules/attachment_html_smuggling_raw_buffer.yml
```diff
@@ -0,0 +1,29 @@
+name: "Attachment: HTML smuggling with raw array buffer"
+description: |
+  Recursively scans files and archives to detect HTML smuggling techniques.
+references:
+  - "https://playground.sublimesecurity.com?id=a6a72d97-39c7-4b12-af4d-370b654944c0"
```
rw-access, line 5: I don't think we can/should rely on these links long term

rw-access (approved):

PR #207, opened 2022-12-07
HTML link smuggling update: s/exe/vba

jkamdjou:
EXE link downloads happen too commonly, eg links to the Zoom binary.


detection-rules/link_download_suspicious_file.yml
```diff
@@ -1,6 +1,6 @@
 name: "Link to auto-download of a suspicious file type (unsolicited)"
 description: |
-  A link in the body of the email downloads a suspicious file type (or embedded file) such as an EXE, LNK, or JS file.
+  A link in the body of the email downloads a suspicious file type (or embedded file) such as an LNK, JS, or VBA.
```
ianthiel, line 3: nit: since we're referring to file types here, this would be "such as LNK, JS, or VBA"

ianthiel (approved):

PR #208, opened 2022-12-07
New Rule: PDF with embedded executable

aidenmitchell:
Similar to #193, but I believe a separate detection for PDF files makes sense.


aidenmitchell:
Closing, I just saw that a rule already exists for this. Will tweak that one.

PR #209, opened 2022-12-07
Update embedded PDF file rule to include mime detection of x-dosexec and x-msdos-batch

aidenmitchell:


morriscode (approved):
This holds up fine in testing. I don't see any issues 👍

PR #210, opened 2022-12-08
New Rule: RTF with embedded content

aidenmitchell:


morriscode:
Updated deprecated functions. Retro'd with no FP's.

morriscode (approved):

PR #211, opened 2022-12-08
Update to use new DNS for CI

alexk307:


cameron-dunn-sublime (approved):

cameron-dunn-sublime:
Was it actually succeeding or silently failing? This is good regardless, but I want to see if additional changes might be needed.

hugh-sublime (approved):

alexk307:
@cameron-dunn-sublime I would guess silently failing based on the ` --fail --silent ` flag. Is that something we still want @hugh-sublime ?

hugh-sublime:
So `--fail --silent` should just be removing the curl output but still returning a non-zero error code.

> -f, --fail

(HTTP) Fail fast with no output at all on server errors. This is useful to enable scripts and users to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22.

This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).

Providing [-f, --fail](https://curl.se/docs/manpage.html#-f) multiple times has no extra effect. Disable it again with --no-fail.

Example:

 curl --fail https://example.com
See also [--fail-with-body](https://curl.se/docs/manpage.html#--fail-with-body). This option is mutually exclusive to [--fail-with-body](https://curl.se/docs/manpage.html#--fail-with-body).

> -s, --silent

Silent or quiet mode. Do not show progress meter or error messages. Makes Curl mute. It will still output the data you ask for, potentially even to the terminal/stdout unless you redirect it.

Use [-S, --show-error](https://curl.se/docs/manpage.html#-S) in addition to this option to disable progress meter but still show error messages.

Providing [-s, --silent](https://curl.se/docs/manpage.html#-s) multiple times has no extra effect. Disable it again with --no-silent.

Example:

 curl -s https://example.com
See also [-v, --verbose](https://curl.se/docs/manpage.html#-v), [--stderr](https://curl.se/docs/manpage.html#--stderr) and [--no-progress-meter](https://curl.se/docs/manpage.html#--no-progress-meter).

PR #213, opened 2022-12-09
Accounted for byte strings in raw buffer rule

ajpc500:
These raw buffer arrays could be in the form `0x1a` or `'0x1a'`. Added an optional character to the regex to account for that


jkamdjou (approved):
thank you @ajpc500! 🙏

PR #217, opened 2022-12-10
New rule - Impersonation using recipient domain (first-time sender)

jkamdjou:
This rule is wildly effective. `strings.icontains` has opened up so many killer detections.

### DIFF
--- a/detection-rules/impersonation_recipient_domain.yml
+++ b/detection-rules/impersonation_recipient_domain.yml
@@ -1,0 +1,35 @@
+name: "Impersonation using recipient domain (first-time sender)"
+description: |
+  The recipient's domain is used in the sender's display name
+  in order to impersonate the organization. The impersonation has been 
+  observed to use both the recipient's full email address, as well as 
+  just the domain.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+
+  // only 1 To: recipient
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_recipient_domain.yml
```diff
@@ -0,0 +1,35 @@
+name: "Impersonation using recipient domain (first-time sender)"
+description: |
+  The recipient's domain is used in the sender's display name
+  in order to impersonate the organization. The impersonation has been 
+  observed to use both the recipient's full email address, as well as 
+  just the domain.
+type: "rule"
+severity: "medium"
```
jkamdjou, line 8: I've seen this deliver anything from HTML smuggling to credential phishing to spam. Since this rule isn't designed to detect any one of those specifically, "high" doesn't feel appropriate

detection-rules/impersonation_recipient_domain.yml
```diff
@@ -0,0 +1,35 @@
...
+source: |
+  type.inbound
+
+  // only 1 To: recipient
+  and length(recipients.to) + length(recipients.bcc) == 1
+  and length(recipients.cc) == 0
```
ianthiel, line 14: It looks like it's possible to CC someone and send with empty To and BCC lines, so we might include CC in the check on line 13?

ianthiel (approved):

PR #218, opened 2022-12-10
New rule - Attachment: HTML smuggling with eval and atob

jkamdjou:


ianthiel (approved):
a reference and/or deeper description would be nice but not blocking

PR #222, opened 2022-12-14
Rule Update: Link to auto-download of a suspicious file type (unsolicited)

ajpc500:
Added additional script types


jkamdjou (approved):
thank you @ajpc500 🙏

PR #223, opened 2022-12-14
New Rule: Qakbot attached file w/Adobe branding

aidenmitchell:
Will require some FP testing.


aidenmitchell:
As discussed in Slack, HTML OCR functionality may be needed to accurately detect all of these emails. The ability to use CredPhish on HTML attachments would also help.

morriscode:
@aidenmitchell  Working the backlog here, are you feeling good with where this is?

aidenmitchell:
Assuming it passed FP check, yep!

detection-rules/attachment_malicious_adobe_branding.yml
```diff
@@ -0,0 +1,40 @@
...
+  type.inbound 
+       and any(attachments,
+           (
+              .file_extension in~ ("html", "htm", "pdf") or
+              .file_extension in~ $file_extensions_common_archives or
+              .file_type == "html"
+           )
+           and any(beta.binexplode(.), 
+               any(.scan.strings.strings, strings.ilike(., "*document password*", "*the file is not displayed correctly*", "*dataset.store*", "*encrypted attachments*"))
+               or strings.ilike(.scan.ocr.raw, "*document password*", "*the file is not displayed correctly*", "*dataset.store*", "*encrypted attachments*")
```
jkamdjou, line 22: @aidenmitchell are there any opportunities to include matching on `Adobe` here, or some way for us to check that it's Adobe-branded that matches the samples you're testing against?

detection-rules/attachment_malicious_adobe_branding.yml
```diff
@@ -0,0 +1,40 @@
...
+  - "https://www.fortinet.com/blog/threat-research/delivery-of-malware-phishing-campaigns-in-q3-2022"
+  - "https://twitter.com/pr0xylife/status/1602604206346608641"
+severity: "high"
+source: |
+  type.inbound 
+       and any(attachments,
+           (
+              .file_extension in~ ("html", "htm", "pdf") or
+              .file_extension in~ $file_extensions_common_archives or
+              .file_type == "html"
```
jkamdjou, line 18: have we seen this delivered via HTML files, or strictly PDF files?

aidenmitchell: I've seen both in testing.

aidenmitchell:
So update, I don't remember _what_ payload I built this rule for, but it's failing on all of the Adobe PDFs I can find in delivr.to 😅

I'm going to refactor this rule to detect those, and do a better job at matching on `Adobe`. Standby!

aidenmitchell:
Could someone check my syntax? I _think_ it's right, but I want to be sure.

I'm going for:
1. attachment must match 1 of the file extensions
2. must match the keywords for a fake password-protected doc **or** match the keywords for a fake update window
3. regardless of step 2, must match `any(.scan.strings.strings, strings.ilike(., "*adobe*", "*acrobat*"))`
4. FTS snippet

morriscode:
I'm getting FP's on non malicious pdf files. Looks like we'll need to go back and update these rules to include the new nomenclature. beta.binexplode v file.explode if we still plan to move forward.

aidenmitchell:
Coming back to this after a bit, I think my syntax is wrong. I'll revise it and hopefully we can re-test. Thanks!

aidenmitchell:
I don't think string detections are the best route for this, closing.

PR #225, opened 2022-12-15
New Rule: HTML smuggling with embedded base64-encoded ISO

aidenmitchell:


morriscode (approved):
No FP's in testing. lgtm!

PR #226, opened 2022-12-15
Updating Rule: HTML smuggling with atob and high entropy

aidenmitchell:
Updating to add a new `atob` detection.
This will need some FP testing.


aidenmitchell:
We might need to add every permutation of the "atob" string in unicode escape sequences... thoughts?

morriscode:
Not sure how tight you want this to be specifically, but in addition to document write and atop. a.download and download wouldn't be terrible additions to this rule.. Thoughts?

Example

```
  type.inbound
  and any(attachments, .size <= 20000 and
      (
          .file_extension in~ ("html", "htm") or
          .file_extension in~ $file_extensions_common_archives or
          .file_type == "html"
      )
      and any(beta.binexplode(.),
          .scan.entropy.entropy >= 4.5
          and (
              all(["download", "document", "write", "atob"], . in ..scan.javascript.identifiers)
              // usage: document['write'](atob)
              or any(.scan.strings.strings, strings.ilike(., "*document*write*atob*"))
              or any(.scan.strings.strings, strings.ilike(., "*document*write*atob*"))
              // usage: some_var = atob();
              or any(.scan.strings.strings, strings.ilike(., "*=*atob*;"))
              or any(.scan.strings.strings, strings.ilike(., "*a.download*=*"))
              // usage: obfuscating "atob"
              or any(.scan.javascript.identifiers, strings.ilike(., '*ato\u0062*'))
          )
      )
  )
```

aidenmitchell:
Sounds like a good idea!

detection-rules/attachment_html_smuggling_atob.yml
```diff
@@ -1,22 +1,29 @@
...
 description: |
   Recursively scans files and archives to detect HTML smuggling techniques.
+references:
+  - "https://delivr.to/payloads?id=7dbf0d83-1557-4345-bf67-d18c4256b0c1"
 type: "rule"
 severity: "high"
 source: |
   type.inbound
-  and any(attachments, .size <= 60000 and
+  and any(attachments, .size <= 20000 and
```
morriscode, line 10:
```suggestion
  and any(attachments, .size <= 60000 and
```
Rule was recently updated to 60000 based on feedback from customers. Should probably keep it at that.

detection-rules/attachment_html_smuggling_atob.yml
```diff
@@ -1,22 +1,29 @@
...
+          .scan.entropy.entropy >= 4.5
           and (
-              all(["document", "write", "atob"], . in ..scan.javascript.identifiers)
-              
+              all(["download", "document", "write", "atob"], . in ..scan.javascript.identifiers)
               // usage: document['write'](atob)
+              or any(.scan.strings.strings, strings.ilike(., "*document*write*atob*"))
+              // usage: some_var = atob();
+              or any(.scan.strings.strings, strings.ilike(., "*=*atob*;"))
+              or any(.scan.strings.strings, strings.ilike(., "*a.download*=*"))
```
morriscode, line 24:
```suggestion
```
pulling download as it doesn't match rule intent (i know it was my original suggestion) :)

morriscode:
Replaced the all function with a count of the document, write, atob == 3 to avoid any issues with vacuous truth.
Rule identified 6 known samples, and matched on the delivr2 samples in the reference that we were previously missing! 

Let's ship it! 🔥

morriscode (approved):

PR #227, opened 2022-12-15
New Rule: HTML smuggling with Javascript hex array encoding

aidenmitchell:
FP test will be needed.

### DIFF
--- a/detection-rules/attachment_html_smuggling_hex.yml
+++ b/detection-rules/attachment_html_smuggling_hex.yml
@@ -16,6 +16,7 @@ source: |
      and any(beta.binexplode(.), 
          // more than 50 hex values in the JS file is indicitive of an array used to build a file
          length(filter(.scan.javascript.strings, strings.ilike(.,"*0x*"))) > 50
+         or length(filter(.scan.javascript.identifiers, strings.ilike(.,"*0x*"))) > 50
  ))
 tags:
   - "Suspicious attachment"
--- a/detection-rules/attachment_html_smuggling_hex.yml
+++ b/detection-rules/attachment_html_smuggling_hex.yml
@@ -14,10 +14,10 @@ source: |
           .file_type == "html"
       ) 
       and any(beta.binexplode(.), 
-          // more than 50 hex values in the JS file is indicitive of an array used to build a file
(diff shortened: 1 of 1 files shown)


aidenmitchell:
This rule is similar to https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/attachment_html_smuggling_raw_buffer.yml

we might want to combine them?

jkamdjou:
> This rule is similar to https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/attachment_html_smuggling_raw_buffer.yml
> 
> we might want to combine them?

yeah that might be the way to go, can you tell why the other one didn't flag this attack?

aidenmitchell:
The existing rule contains:
```
and any(.scan.strings.strings,
                    strings.contains(., ".map")
                )
```

which some payloads _don't_ contain. The rest of the rule works though.

morriscode:
@jkamdjou  @aidenmitchell's version is firing on the html reported by MH yesterday. Rather then merge two, I think we can cut 'map' without an increase in FP's. Cutting map and including Aidens string as an OR matches both samples. 

`type.inbound and
  any(attachments,
        (
            .file_extension in~ ("html", "htm") or
            .file_extension in~ $file_extensions_common_archives or
            .file_type == "html"
        )
        and any(beta.binexplode(.),
                any(.scan.strings.strings,
                    // arrayBuffer, eg repetition:
                    // [0xa0,0x8e,0xd0,0x60,...]
                    regex.icontains(., '(\W?0x[a-zA-Z0-9]{2}\W?,\s*){100}')
                )
        or any(.scan.strings.strings,
                    regex.contains(., "(?:(?:0x).*?){30,}")
                )
        )
  )
  `
If you all are cool with it I'll update the other signature and we can kill this PRl

morriscode:
Testing the hybrid rule bumped aidens repeat block to a min of 100 seems to work fine.

morriscode:
Further testing indicates this looks distinct enough to stand on its own. thanks @aidenmitchell

morriscode (approved):
testing finalized, rule seems to hold true, while matching previously missed messages! approving

PR #228, opened 2022-12-17
Updates to mitigate FPs

jkamdjou:


detection-rules/attachment_pdf_file_with_embedded_content.yml
```diff
@@ -36,7 +36,6 @@ source: |
         "rar_file",
         "tar_file",
         "xar_file",
-        "zip_file",
```
jkamdjou, line 39: @ajpc500 any issues with removing this? seeing some FPs.

ajpc500: go for it!

detection-rules/attachment_any_html_unsolicited.yml
```diff
@@ -7,7 +7,7 @@ references:
   - "https://ired.team/offensive-security/defense-evasion/file-smuggling-with-html-and-javascript"
   - "https://sandbox.sublimesecurity.com?id=106315e9-166a-4e0f-946e-88ff6fd5f9fd"
 type: "rule"
-severity: "medium"
+severity: "low"
```
ianthiel, line 10: should the "new sender" version of this rule also be `low`?

ianthiel (approved):
fwiw the removed rule generated a TP in a recent IMAP NUX test

detection-rules/attachment_any_html_unsolicited.yml
jkamdjou: from my exp "new sender" is more likely to be a TP than this one

PR #229, opened 2022-12-19
New Rule: Office document loads remote document template

aidenmitchell:
Similar to https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/attachment_office_remote_html_template.yml, however I don't believe they should be combined (this is a different method of attack, and not zero-click)

### DIFF
--- a/detection-rules/attachment_office_remote_doc_template.yml
+++ b/detection-rules/attachment_office_remote_doc_template.yml
@@ -14,7 +14,7 @@ source: |
               .file_extension in~ $file_extensions_macros
               or .file_extension in~ $file_extensions_common_archives
           )
-          and any(beta.binexplode(.),
+          and any(file.explode(.),
               .flavors.mime == "text/xml" and
               any(.scan.strings.strings, regex.icontains(., ".*Target.*http.*dotm.*"))
           )
--- a/detection-rules/attachment_office_remote_doc_template.yml
+++ b/detection-rules/attachment_office_remote_doc_template.yml
@@ -11,7 +11,7 @@ source: |
       (
           (
               // office files
-              .file_extension in~ ("doc", "docm", "docx", "dot", "dotm", "pptm", "ppsm", "xlm", "xls", "xlsb", "xlsm", "xlt", "xltm")
+              .file_extension in~ $file_extensions_macros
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_office_remote_doc_template.yml
```diff
@@ -0,0 +1,24 @@
...
+  - "https://delivr.to/payloads?id=c7a7195e-0de3-428d-a32c-5fd59a3012da"
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(attachments,
+      (
+          (
+          .file_extension in~ $file_extensions_macros 
+          or .file_extension in~ $file_extensions_common_archives
```
jkamdjou, line 14:
```suggestion
              // office files
              .file_extension in~ ("doc", "docm", "docx", "dot", "dotm", "pptm", "ppsm", "xlm", "xls", "xlsb", "xlsm", "xlt", "xltm")
              or .file_extension in~ $file_extensions_common_archives
```

templates can be loaded from non-macro enabled office files, so this might be more desirable. we have an example of this usage in this rule: https://github.com/sublime-security/sublime-rules/blob/1a5c6923ba6f61c7b8b1e6fae22bb669af192502/detection-rules/attachment_office_remote_html_template.yml#L31

detection-rules/attachment_office_remote_doc_template.yml
```diff
@@ -0,0 +1,24 @@
+name: "Attachment: Office document loads remote document template"
+description: |
+  Recursively scans archives and Office documents to detect remote document template injection.
+references:
+  - "https://delivr.to/payloads?id=c7a7195e-0de3-428d-a32c-5fd59a3012da"
+type: "rule"
+severity: "high"
```
jkamdjou, line 7:
```suggestion
severity: "medium"
```
since we don't know for sure this is bad, medium might be better

jkamdjou (commented):
this looks solid - just a couple minor comments! (cc @morriscode)

aidenmitchell:
Thanks for the suggestions! They've been added.

morriscode:
s/beta.binexplode/file.explode/g

morriscode (approved):
⭐

PR #230, opened 2022-12-19
New Rule: Document with DDE command

aidenmitchell:
This one will need FP testing. It's _very_ open-ended in its file extension filtering, and there may also be other DDE variations; I'll be working on both of those issues.

### DIFF
--- a/detection-rules/attachment_office_dde_function.yml
+++ b/detection-rules/attachment_office_dde_function.yml
@@ -1,0 +1,22 @@
+name: "Attachment: Document with DDE command"
+description: |
+  Recursively scans files and archives to detect Office documents with DDE functions that contain commands.
+  If a user disregards Office warnings, unwanted commands could execute on their machine.
+type: "rule"
+references:
+  - "https://delivr.to/payloads?id=55075047-bdd7-41aa-a791-0cf509c38bc9"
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_office_dde_function.yml
```diff
@@ -0,0 +1,22 @@
...
+source: |
+  type.inbound 
+      and any(attachments,
+          (
+          .file_extension in~ $file_extensions_macros 
+          or .file_extension in~ $file_extensions_common_archives
+          or .file_extension == "csv"
+          )
+          and any(beta.binexplode(.),
+              any(.scan.strings.strings, strings.ilike(., "*cmd|*/c*", "*powershell|*", "*DDEAUTO*", "*DDE*cmd*/c*", "*DDE*powershell*"))
```
morriscode, line 18: do we want all of these strings to include `DDE`? as written, we might fire on non-DDE for the first 2 strings. also do we need to include DDEAUTO to flag this payload?

aidenmitchell: Yeah, I'll remove the first 2, they might be FP prone. `DDEAUTO`, I believe I included it because it's another possibility, but I didn't see an actual payload.

PR #233, opened 2023-01-06
New Rule: HTML smuggling sign-in prompt

aidenmitchell:
This will need some FP testing (and maybe a better name 😛)


jkamdjou:
it looks like this will hit on legit sign-in pages, eg "Secure Message" sharing applications, so we're going to go ahead and close this but let us know if you'd like to discuss further!

PR #237, opened 2023-01-14
Ignore YARA files for repo language stats

jkamdjou:
I think this is a bit misleading:

image

image

I tested this on my own fork and this will hide YARA from the language stats.


cameron-dunn-sublime (approved):
👍 good call!

PR #240, opened 2023-01-26
New Rule: MSFT Forms Google Impersonation

bfilar:
### Overview
Some attacks impersonate Google, where users are urged to secure their accounts. All emails come from the same sender domain: `email[.]formspro[.]microsoft[.]com`. 

### Example
Screenshot 2023-01-26 at 11 43 12 AM

### DIFF
--- a/detection-rules/impersonation_google.yml
+++ b/detection-rules/impersonation_google.yml
@@ -1,0 +1,24 @@
+name: "Brand impersonation: Google"
+description: |
+  Impersonation of Google.
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and (
+  (
+      strings.ilike(subject.subject, '*urgent*') 
+      or strings.ilike(subject.subject, '*alert*')
+  )
+  or
+      (
+      strings.ilike(body.html.display_text, "*google*")
+      and strings.ilike(body.html.display_text, "*suspicious*")
+      and strings.ilike(body.html.display_text, "*secure*")
+      )
+  )
+  and sender.email.domain.domain == 'email.formspro.microsoft.com'
+
+tags:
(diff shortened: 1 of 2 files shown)


detection-rules/impersonation_google.yml
```diff
@@ -0,0 +1,24 @@
+name: "Brand impersonation: Google"
```
jkamdjou, line 1:
```suggestion
name: "Brand impersonation: Google using Microsoft Forms"
```

detection-rules/impersonation_google.yml
```diff
@@ -0,0 +1,24 @@
...
+description: |
+  Impersonation of Google.
```
jkamdjou, line 3:
```suggestion
  Abuses Microsoft Forms to impersonate Google.
```

detection-rules/impersonation_google.yml
```diff
@@ -0,0 +1,24 @@
...
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and (
+  (
+      strings.ilike(subject.subject, '*urgent*') 
+      or strings.ilike(subject.subject, '*alert*')
+  )
+  or
```
jkamdjou, line 13: these should be indented

jkamdjou, line 13: should this be an `and`? as is, it doesn't have to include "google" to flag

jkamdjou (commented):
🙏🙏 thank you @bfilar ! a couple comments and suggestions since we're looking at a specific attack type in this rule

bfilar:
Renamed the file as well to better match the description

jkamdjou (approved):

PR #242, opened 2023-01-31
New Rule: HTML Smuggling with hex encoded strings

ajpc500:
similar to the raw buffer rule but different formatting


jkamdjou (approved):
LGTM! Thanks @ajpc500! 🙌

PR #243, opened 2023-01-31
Updating rule logic after efficacy testing

morriscode:
Updating rule logic to reduce FP's. 
Removed old rule to simplify nomenclature and tags. 
New Rule has parity in terms of detection with old rule + more TP's and less FP's.


morriscode:
please ignore

PR #245, opened 2023-02-02
Adjusted entity rule logic for employee impersonation

bfilar:


morriscode (approved):
Looks good. Testing resulted in a TP and no FP's!

morriscode:
@bfilar you're GTG 🔥

PR #249, opened 2023-02-05
New Rule: Credential theft attempt from first-time sender

aidenmitchell:


morriscode (commented):
Hey @aidenmitchell This looks conceptually sound. During my testing it's flagging on some PDFs that do have a link, but they're not credential phishing. PDF files, containing links to fake portals, however the emails give the user a set of creds, and are attempting to solicit funds rather than credentials. (i actually just published a rule specifically for these) 

Here's an example: Just wanted to clarify if this matches your expected intent. 
https://playground.sublimesecurity.com?id=884f82fa-6d3f-4731-a905-66460a956e55

aidenmitchell:
Interesting, that is not my intent. I'll filter the NLU part a little more (I'll probably end up excluding financial keywords or something)

Thanks!

PR #250, opened 2023-02-05
New Rule: File execution via Javascript

aidenmitchell:
This one might be _very_ FP prone; will need a through test.


aidenmitchell:
This rule was created in response to delivr.to's bad_signature_js_motw_bypass.zip file (js file with an invalid signature, exploiting Windows to bypass MOTW), so I'm currently looking into ways to see if we can specifically detect invalid code signatures... that would be a better way imo

morriscode:
Heads up I'm seeing a few FP's with this. Predominantly looks like Zip files from Varonis and other monitoring tools sending zipped text files that contain these strings. Might be able to filter these with FTS, or other $sent/$received lists?

morriscode:
Apologies for delays! Retroing now.

morriscode:
I see why... 

all(input: [T], expression: bool) -> bool
all returns true if all of the values in the original array match an expression, which is provided by the second argument. **If the array is empty, then all is vacuously `true.**

detection-rules/attachment_js_file_execution.yml
```diff
@@ -0,0 +1,27 @@
...
+  Javascript contains identifiers or strings that may attempt to execute files.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound 
+  and any(attachments, .file_type in $file_extensions_common_archives
+      and any(beta.binexplode(.), 
+          any(.scan.javascript.identifiers, strings.ilike(., 'ActiveXObject', 'ShellExecute'))
+          or all(.scan.javascript.strings, strings.ilike(., 'Shell.Application', '*.exe'))
+      )
```
morriscode, line 12:
```suggestion
        or (length(.scan.javascript.strings) > 0 and 
        all(.scan.javascript.strings, strings.ilike(., 'Shell.Application', '*.exe')))
    )
)
```
In order to avoid the vacuous truth the all was firing we could encapsulate a length check on .scan.javascript.strings prior to calling the all like this.

detection-rules/attachment_js_file_execution.yml
```diff
@@ -0,0 +1,27 @@
...
+description: |
+  Javascript contains identifiers or strings that may attempt to execute files.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound 
+  and any(attachments, .file_type in $file_extensions_common_archives
+      and any(beta.binexplode(.), 
+          any(.scan.javascript.identifiers, strings.ilike(., 'ActiveXObject', 'ShellExecute'))
+          or all(.scan.javascript.strings, strings.ilike(., 'Shell.Application', '*.exe'))
```
morriscode, line 11:
```suggestion
          or (length(.scan.javascript.strings) > 0
            and all(.scan.javascript.strings, strings.ilike(., 'Shell.Application', '*.exe')))
```

In order to avoid the vacuous truth caused by .scan.javascript.strings being null, we can encapsulate a length check with the all condition.

morriscode (approved):
This has been in testing for 2 weeks with no FP's... but also no TP's. Should be safe to deploy!

PR #255, opened 2023-02-08
New Rule: Body: Callback Phishing PayPal Invoice Abuse

morriscode:


detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
```diff
@@ -0,0 +1,43 @@
...
+description: |
+  A fraudulent invoice/receipt found in the body of the message sent by exploiting Paypal's invoicing service.
+  Callback Phishing is an attempt by an attacker to solicit the victim (recipient) to call a phone number. 
+  The resulting interaction could lead to a multitude of attacks ranging from Financial theft, Remote Access Trojan (RAT) Installation or Ransomware Deployment.
+type: "rule"
+references:
+  - "https://anderegg.ca/2023/02/01/a-novel-paypal-scam"
+severity: "medium"
+source: |
+  type.inbound
```
jkamdjou, line 11:
```suggestion
   type.inbound
```

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
```diff
@@ -0,0 +1,43 @@
...
+               strings.ilike(body.html.inner_text, '*cancel*'),
+               strings.ilike(body.html.inner_text, '*renew*'),
+               strings.ilike(body.html.inner_text, '*refund*'),
+               strings.ilike(body.html.inner_text, '*+1*'),
+               strings.ilike(body.html.inner_text, '*help*desk*')
+           )
+       )
+   or (
+       //Unicode confusables words obfuscated in note
+       regex.contains(body.html.inner_text, '\+𝟭|𝗽𝗮𝘆𝗺𝗲𝗻𝘁|𝗛𝗲𝗹𝗽 𝗗𝗲𝘀𝗸|𝗿𝗲𝗳𝘂𝗻𝗱|𝗮𝗻𝘁𝗶𝘃𝗶𝗿𝘂𝘀|𝗰𝗮𝗹𝗹|𝗰𝗮𝗻𝗰𝗲𝗹')
```
jkamdjou, line 38:
```suggestion
       or (
           //Unicode confusables words obfuscated in note
           regex.contains(body.html.inner_text, '\+𝟭|𝗽𝗮𝘆𝗺𝗲𝗻𝘁|𝗛𝗲𝗹𝗽 𝗗𝗲𝘀𝗸|𝗿𝗲𝗳𝘂𝗻𝗱|𝗮𝗻𝘁𝗶𝘃𝗶𝗿𝘂𝘀|𝗰𝗮𝗹𝗹|𝗰𝗮𝗻𝗰𝗲𝗹')
```

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
```diff
@@ -0,0 +1,43 @@
...
+type: "rule"
+references:
+  - "https://anderegg.ca/2023/02/01/a-novel-paypal-scam"
+severity: "medium"
+source: |
+  type.inbound
+   and length(attachments) == 0
+   and sender.email.domain.root_domain == "paypal.com"
+   and not any(headers.hops, .authentication_results.dmarc == "fail")
+   and strings.ilike(body.html.display_text, "*seller note*")
```
jkamdjou, line 15: is this an artifact that will always be here, i.e. it's on legit paypal invoices?

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
jkamdjou: fixing some spacing to make the hierarchy more clear

jkamdjou (commented):
LGTM, just some minor formatting

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
rw-access: we definitely need a function to strip out confusables, and it should probably be automatic for `display_text`

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
```diff
@@ -0,0 +1,43 @@
+name: "Body: Callback Phishing PayPal Invoice Exploitation"
```
jkamdjou, line 1:
```suggestion
name: "PayPal Invoice Abuse"
```

jkamdjou: since we already have `Callback Phishing` as a Tag, this could simplify the rule name a bit

if you like that, we could also change the file name to `paypal_invoice_abuse.yml`

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
morriscode: It can be on legit invoices if the invoicer decides to leave a note. However it is not on all invoices spawned from the generator. 

It should be ever present on **these** attacks as without it, it will just say "X sent you and invoice for $xxx.xx". Without it is no place for them to put the phone number and have that show up in the Email html body.  

That said... It is possible to send an invoice and not use the customer note box. You could put the phone number and callback context in the "description" field instead. However that would require the end user click the link and view the invoice on paypals website. the context and phone number would not be in the html body of the email. 

I haven't seen this exploited in callback phishing .... yet but we could explore the lack of note using a different rule.  eg. dmarc !=fail and sender is paypal, and does not contain seller note... then use LA to analyze the returned page for the same strcontent and phone number inspection we're doing here.

detection-rules/body_paypal_invoice_exploit_callback_phishing.yml
```diff
@@ -0,0 +1,43 @@
...
+               strings.ilike(body.html.inner_text, '*invoice*'),
+               strings.ilike(body.html.inner_text, '*call*'),
+               strings.ilike(body.html.inner_text, '*cancel*'),
+               strings.ilike(body.html.inner_text, '*renew*'),
+               strings.ilike(body.html.inner_text, '*refund*'),
+               strings.ilike(body.html.inner_text, '*+1*'),
+               strings.ilike(body.html.inner_text, '*help*desk*')
+           )
+       )
+       or (
```
jkamdjou: this `or` is at the same level as the previous `and`, so really i think we're or'ing these 3 expressions. i'm fine with it if you are, just making sure that's what you intended!

morriscode: Good call that or should be matching the upper and. I think it was functioning appropriately as the parens were correct but yeah it was indented incorrectly.

jkamdjou (approved):

PR #257, opened 2023-02-10
add insertAdjacentHTML to HTML Smuggling atob rule

arnydo:
Example:  `document.head.insertAdjacentHTML("beforeend", atob(...`


morriscode (approved):
thanks @arnydo!! lgtm

PR #258, opened 2023-02-10
Updated Rule: HTML smuggling with eval and atob

arnydo:
Proposing a small change to detect more `eval*atob` activity.
Current detection requires `onerror` to precede `eval(atob` which is not found in all of these smuggling attacks I have detected.

Let me know your thoughts.

Thanks!

### DIFF
--- a/detection-rules/attachment_html_smuggling_eval_atob.yml
+++ b/detection-rules/attachment_html_smuggling_eval_atob.yml
@@ -13,7 +13,7 @@ source: |
       )
       and any(beta.binexplode(.),
           // usage: onerror="eval(atob('
-          any(.scan.strings.strings, strings.ilike(., "*onerror*eval*atob*"))
+          any(.scan.strings.strings, regex.imatch(., ".*eval.{1,4}atob.*"))
       )
   )
 tags:


morriscode:
thanks @arnydo that's interesting to me, i was under the impression that the onerror event handler was what was being used as a trigger to execute... do the other samples not match the other [atob](https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/attachment_html_smuggling_atob.yml) signature either?

arnydo:
onerror was not present in this HTML file at all.

`Attachment: HTML smuggling with atob and high entropy` did flag this one.

arnydo:
What are your thoughts on combining these two rules?

morriscode:
I just wrote the combo rule, the only question I have is around the size and entropy gating reqs from the other rule are still relevant.

morriscode:
@jkamdjou thoughts on merging these two rules?  I sent you a snippet of the combo rule on slack.

jkamdjou:
> @jkamdjou thoughts on merging these two rules? I sent you a snippet of the combo rule on slack.

i'm thinking let's hold off for now, as there's a chance it could introduce FPs and the risk doesn't seem worth the gain atm. lmk if you feel strongly tho / wanna discuss more

morriscode:
roger that! i'm merging changes as requested since this was missing samples @arnydo was seeing.  we can revisit combo in a couple weeks

morriscode (approved):
changes match missing samples, no additional fps uncovered in testing.

PR #259, opened 2023-02-11
Adding new relationship detection to remote template rule

aidenmitchell:
As a result of https://delivr.to/payloads?id=ef187d5d-3188-483e-b3b5-9ab5e0e032f7.


morriscode:
💪 
`regex.icontains(.target, "html?!", "dotm"))`

Since we're looking for dotm, should we look for other macro enabled ole file types too? Thinking about how exhaustive the relationship inspection should be?

aidenmitchell:
> should we look for other macro enabled ole file types too

Yeah that's a good idea, I'll add those. I don't think we can reference the macro extension list, due to the regex, but that's ok

morriscode:
Apologies for delays, retroing now.!

morriscode (approved):
👍

detection-rules/attachment_office_remote_html_template.yml
```diff
@@ -33,7 +33,7 @@ source: |
               // detect files without an extension
               or .file_type in ("docx")
           )
-          and any(file.oletools(.).relationships, regex.icontains(.target, "html?!"))
+          and any(file.oletools(.).relationships, regex.icontains(.target, "html?!", "doc", "docm", "docx", "dot", "dotm", "pptm", "ppsm", "xlm", "xls", "xlsb", "xlsm", "xlt", "xltm"))
```
rw-access, line 36: what do `htm!` and `html!` mean? looks like the `!` is interpreted literally in RE2, not sure if that's intended.

rw-access: maybe @jkamdjou knows since those were in the original rule

aidenmitchell: yeah not sure...

PR #260, opened 2023-02-12
Update Rule: Brand impersonation: Amazon

morriscode:
- Updating Rule: Strengthen Brand impersonation: Amazon
- Updating Rule Logic to match sample

### DIFF
--- a/detection-rules/impersonation_amazon.yml
+++ b/detection-rules/impersonation_amazon.yml
@@ -42,4 +42,3 @@ source: |
 tags:
   - "Brand impersonation"
   - "Suspicious sender"
-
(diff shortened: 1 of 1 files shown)


jkamdjou (commented):
🔥 just to confirm:

> aaa𝝰aａ𝑎𝗮𝕒𝖆𝓪𝚊𝞪аɑα𝔞𝒂𝘢𝛂⍺𝒶𝙖𝜶𝛼𝐚𝖺
these are all homoglyphs

> mazon
these are all ascii

is that right?

jkamdjou (approved):

PR #261, opened 2023-02-13
Update Insight: Body domains/Sender domain - adding count >0

morriscode:


jkamdjou (approved):
nice 🔥 (reminder to update in frontend too 🥴)

PR #262, opened 2023-02-13
Prioritize display_text as NLU input for BEC rule

bfilar:


morriscode (approved):
lgtm! sidenote: there's nothing in the docs about coalesce as a function..

PR #263, opened 2023-02-13
Removing FP Prone rules.

morriscode:
These rules, are too generalized to be accurate in an alerting capacity. 
Blank Message relies on a char count, which is inaccurate due to mail client/warning tag inclusion. 
The specific url presence (firebaseurl/freesubdom blocklist) are not strong enough indicators to not produce FP's.


jkamdjou (approved):
we could consider moving at least `link_firebase_storage.yml` to `./discovery-rules`

morriscode:
> 

Good idea! I'll merge this and put link_firebase_storage.yml to discover-rules in a future PR!

PR #264, opened 2023-02-13
Updating Cryptocurrency insight queries to use regex.icontains

morriscode:


jkamdjou (approved):
are these case insensitive, or could `regex.contains` work (it'd be more efficient for a backtest)?

morriscode:
> are these case insensitive, or could `regex.contains` work (it'd be more efficient for a backtest)?

I initially replaced it with contains but found usage of attackers putting them in all lower, or all upper.  :/

PR #265, opened 2023-02-13
New Rules using org_sld lists

morriscode:


detection-rules/impersonation_org_sld_in_filename_fts.yml
```diff
@@ -0,0 +1,28 @@
+name: "Impersonation: Organizational SLD found in attachment file name from FTS"
+type: "query"
```
cameron-dunn-sublime, line 2:
```suggestion
type: "rule"
```

detection-rules/impersonation_org_sld_in_sender_sld_fts.yml
```diff
@@ -0,0 +1,26 @@
+name: "Impersonation: Organizational SLD found in Sender domain"
+type: "query"
```
cameron-dunn-sublime, line 2:
```suggestion
type: "rule"
```

jkamdjou:
should we just move these to signals?

PR #266, opened 2023-02-14
New Rule: Link to auto-downloaded file with Google Drive branding

aidenmitchell:
Heavily inspired by https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/link_html_smuggling_with_adobe_branding.yml (thanks @ajpc500!)


morriscode (approved):
👍 thanks @aidenmitchell this looks great! no fp's in testing! lets roll!

PR #269, opened 2023-02-16
New rule: Attachment: Double File Extension in Filename FTS

morriscode:


morriscode:
The more i think about this, I think it should probably be an insight... @jkamdjou thoughts? This seems a bit weak even coupled with first time sender.

jkamdjou:
> The more i think about this, I think it should probably be an insight... @jkamdjou thoughts? This seems a bit weak even coupled with first time sender. 

👍 it's a bit risky as a rule on its own i think too. down to make it an insight!

PR #272, opened 2023-02-18
New Rule: Spam Campaign with Excessive space obfuscation

morriscode:


detection-rules/spam_campaign_excessive_space_obfuscation_free_file_hosting.yml
```diff
@@ -0,0 +1,26 @@
+name: "Spam: Campaign with excessive space/char obfuscation and free file hosted link"
+description: |
+  This rule detects a mass spam campaigns using excessive space padding with
```
jkamdjou, line 3:
```suggestion
  This rule detects mass spam campaigns using excessive space padding with
```

jkamdjou (approved):
🔥 just synced up with @rw-access - his advice is the way we parse `.inner_text` may change in the future, so a more resilient method would be to look for the excessive `<span`> pattern in the html raw. no need to worry about that right now tho, just a note for future ref / consideration!

PR #285, opened 2023-02-22
New Rules: New Domain in Link and New From Domain + FTS

morriscode:


detection-rules/link_new_domain_in_link_first_time_sender.yml
```diff
@@ -0,0 +1,22 @@
+name: "Link: New domain (<7d) in link from first-time sender"
```
jkamdjou, line 1:
```suggestion
name: "New link domain (<7d) from first-time sender"
```

i think we can simplify a bit

detection-rules/link_new_domain_in_link_first_time_sender.yml
```diff
@@ -0,0 +1,22 @@
...
+description: |
+  Detects links in the body of an email where the linked domain is less than 7 days old.
+type: "rule"
+severity: "high"
```
jkamdjou, line 5:
```suggestion
severity: "medium"
```

thoughts on medium? i think if we combined this with other clearly bad stuff it could definitely be high

detection-rules/sender_new_from_domain_first_time_sender.yml
```diff
@@ -0,0 +1,21 @@
+name: "Sender: new domain (<7d) from first-time sender"
```
jkamdjou, line 1:
```suggestion
name: "New sender domain (<7d) from first-time sender"
```

detection-rules/link_new_domain_in_link_first_time_sender.yml
```diff
@@ -0,0 +1,22 @@
+name: "Link: New domain (<7d) in link from first-time sender"
+description: |
+  Detects links in the body of an email where the linked domain is less than 7 days old.
```
jkamdjou, line 3:
```suggestion
  Detects links in the body of an email where the linked domain is less than 7 days old from first-time senders.
```

detection-rules/sender_new_from_domain_first_time_sender.yml
```diff
@@ -0,0 +1,21 @@
...
+description: |
+  Detects inbound emails where the sender domain is less than 7 days old from first-time sender
```
jkamdjou, line 3:
```suggestion
  Detects inbound emails where the sender domain is less than 7 days old from first-time senders.
```

detection-rules/sender_new_from_domain_first_time_sender.yml
```diff
@@ -0,0 +1,21 @@
...
+                and sender.email.email not in $sender_emails
+            )
+            or (
+                sender.email.domain.root_domain not in $free_email_providers
+                and sender.email.domain.domain not in $sender_domains
+            )
+  )
+tags:
+  - "Suspicious sender"
+  - "Domain age"
```
jkamdjou, line 21:
```suggestion
  - "New domain"
```

i think this would be more informative - eg when you're seeing the Tags of rules that fired on a message in the MDV. wdyt?

detection-rules/link_new_domain_in_link_first_time_sender.yml
```diff
@@ -0,0 +1,22 @@
...
+      and sender.email.email not in $sender_emails
+    )
+    or (
+      sender.email.domain.root_domain not in $free_email_providers
+      and sender.email.domain.domain not in $sender_domains
+    )
+  )
+tags:
+  - "Suspicious link"
+  - "Domain age"
```
jkamdjou, line 22:
```suggestion
  - "New domain"
```

detection-rules/sender_new_from_domain_first_time_sender.yml
morriscode: I dig it! I was going to just add Whois enrichment but I like this better

detection-rules/sender_new_from_domain_first_time_sender.yml
```diff
@@ -0,0 +1,21 @@
...
```
jkamdjou, line 21:
```suggestion
  - "New domain"
```

jkamdjou (approved):

PR #287, opened 2023-02-23
Callback phishing updates: Split img/pdf rules into two to mitigate FPs in multipage pdfs

morriscode:


detection-rules/attachment_callback_phish_with_img.yml
```diff
@@ -51,4 +51,4 @@ source: |
             )
 tags:
   - "Callback Phishing"
-  - "Customer service fraud"
\ No newline at end of file
+  - "Customer service fraud"
```
jkamdjou, line 54:
```suggestion
  - "Customer service fraud"
id: "60acbb36-8ed1-562e-8027-260c2fdf0f04"
```

jkamdjou: image

jkamdjou (approved):

PR #291, opened 2023-02-24
New Rule: eml file attachment containing html file

morriscode:


detection-rules/attachment_eml_file_with_html_attachment.yml
```diff
@@ -0,0 +1,31 @@
+name: "Attachment: EML file with HTML attachment (first-time sender)"
+description: |
+  Potential HTML smuggling via HTML file inside embedded rfc822 (eml)
+type: "query"
+severity: "medium"
+source: |
+  type.inbound
+  and not strings.like(sender.email.local_part, "postmaster", "mailer-daemon")
```
jkamdjou, line 8:
```suggestion
  // exclude bounce backs
  and not strings.like(sender.email.local_part, "postmaster", "mailer-daemon")
```

detection-rules/attachment_eml_file_with_html_attachment.yml
```diff
@@ -0,0 +1,31 @@
+name: "Attachment: EML file with HTML attachment (first-time sender)"
+description: |
+  Potential HTML smuggling via HTML file inside embedded rfc822 (eml)
+type: "query"
+severity: "medium"
```
jkamdjou, line 5:
```suggestion
severity: "low"
```

wdyt about `low` here, since we're not looking for any specific smuggling signals?

jkamdjou (approved):

PR #293, opened 2023-02-28
Updating Rule: Brand impersonation: Github

Jeffreyhung:
[githubstatus.com](https://githubstatus.com) is a legit domain used by GitHub to send out operational status updates. 
Adding it to the legit domain list to prevent status update emails from creating false alerts.


jkamdjou (approved):
thank you @Jeffreyhung!

PR #295, opened 2023-03-06
Update Rule: Body HTML Callback phish to mitigate fp

morriscode:
- Updating Rule: Body HTML: Callback phish to mitigate FP


jkamdjou (approved):
🙌

PR #300, opened 2023-03-07
Update Rule: Notion suspicious file share

Jeffreyhung:
Notion supports customizing slug for each organization.
Notion screenshot

By providing the option to exclude your organization slug in the detection rule, we can prevent internal Notion links from triggering the alert and cause false positives.

Reference: [Public pages & web publishing](https://www.notion.so/help/public-pages-and-web-publishing)


morriscode:
Thanks @Jeffreyhung  I love the rule and I like how you're thinking! Presently our rule feed is for rules that function out of the box with no customization required.  In the near future we hope to have more modular capabilities, and we’d love to re-open this at that time!

PR #301, opened 2023-03-08
New Rule: Attachment: Emotet heavily padded doc in zip

morriscode:


morriscode:
Including First-Time-Sender for now, I don't foresee many orgs having previously received messages from emotet senders at this time.

morriscode:
Open to suggestions on file and rule name.

detection-rules/attachment_emotet_heavily_padded_doc_in_zip.yml
```diff
@@ -0,0 +1,32 @@
...
+  - "https://twitter.com/Cryptolaemus1/status/1633099154623803394"
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(attachments, 
+    .file_extension == "zip"
+    and any(file.explode(.), 
+      .depth == 0 and .size < 1000000
+      and not .depth > 0
```
cameron-dunn-sublime, line 15:
```suggestion
      .depth == 0 and .size < 1000000
```

morriscode: hmm. I want to ensure there's never any object at .depth >0

morriscode, line 15:
```suggestion
      not .depth > 0
      and .depth == 0 and .size < 1000000
```

I could see reversing that order to bail if there are depths greater than 0?

jkamdjou: it might be clearer as 2 calls to fileexplode, where we check the depth separately in each? since depth can never be both 0 and greater than 0 on the same file that's analyzed

but i'm not sure it actually matters anyway - as is should have the same effect i think? cc @rw-access

jkamdjou (approved):
going to approve so this gets out to users ASAP as the technique is currently being observed ITW. @rw-access feel free to chime in in the AM and we can make updates to the rule for clarity if needed

jkamdjou:
great work @morriscode !

detection-rules/attachment_emotet_heavily_padded_doc_in_zip.yml
```diff
@@ -0,0 +1,32 @@
...
+      and strings.ends_with(.scan.exiftool.zip_file_name, ".doc")
+      and .scan.exiftool.zip_uncompressed_size > 500000000
```
rw-access, line 17: WDYT about checking the `.size` on the attachment itself instead? could be a lot more efficient that way

detection-rules/attachment_emotet_heavily_padded_doc_in_zip.yml
rw-access: `not .depth > 0` will automatically be rewirrten to `.depth <= 0`, so this rule will evaluate: `.depth == 0 and .size < 1000000 and .depth <= 0`.

rw-access: TBH, I'm pretty confused by the rule. Something like this makes more sense to me intuitvely, if I'm understanding right. WDYT @jkamdjou @morriscode?

```sql
  type.inbound
  and any(attachments, 
    .file_extension == "zip"
    and .size < 1000000
    and any(file.explode(.),
      and strings.ends_with(.scan.exiftool.zip_file_name, ".doc")
      and .scan.exiftool.zip_uncompressed_size > 500000000
    )
  )
  and (
    (
      sender.email.domain.root_domain in $free_email_providers
      and sender.email.email not in $sender_emails
    )
    or (
      and sender.email.domain.domain not in $sender_domains
    )
  )
```

rw-access: sounds like you want to do two checks?

PR #303, opened 2023-03-08
New Rule: Attachment PDF single link to zip no creator

morriscode:
- New Rule: Attachment: PDF link to zip no creator
- Renaming file
- Deleting old file
- Removing dev/shm logic adding proper creator line....
😮


detection-rules/attachment_pdf_link_to_zip_no_creator.yml
```diff
@@ -0,0 +1,17 @@
...
+description: "Detects a PDF attachment with a single link to a ZIP file. Bumblebee delivery mechanism"
+type: "rule"
+severity: "high"
+source: |
+  type.inbound 
+  and all(attachments, 
+    .file_extension == "pdf"
+    and any(file.explode(.), (
+      .depth == 0 
+      and length(filter(.scan.strings.strings, regex.contains(.,'\/S\/URI\/URI\(http(s):\/\/.+\.zip'))) == 1
```
jkamdjou, line 11: i'm guessing the URL scanner isn't pulling this out reliably?

jkamdjou:
> * Renaming file
> * Deleting old file
> * Removing dev/shm logic adding proper creator line....
>   😮

did you mean to commit the deletion of an existing rule?

detection-rules/attachment_pdf_link_to_zip_no_creator.yml
morriscode: ScanUrl doesn't run on PDFs unfortunately. This is resolved in the new scanner, but I don't think we're ready to incorporate that yet.

morriscode:
> 

Haha, yes, so I meant to just replace the file I put in there with the corrected one, but the one I put in there was named differently. Cascading failures on my part but this is right now.

jkamdjou (approved):

PR #312, opened 2023-03-13
New Rule (Link): PDF with link to zip and no 'Creator' exif metadata

aidenmitchell:


morriscode:
Thanks @aidenmitchell. I like the concept here, and I'm not worried about the logic as the foundational rule exists. 

These are just really difficult to backtest due to the "any(body.links" being an expensive historical function. 

@jkamdjou I'm ok with approving this and [310](https://github.com/sublime-security/sublime-rules/pull/311) and [311](https://github.com/sublime-security/sublime-rules/pull/310) since they're just link -> files downloaded evaluations of existing logic... but I wanted your thoughts.

jkamdjou:
> Thanks @aidenmitchell. I like the concept here, and I'm not worried about the logic as the foundational rule exists.
> 
> These are just really difficult to backtest due to the "any(body.links" being an expensive historical function.
> 
> @jkamdjou I'm ok with approving this and [310](https://github.com/sublime-security/sublime-rules/pull/311) and [311](https://github.com/sublime-security/sublime-rules/pull/310) since they're just link -> files downloaded evaluations of existing logic... but I wanted your thoughts.

yeah i think we'll want some logic to narrow down the scope a bit. as-is, we'd be sending all links from every inbound message to LinkAnalysis, which will be too much of a burden for performance reasons and also introduces some additional "click" risk

one option might be our standard "unsolicited" or FTS snippet

aidenmitchell:
that's supposed to say "FTS snippet", my bad!

morriscode:
Going to close this Aiden, the logic itself was FP prone.

PR #314, opened 2023-03-13
New Rule: RTF body exploiting CVE-2023-21716

aidenmitchell:


morriscode:
Retro'ing now!

morriscode:
Are we likely to see RTF in a body? It's my understanding that RTF is only preserved on internal to internal exchange messages, as soon as you send it outside the org, I believe it's converted to HTML. Let me confirm that's the case.

PR #319, opened 2023-03-16
Authentication updates, and New DMARC +  credphish rule

morriscode:
Updates: 

- Queries: Authentication queries like to ilike to account for casing
- Rule: Impersonation_Docusign to account for DMARC failures

New Rule: 

- impersonation_dmarc_failure_high_cred_phish_conclusion.yml

### DIFF
--- a/detection-rules/impersonation_dmarc_failure_high_cred_phish_conclusion.yml
+++ b/detection-rules/impersonation_dmarc_failure_high_cred_phish_conclusion.yml
@@ -1,0 +1,16 @@
+name: "DMARC failure with high confidence credential theft intent"
+description: "Detects DMARC failures and messages with a high confidence of credential theft"
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(distinct(headers.hops, .authentication_results.dmarc is not null), strings.ilike(.authentication_results.dmarc, "*fail"))
+  and any([body.plain.raw, body.html.inner_text], 
+    any(ml.nlu_classifier(.).intents,
+      .name in ("cred_theft")
+      and .confidence == "high"
+    )
+  )
+tags:
+  - "DMARC failure"
+  - "Credential theft intent"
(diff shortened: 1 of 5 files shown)


detection-rules/impersonation_dmarc_failure_high_cred_phish_conclusion.yml
```diff
@@ -0,0 +1,16 @@
+name: "Impersonation: DMARC failure with high confidence credential theft intent"
+description: "Detects DMARC failures and messages with a high confidence of credential theft"
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(distinct(headers.hops, .authentication_results.dmarc is not null), strings.ilike(.authentication_results.dmarc, "*fail"))
```
jkamdjou, line 7:
```suggestion
  and any(distinct(headers.hops, .authentication_results.dmarc is not null),
    strings.ilike(.authentication_results.dmarc, "*fail")
  )
```
wrapping my head around this (new line makes it a bit easier to read) - why do we need `distinct` here again?

detection-rules/impersonation_dmarc_failure_high_cred_phish_conclusion.yml
```diff
@@ -0,0 +1,16 @@
...
+  and any([body.plain.raw, body.html.inner_text],
```
jkamdjou, line 8: do we need `coalesce` here or nah?

detection-rules/impersonation_docusign.yml
```diff
@@ -34,8 +34,11 @@ source: |
     and .href_url.domain.root_domain not in ("docusign.com","docusign.net")
   )
 
-  and sender.email.domain.root_domain not in ('docusign.net', 'docusign.com')
-
+  and (
+        sender.email.domain.root_domain not in ('docusign.net', 'docusign.com')
+      or any(distinct(headers.hops, .authentication_results.dmarc is not null),
+        strings.ilike(.authentication_results.dmarc, "*fail"))
```
jkamdjou, line 40:
```suggestion
        sender.email.domain.root_domain not in ('docusign.net', 'docusign.com')
        // check for DMARC fail for spoofs
        or any(distinct(headers.hops, .authentication_results.dmarc is not null),
            strings.ilike(.authentication_results.dmarc, "*fail")
        )
```

detection-rules/impersonation_dmarc_failure_high_cred_phish_conclusion.yml
morriscode: In theory distinct will return the first occurrence, and since we have them in a relatively structured order we're assuming that that would be the authoritative answer.

jkamdjou (approved):

PR #321, opened 2023-03-17
Update Rule: Brand impersonation: Facebook

Jeffreyhung:
facebookenterprise.com is a domain owned by Meta that they use for enterprise related communications.
Whitelisting it in the brand protection rule for Facebook to prevent it from triggering false alert.

https://www.whois.com/whois/facebookenterprise.com
Screenshot 2023-03-17 at 2 14 05 PM


jkamdjou (approved):
thank you @Jeffreyhung !

PR #323, opened 2023-03-21
Updated HTML Unsolicited rules to exclude proofpoint encryption

morriscode:


morriscode:
I'm closing this because it hasn't moved in 3 months.

morriscode:
We could add these encrypted emails to exclusion lists.

PR #324, opened 2023-03-21
Update Rule: Free subdomain link with login or captcha (first-time sender)

morriscode:
Updating rule to negate Zendesk and Zendesk Mailer to reduce FP's.

### DIFF
--- a/detection-rules/link_login_or_captcha.yml
+++ b/detection-rules/link_login_or_captcha.yml
@@ -5,6 +5,10 @@ type: "rule"
 severity: "high"
 source: |
   type.inbound 
+  and (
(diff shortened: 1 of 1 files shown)


detection-rules/link_login_or_captcha.yml
```diff
@@ -5,6 +5,10 @@ type: "rule"
 severity: "high"
 source: |
   type.inbound 
+  and (
+    not strings.like(headers.mailer, "Zendesk Mailer")
+    and not sender.email.domain.root_domain =="zendesk.com"
```
jkamdjou, line 10:
```suggestion
    and not sender.email.domain.root_domain == "zendesk.com"
```

jkamdjou (approved):

PR #338, opened 2023-03-28
New Rule (Attachment): URL file with UNC path

aidenmitchell:


morriscode (approved):
❤️  Matching my samples now! Nice work!

PR #341, opened 2023-03-31
Update Rule: Add % character to normal UNC regex

aidenmitchell:


aidenmitchell:
@morriscode just a quick fix to #338

morriscode (approved):
👍

PR #342, opened 2023-03-31
Create open_redirect_youtube.yml

vector-sec:
Does what it says on the tin


morriscode (approved):
Thanks @vector-sec this is great!!

PR #346, opened 2023-04-03
New Rule (Attachment): Exploitable file type with UNC path

aidenmitchell:
~~This is incomplete, as the regex is throwing some FPs, but just getting it out there.~~

EDIT: good to go!

### DIFF
--- a/detection-rules/attachment_file_with_unc_path.yml
+++ b/detection-rules/attachment_file_with_unc_path.yml
@@ -9,7 +9,7 @@ source: |
   type.inbound 
   and any(attachments, .file_extension in ("asx", "doc", "docx", "jnlp", "m3u", "wax", "xls", "xlsx", "xml")
       and any(file.explode(.), 
-          any(.scan.strings.strings, regex.icontains(., '[\\\\]([a-z0-9_%.$-]+)[\\|/]([a-z0-9_%.$-]+)') 
+          any(.scan.strings.strings, regex.icontains(., '\\\\([a-z0-9_%.$-]+)[\\|/]([a-z0-9_%.$-]+)') 
               and (regex.icontains(., '(?:[0-9]{1,3}\.){3}[0-9]{1,3}')
               // normal UNC paths with IP or domain
               or regex.icontains(., '([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}')))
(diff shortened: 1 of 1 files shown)


aidenmitchell:
TL;DR: strings like `\\2\u003cG1M` and `\\Ie\u003eU` are being flagged by the regex. While these are _technically_ valid UNC paths, you can tell they aren't by looking at them.

Refining the regex and/or adding some additional detection logic will be required to avoid this, or
omitting PDF files from this rule, as these strings seem to be unique to them. Worth noting, for the NTLM exploit to work using a PDF, the user both needs to be using Adobe Acrobat AND click through an external content warning.

aidenmitchell:
Should be good to go!

detection-rules/attachment_file_with_unc_path.yml
```diff
@@ -0,0 +1,18 @@
...
+type: "rule"
+severity: "low"
+source: |
+  type.inbound 
+  and any(attachments, .file_extension in ("asx", "doc", "docx", "jnlp", "m3u", "pdf", "wax", "xls", "xlsx", "xml")
+      and any(file.explode(.), 
+          any(.scan.strings.strings, regex.icontains(., '\\\\([a-z0-9_%.$-]+)\\([a-z0-9_%.$-]+)') and (regex.icontains(., '(?:[0-9]{1,3}\.){3}[0-9]{1,3}') or regex.icontains(., '([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}'))) // normal UNC paths with IP or domain
+          or any(.scan.strings.strings, regex.icontains(., '%5C%5C([a-z0-9_%.$-]+)%5C([a-z0-9_%.$-]+)')) // percent-encoded UNC paths
+      )
+  )
```
morriscode, line 15:
```suggestion
  type.inbound 
  and any(attachments, .file_extension in ("asx", "doc", "docx", "jnlp", "m3u", "pdf", "wax", "xls", "xlsx", "xml")
      and any(file.explode(.), 
          any(.scan.strings.strings, regex.icontains(., '\\\\([a-z0-9_%.$-]+)\\([a-z0-9_%.$-]+)') 
              and (regex.icontains(., '(?:[0-9]{1,3}\.){3}[0-9]{1,3}')
              // normal UNC paths with IP or domain
              or regex.icontains(., '([a-z0-9]+(-[a-z0-9]+)*\.)+[a-z]{2,}')))
              // percent-encoded UNC paths
          or any(.scan.strings.strings, regex.icontains(., '%5C%5C([a-z0-9_%.$-]+)%5C([a-z0-9_%.$-]+)')) 
      )
  )
```

morriscode: Tried to update the formatting for readability, moved the comments before the lines they were decorating. I don't think I broke anything but please validate :)

aidenmitchell:
thank you @morriscode! LGTM!

morriscode:
Retroing

morriscode:
Hit a couple benign files in testing. @aidenmitchell

PR #349, opened 2023-04-06
New Rule: Attachment: Hyperlinked false attachment image lure

morriscode:


morriscode:
I like this rule, but it's matching at a pretty low rate. Might be best to pair with NLU or other logic, closing for now.

PR #353, opened 2023-04-12
Updating UNC path rule to include LNK files

aidenmitchell:


morriscode (approved):
LTGM! no fp's in testing.

PR #361, opened 2023-04-18
New Rule (Link): PDF file with embedded content

aidenmitchell:
Thank you to @ajpc500 for the original logic! This is the link equivalent to the attachment rule: https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/attachment_pdf_file_with_embedded_content.yml


aidenmitchell:
This rule extends beyond Qakbot detection IMO, so I'd like to keep it around. Will look into expanding on the body detection to avoid scanning all URLs.

PR #362, opened 2023-04-18
New Rule (Link): Archive with pdf, txt and wsf files

aidenmitchell:


aidenmitchell:
Right… I forgot about that. I’ll do some looking into what the emails look like with this technique, maybe there’s some pattern there. It’s also possible that Qakbot doesn’t ever use links, in which case this is unnecessary.

aidenmitchell:
Closing, no evidence to suggest Qakbot uses links in body. Will also be re-evaluating other link PRs.

PR #364, opened 2023-04-19
New Rule: Body HTML: HTML smuggling with atob in message body

morriscode:


morriscode:
@jkamdjou Feel free to rename if you can think of a smoother sounding rule name. :) This is a recent deviation from the group that was sending the icloud HTML cred phish campaigns a few weeks ago. They're still firing the Zimbra mailer, but decided to stuff their <script> blocks in the body.html.raw.  Rule fires conclusively with no FP's in testing.

detection-rules/body_html_smuggling_atob_in_body.yml
```diff
@@ -0,0 +1,13 @@
+name: "Body HTML: HTML smuggling with atob in message body"
+description: | 
+  Detects if the email body HTML contains the document write or insertAdjacentHTML method and atob function call. 
+  This technique has been observed leading to credential phishing. 
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and (length(body.plain.raw) < 200 or body.plain.raw is null)
+  and strings.ilike(body.html.raw, "*document*write*atob*", "*document*insertAdjacentHTML*atob*")
```
jkamdjou, line 10: this will match _either_ one of these strings, doesn't have to be both -- just making sure that's intended

jkamdjou (approved):

PR #369, opened 2023-04-20
Update Rule: Docusign "via" in sender display name negation

morriscode:
Added logic to prevent emails sent "via Docusign" from firing by ensuring that if the sender display name contains via,  that a hop field exists for X-Api-Host and ends with a value of docusign.net.

I have only observed docusign.net in this value but it's possible we may have to update to include docusigns other tlds.


aidenmitchell:
Tested this with a new Docusign account, LGTM.
https://analyzer.sublime.security/?id=c618fa19-4be7-4aa6-b055-6b21959cb408

morriscode:
Had a few more test messages come in and feel confident in this logic. Moving to review

bfilar (approved):
💪

PR #378, opened 2023-04-26
Updating Rule: extending char limit and reordering tuples

morriscode:


morriscode:
-  Added "bmp" to file_extensions
-  Restructured tuples 
-  Increased char limit

 No FP's were found in testing this retroactively.  2 previous FN's are now properly detected.  3 previous FP's are now properly ignored.

detection-rules/attachment_office365_image.yml
```diff
@@ -7,18 +7,18 @@ authors:
...
-                  any(.scan.strings.strings, strings.ilike(., "*password*", "*expire*", "*expiry*"))
+                  any(.scan.strings.strings, strings.ilike(., "*password*", "*microsoft*")) and
+                  any(.scan.strings.strings, strings.ilike(.,  "*expire*", "*expiry*","*office365*", "*office*"))
               )
           or
               (
-                  strings.ilike(.scan.ocr.raw, "*microsoft*", "*office365*", "*office*") and
-                  strings.ilike(.scan.ocr.raw, "*password*", "*expire*", "*expiry*")
+                  strings.ilike(.scan.ocr.raw, "*microsoft*", "*password*") and
+                  strings.ilike(.scan.ocr.raw,  "*office365*", "*office*", "*expire*", "*expiry*")
```
rw-access, line 21: this right? `(microsoft or password) and (office or expire or expiry)`
also `office365` here is redundant with `office`

morriscode: Yes that's correct, the way they were split up before wasn't ensuring that "microsoft" or "password" was present and was firing on images containing "office" and "expire".  Requiring either 'Microsoft' or 'password' in addition to one of the others removed the FP's and more directly captured the intent imo.

Nice catch on the redundancy, since it was a community authored rule and we adopted that redundancy I just left it in. But I'll update it tomorrow!! Thanks!!! 👍

rw-access: so if you have a JPG with the string `microsoft office` in the strings, it'll flag? That sounds like an FP storm

morriscode: I thought it would be as well, it performs well in our rules repo, and other instances without FP's. That said our test environment can be biased.  I'm happy to redo the rule altogether,  folks were asking to loosen it as it was missing detectable opportunities like:

image

I originally had it split like this, but it was still missing the FN's. I'll revisit tomorrow! Appreciate the eyes!
`
        (
            (   any(.scan.strings.strings, strings.ilike(., "*password*", "*microsoft*")) and
                any(.scan.strings.strings, strings.ilike(.,  "*office365*", "*office*")) and
                any(.scan.strings.strings, strings.ilike(.,  "*expire*", "*expiry*"))
            )
        or
            (
                strings.ilike(.scan.ocr.raw, "*microsoft*", "*password*") and
                strings.ilike(.scan.ocr.raw,  "*office365*", "*office*") and
                strings.ilike(.scan.ocr.raw, "*expire*", "*expiry*")
            )
        )
      '

morriscode:
@rw-access  Appreciate the pushback here, I rebuilt the rule completely and ended up identifying over 100 more FN's we were missing. Going to close this PR and will resubmit.

PR #379, opened 2023-04-28
Updating Rule: Adding detection for zip download via php webpage

aidenmitchell:
Updating rule to include new Qakbot second stage method. Waiting for an update on LA so that we can confirm this works before merging.


aidenmitchell:
Works! https://sandbox.sublimesecurity.com?id=9cb95b55-0628-48da-82f0-7b96c0128d73

morriscode (approved):
Nice addition to this rule!! thanks!

PR #384, opened 2023-05-01
New Rule (Attachment): PowerPoint with suspicious hyperlink

aidenmitchell:


morriscode (approved):
This has been in testing for a while. No concerns over FP's.

PR #389, opened 2023-05-04
Updating Rule: Docusign Image with sus link incorporating customer feedback

morriscode:
- Switching from inner_text length to attachment ==1 
- Including bmp in file extensions

___________________

155 messages found  (New Rule)
3 Already Flagged. (Old Rule)

No FP's.


morriscode:
- Switching from inner_text length to attachment ==1 
- Including bmp in file extensions

___________________

155 messages found  (New Rule)
3 Already Flagged. (Old Rule)

No FP's.

PR #391, opened 2023-05-04
Flatten multi_basic_queries to individual files

cameron-dunn-sublime:
Keeping this simplest and dropping support for multiple `queries` in a single file.


morriscode (approved):
LGTM! 👍

PR #403, opened 2023-05-12
Add WIP workflow for updating test rules

cameron-dunn-sublime:
This is just the first steps -- triggering the job, checkout out the branch, and validating that the basic validation workflow completed successfully.

I've validated these steps, but need to figure out some behavior related to forks. To do that testing I'm bringing this int sublime-rules so I can fork and test.

### DIFF
--- a/.github/workflows/update-test-rules.yml
+++ b/.github/workflows/update-test-rules.yml
@@ -14,7 +14,7 @@ jobs:
 
     steps:
       - name: Get PR branch
-        uses: xt0rted/pull-request-comment-branch@v1
+        uses: xt0rted/pull-request-comment-branch@v2
         id: comment-branch
 
       - name: Checkout PR branch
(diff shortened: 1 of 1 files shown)


.github/workflows/update-test-rules.yml
```diff
@@ -0,0 +1,31 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  update-branch:
+    name: Update test-rules Branch
+    runs-on: ubuntu-20.04
+    # https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/update=test-rules')
```
rw-access, line 12: Probably need to figure out how to restrict this to the sublime-security org

rw-access (approved):

rw-access (approved):

.github/workflows/update-test-rules.yml
cameron-dunn-sublime: That'll happen through an environment!

PR #406, opened 2023-05-15
Finish out workflows for test-rules

cameron-dunn-sublime:
Adds workflows for adding to and cleaning the `test-rules` branch. Additional context in "Initiating Rule Signal Collection GitHub Action" task.

I've tested this a bunch in a separate repo, but nothing is looking at the `test-rules` branch so we can continue to adjust without impact if there are any issues.

The basics of how this works:
* Any sublime org member can comment `/update-test-rules` on an open PR.
* The PR workflow will require approval by someone in the `test-rules` environment (me, @rw-access, @morriscode , @jkamdjou )
* The `test-rules` branch will be updated with any new or changed rules from the PR.
    * If the same rule file is being tested in another PR, it will be overwritten.
* When the PR is closed (merged or Close PR button), any rules being tested from the PR will be removed from the `test-rules` branch.

All in all it should be good for what we need now, and we can adjust as needed in the future. Potential situations I could see us expanding for:
* Remove rules from `test-rules` branch via comment.
* Allow parallel testing of multiple changes on the same file, or at least confirm the overwrite.

### DIFF
--- a/.github/workflows/clear-test-rules.yml
+++ b/.github/workflows/clear-test-rules.yml
@@ -1,0 +1,50 @@
+name: Remove from test-rules Branch
+
+on:
+  pull_request:
+    types: [ closed ]
+
+jobs:
+  remove-fromt-test-branch:
+    name: Remove from test-rules Branch
+    permissions:
+      contents: write
+      issues: read
+      pull-requests: read
+    runs-on: ubuntu-20.04
+    concurrency: test-rules
+
+    steps:
+      - name: Install yq
+        uses: mikefarah/yq@v4.33.3
+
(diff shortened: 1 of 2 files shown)


.github/workflows/clear-test-rules.yml
```diff
@@ -0,0 +1,50 @@
+name: Remove from test-rules Branch
+
+on:
+  pull_request:
+    types: [ closed ]
+
+jobs:
+  remove-fromt-test-branch:
```
itsojon, line 8:
```suggestion
  remove-from-test-branch:
```

itsojon (approved):

PR #421, opened 2023-05-19
New Rule: Link fake thread with suspicious link and NLU financial request

morriscode:


detection-rules/link_fake_thread_nlu_financial_request.yml
```diff
@@ -0,0 +1,46 @@
+name: "Link: Fake message thread with suspicious links and NLU financial request"
```
jkamdjou, line 1:
```suggestion
name: "Fake message thread with a suspicious link and engaging language from an unknown sender"
```

morriscode: 💍

jkamdjou (approved):

jkamdjou (approved):

PR #422, opened 2023-05-19
Simplify docusign logic

rw-access:
The `length(filter(x)) == 0` is the same thing as `not any(x)`. If we also require that `file_type` is not null, then we can go a step further: `all(not x)`

All these optimizations happen under the hood anyway, so the rule is still logically identical. This seems simpler to me though, and I think it's more readable (although this might be triggering for anyone with some `all` trauma). We can add a `length` check if we'd like but it's already combined with the `any` so it's redundant

### DIFF
--- a/detection-rules/attachment_docusign_image_suspicious_links.yml
+++ b/detection-rules/attachment_docusign_image_suspicious_links.yml
@@ -6,7 +6,7 @@ source: |
   type.inbound
   and all(attachments, .file_type in~ ('png', 'jpeg', 'jpg', 'bmp'))
   and any(body.links, not strings.ilike(.href_url.domain.root_domain, "docusign.*"))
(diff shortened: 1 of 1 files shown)


morriscode:
😓 anxiety building.... In theory we can do the same thing with this [rule](https://github.com/sublime-security/sublime-rules/blob/767f396e4b27a71d61f78113c742ba7c004595c6/detection-rules/attachment_office365_image.yml#L4) 

I'll kick off some hunt comparisons just so I can sleep at night.

rw-access:
Maybe it's that null case?
`and length(filter(attachments, .file_type not in~ ('png', 'jpeg', 'jpg', 'bmp'))) == 0` will ignore null `.file_type` but my change won't run if there are any null `.file_type.` I'll go to `not any` then

morriscode:
This hurts my 🧠

PR #426, opened 2023-05-19
New Rule: Link: Microsoft impersonation using MS hosted png

morriscode:


detection-rules/link_microsoft_impersonation_using_hosted_png.yml
```diff
@@ -0,0 +1,41 @@
+name: "Link: Microsoft impersonation using hosted png with suspicious link"
+description: "Detects emails containing a link to a microsoft hosted logo and the display name and displayurl text is in all caps
+  and an NLU request from first time sender."
```
jkamdjou, line 3:
```suggestion
description: |
  Detects messages with a link to a Microsoft hosted logo where the sender's display name
  and the display text of a link in the body are in all caps, and a request is being made from a first-time sender.
```

jkamdjou (approved):

PR #427, opened 2023-05-22
New Rule: Dropbox image lure with no dropbox links

morriscode:
No FP's in testing.


detection-rules/attachment_dropbox_image_suspicious_links.yml
```diff
@@ -0,0 +1,30 @@
+name: "Attachment: Dropbox image lure with no Dropbox domains in links"
+description: "Detects Dropbox phishing emails with image attachments and short message bodies
```
jkamdjou, line 2: we aren't looking for "short message bodies" are we?

jkamdjou (approved):

PR #431, opened 2023-05-23
New Rule: URLhaus: Malicious domain in body or pdf

morriscode:


detection-rules/urlhaus_malicious_dom_in_body_or_pdf.yml
```diff
@@ -0,0 +1,50 @@
+name: "URLhaus: Malicious domain in message body or pdf attachment (trusted reporters)"
+description: "Detects URLhaus domains submitted by trusted reporters in message bodies or pdf attachments"
+type: "query"
```
jkamdjou, line 3:
```suggestion
type: "rule"
```

detection-rules/urlhaus_malicious_dom_in_body_or_pdf.yml
```diff
@@ -0,0 +1,50 @@
...
+
+                  .domain.domain in $abuse_ch_urlhaus_domains_trusted_reporters
+              )
+          )
+      )
+  )
+tags:
+  - "Suspicious links"
+  - "Suspicious attachments"
+  - "URLhaus"
```
jkamdjou, line 50:
```suggestion
  - "URLhaus"
  - "Abusech"
```

morriscode: URLhaus is right... I've seen in text as abusech, AbuseCh, abuseCH, their logo has it as ABUSEch... but I think what you have is fine. imo.

jkamdjou (approved):

PR #448, opened 2023-05-25
New Rule: BEC: Freemail Reply-to or Return-path mismatch with NLU BEC

morriscode:


bfilar (approved):

detection-rules/headers_freemail_replyto_returnpath_mismatch.yml
```diff
@@ -0,0 +1,25 @@
+name: "BEC: Freemail Reply-to or Return-path sender mismatch and NLU BEC intent"
+description: "Detects messages with a NLU BEC intent with medium or high confidence,
+  validates reply-to domains or return path domain are not free email providers
+  and the sender domain is not freemail"
```
jkamdjou, line 4:
```suggestion
description: "Detects an unusual header mismatch where the sender is not a freemail address, but the reply-to or return-path are. NLU also detects a BEC intent with medium or high confidence."
```

jkamdjou (approved):

PR #450, opened 2023-05-25
New Rule: Attachment: Adobe image lure with suspicious link from first time sender

morriscode:
284 messages found ,2 Already Flagged. All bad. No FP's in testing.

### DIFF
--- a/detection-rules/attachment_adobe_image_lure_fts.yml
+++ b/detection-rules/attachment_adobe_image_lure_fts.yml
@@ -1,0 +1,32 @@
+name: "Attachment: Adobe image lure with with suspicious link from first time sender"
+description: "Detects Adobe phishing emails with an Adobe logo attached, with suspicious link language
+  from a first-time sender."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and length(filter(attachments, .file_type not in~ ('png', 'jpeg', 'jpg', 'bmp'))) == 0
+  and length(body.links) >0
+  and all(body.links, .display_text is null)
+  and any(attachments,
+       beta.logo_detect(.).brand.confidence in ("high")
+      and beta.logo_detect(.).brand.name == "Adobe"
+      and any(file.explode(.),
+            any(.scan.strings.strings, strings.ilike(.,
+                "*review*", "*sign*", "*view*", "*completed document*", "*open agreement*"))
+      )
+  )
+  and (
+          (
+              sender.email.domain.root_domain in $free_email_providers
+              and sender.email.email not in $sender_emails
+          )
+          or (
+              sender.email.domain.root_domain not in $free_email_providers
+             and sender.email.domain.domain not in $sender_domains
+          )
+  )
+tags:
+  - "Suspicious link"
+  - "Brand impersonation"
+  - "Computer Vision"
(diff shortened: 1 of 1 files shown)


cameron-dunn-sublime:
/update-test-rules

Test message

detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -0,0 +1,32 @@
+name: "Attachment: Adobe image lure with with suspicious link from first time sender"
```
jkamdjou, line 1: do we want to add a check for a suspicious link? right now we're just looking for phrases and existence of a link

detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -0,0 +1,32 @@
...
+description: "Detects Adobe phishing messages with an Adobe logo attached, with suspicious link language
+  from a first-time sender."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and length(filter(attachments, .file_type not in~ ('png', 'jpeg', 'jpg', 'bmp'))) == 0
+  and length(body.links) >0
+  and all(body.links, .display_text is null)
```
jkamdjou, line 10: this is surprising. would have thought the phrases below would be on the display text

detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -0,0 +1,32 @@
...
+  and any(attachments,
+       beta.logo_detect(.).brand.confidence in ("high")
```
jkamdjou, line 12:
```suggestion
      beta.logo_detect(.).brand.confidence in ("high")
```

detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -0,0 +1,32 @@
...
+  and (
+          (
+              sender.email.domain.root_domain in $free_email_providers
+              and sender.email.email not in $sender_emails
+          )
+          or (
+              sender.email.domain.root_domain not in $free_email_providers
+             and sender.email.domain.domain not in $sender_domains
+          )
+  )
```
jkamdjou, line 28: should we exclude emails from Adobe?

detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -0,0 +1,32 @@
+name: "Attachment: Adobe image lure with with suspicious link from first time sender"
+description: "Detects Adobe phishing messages with an Adobe logo attached, with suspicious link language
+  from a first-time sender."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and length(filter(attachments, .file_type not in~ ('png', 'jpeg', 'jpg', 'bmp'))) == 0
```
jkamdjou, line 8: we could totally add another version of this rule looking for branded PDFs now too 😎

detection-rules/attachment_adobe_image_lure_fts.yml
morriscode: we don't check other rules with similar intent.. eg DocuSign. Testing validates it's uneeded.

detection-rules/attachment_adobe_image_lure_fts.yml
morriscode: The image is hyperlinked so no need for display text

detection-rules/attachment_adobe_image_lure_fts.yml
morriscode: Doesn't seem to be required. Adobe isn't really notorious for unsolicited spam, and legit emails won't match this criteria.

jkamdjou (approved):

PR #457, opened 2023-05-26
make fake thread detection even yet broader

jkamdjou:


morriscode (approved):
Looks good. Reminder to lower Message Thread in "Fake Message Thread" in later PR

PR #458, opened 2023-05-26
New Rule: Wise impersonation with suspicious attachment

aidenmitchell:


aidenmitchell:
Running a hunt now, `*wise*` might be too broad.

morriscode:
Edited rule to use coalesce, queuing  test rules again

morriscode (approved):
Rule has been in testing for a while with no FP's or TP's. Should be safe to deploy.

detection-rules/impersonation_wise_suspicious_attachment.yml
```diff
@@ -0,0 +1,29 @@
...
+  Impersonating Wise Financial, as confirmed by NLU, and contains a suspicious attachment type.
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and (
+          strings.ilike(sender.display_name, '*wise*')
+          or strings.ilike(sender.email.domain.domain, '*wise*')
+      )
+  and any(ml.nlu_classifier(coalesce(body.html.display_text, body.plain.raw)).tags, .name in ("payment", "invoice"))
```
jkamdjou, line 12: @morriscode  we can update this to current thread

detection-rules/impersonation_wise_suspicious_attachment.yml
```diff
@@ -0,0 +1,29 @@
...
+  and any(attachments, .file_type 
+      in~ $file_extensions_common_archives
+      or .file_type in~ $file_extensions_executables
+      or .file_type in~ $file_extensions_macros
```
jkamdjou, line 16: file type is already guaranteed to be lowercase

detection-rules/impersonation_wise_suspicious_attachment.yml
morriscode: ahh for sure!

PR #459, opened 2023-05-26
Updating Rule (Callback image): Adding new strings

aidenmitchell:


morriscode (approved):
Looks good makes sense to me.

PR #470, opened 2023-05-30
New Rule (Brand spoof): Dropbox

aidenmitchell:


aidenmitchell:
FTS causes issues, due to the fact that the impersonation we're seeing is "coming" from the legitimate Dropbox sender address (no-reply@dropbox.com)

morriscode (approved):

PR #473, opened 2023-05-30
New-Rule: BEC masked recipients with no links and freemail reply-to mismatch

morriscode:


detection-rules/headers_bec_masked_recipients_no_links_freemail_replyto.yml
```diff
@@ -0,0 +1,33 @@
+name: "BEC - Masked Recipients with freemail reply-to and no links"
```
jkamdjou, line 1:
```suggestion
name: "Business Email Compromise: Masked recipients with suspicious reply-to mismatch (unsolicited)"
```

detection-rules/headers_bec_masked_recipients_no_links_freemail_replyto.yml
```diff
@@ -0,0 +1,33 @@
...
+description: |
+  This rule detects Unsolicited messages where the recipient matches the sender address and no other recipients are identified. And the reply-to address does not match the sender, and is freemail with no links in the body.
+  This a common combination of techniques used by low level BEC threats.
```
jkamdjou, line 4:
```suggestion
  This rule detects unsolicited messages where the recipient matches the sender address and no other recipients are identified.
  The reply-to address does not match the sender, and is a freemail with no links in the body.
  This a common combination of techniques used by low level BEC threats. 
```

detection-rules/headers_bec_masked_recipients_no_links_freemail_replyto.yml
```diff
@@ -0,0 +1,33 @@
...
+              and sender.email.email not in $recipient_emails
+          )
+          or (
+              sender.email.domain.root_domain not in $free_email_providers
+              and sender.email.domain.domain not in $recipient_domains
+          )
+  )
+tags: 
+  - "Suspicious sender"
+  - "Suspicious ReplyTo"
```
jkamdjou, line 33:
```suggestion
  - "Suspicious headers"
  - "Business Email Compromise"
```

jkamdjou (approved):

PR #477, opened 2023-05-31
New Rule: Suspicious sender name and content

aidenmitchell:


morriscode:
@aidenmitchell ping me when you're around and we can review the test rules on this.

PR #480, opened 2023-06-01
New Rule: Brand Spoof: UPS

morriscode:


jkamdjou (commented):
do we need to account for any of the auth failures that could result from the SEG like we saw with DocuSign before merging this?

also do we want to include a Reference for the recent Gmail issue for context?

aidenmitchell (approved):

jkamdjou:
stale, closing but we can re-open if there's a use case

PR #482, opened 2023-06-01
New Rule (Brand impersonation): Gusto

aidenmitchell:


morriscode:
Volume on this is a touch higher than I'd anticipate. @aidenmitchell hmu when you're around and we can review.

morriscode:
Going to close this and replace with logodetect

PR #487, opened 2023-06-02
New Rule (Attachment): HTML smuggling with auto-downloaded file

aidenmitchell:


morriscode (approved):
The corresponding actions taken when reviewing test result look good. Overlap with other HTML smuggling rule also. Approving.

PR #491, opened 2023-06-05
Updating rule: Fixing attachment logic and adding logo_detect

aidenmitchell:


morriscode:
Retroing now!

morriscode (approved):
new messages flagged! Not seeing any FP's. Thanks!!

PR #495, opened 2023-06-06
Updating rule: attachment_office365_image.yml

aidenmitchell:
Missing some detecting using strings, added OCR check as well.


morriscode (approved):
Retroing now

PR #496, opened 2023-06-06
New Rule: Link: Left-to-right mark (LRM) in base tag

morriscode:


jkamdjou (approved):
@morriscode do we def want to call this an exploit?

detection-rules/link_outlook_left_to_right_exploit.yml
```diff
@@ -0,0 +1,13 @@
+name: "Link: Left-to-right mark (LRM) in base tag"
+description: |
+  The exploit involves tricking Outlook for Windows into displaying a fake domain while opening another one. This is achieved by adding a <base> HTML tag with a fake domain and a left-to-right mark (Unicode U+200E). Links within <a> tags will display the fake domain but open the actual domain when clicked on.
+references:
+  - "https://twitter.com/ldionmarcil/status/1665732725767122946?s=20"
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound 
+  and regex.contains(body.html.raw, 'base.{0,100}‎/>')
```
jkamdjou, line 10:
```suggestion
  // this regex contains a Unicode U+200E character (may not be visible in many clients)
  and regex.contains(body.html.raw, 'base.{0,100}‎/>')
```

PR #500, opened 2023-06-07
New Rule: Stripe invoice abuse

morriscode:


morriscode:
> do we care to look for the existence of a phone number or is that in all of these? if the latter, are there 2 numbers or just the 1?

Originally I had the phone number check in there, but OCR was not seeing it at all, and it's not in strings either :/

PR #501, opened 2023-06-08
First batch of NLU signals

bfilar:


signals/content/nlu_bec_high_conf.yml
```diff
@@ -0,0 +1,8 @@
+name: "NLU Intent: Business Email Compromise"
```
jkamdjou, line 1: do u want the confidence in the name like the others? missing on a few of em

signals/content/nlu_bec_high_conf.yml
```diff
@@ -0,0 +1,8 @@
...
+type: "query"
+source: |
+  any(ml.nlu_classifier(coalesce(body.html.display_text, body.plain.raw)).intents, .name=="bec" and .confidence == "high")
+
```
jkamdjou, line 5: extra new line here and in some others

jkamdjou (commented):
l f g

signals/content/nlu_bec_high_conf.yml
bfilar: yeah good call.

bfilar:
Ready for review

morriscode:
These look good to me. The only thing I question is the use of Tags. None of the other signals have any tags on them, I'm not sure if those actually do anything, or if they're helpful in your model, or even if they work on signals, I assumed they would as it's probably the same schema, but I don't know.

bfilar:
that's a fair point. i should probably add them

bfilar:
Oh I see what you're saying. Let's get them in as they could be useful indicators when observed with other signals

bfilar:
Based on internal testing from @rw-access this should be good to merge. Someone review & give a 👍

rw-access:
FWIW this PR wasn't blocked before testing and isn't now, so don't let me hold you up!

morriscode (approved):
:shipit:

PR #505, opened 2023-06-08

vector-sec:
Excluded commonly seen root domains in legitimate Onedrive emails from tranco_1m check


jkamdjou:
@vector-sec @morriscode  i moved the exclusion slightly down, just in case any of the domains make their way into any of those lists in the future (sharepoint used to be in $free_subdomain_hosts until recently)

morriscode (approved):
👍

PR #506, opened 2023-06-09
Switch from logo_detect().brand to brands

rw-access:
We're updating `logo_detect` to always return `brands` and stop returning `brand`.


morriscode (approved):
✔️

PR #507, opened 2023-06-09
New Rule: Suspicious Request for quote or purchase

morriscode:
I'm not married to the name


detection-rules/suspicious_request_for_quote_or_purchase.yml
```diff
@@ -0,0 +1,69 @@
...
+          (
+              sender.email.domain.root_domain in $free_email_providers
+              and sender.email.email not in $recipient_emails
+          )
+          or 
+          (
+              sender.email.domain.root_domain not in $free_email_providers
+              and sender.email.domain.domain not in $recipient_domains
+          )
+      )
```
jkamdjou, line 33: may cause FPs since it's 1 of

morriscode: I think you're right... Pulling unsolicited increases accuracy

aidenmitchell (approved):

detection-rules/suspicious_request_for_quote_or_purchase.yml
```diff
@@ -0,0 +1,58 @@
...
+            )
+            and length(recipients.cc) == 0
+            and length(recipients.bcc) == 0
+        ),
+        (
+            sender.email.domain.root_domain in $free_email_providers
+            and any(headers.reply_to, .email.email != sender.email.email)
+            and any(headers.reply_to, .email.email not in $recipient_emails)
+        )
+  )
```
jkamdjou, line 23: this snippet is 🔥  just heads up that this doesn't mean there's necessarily a suspicious sender (in the rule name)

detection-rules/suspicious_request_for_quote_or_purchase.yml
```diff
@@ -0,0 +1,58 @@
+name: "Request for Quote or Purchase (RFQ|RFP) from suspicious sender"
```
jkamdjou, line 1:
```suggestion
name: "Request for Quote or Purchase (RFQ|RFP) with suspicious sender or recipient pattern"
```

detection-rules/suspicious_request_for_quote_or_purchase.yml
morriscode: Love the name change ty ty ty!

morriscode:
New batch of test show high accuracy and overlap with other confidence rules. This is GTG

morriscode:
I think i'm seeing a null value in the test data for this, going to close the PR and reopen and run test-rules again.

jkamdjou:
> I think i'm seeing a null value in the test data for this, going to close the PR and reopen and run test-rules again.

copy, lmk when you're ready for a re-review

jkamdjou:
@morriscode it looks like it's been running: https://github.com/sublime-security/sublime-rules/blob/test-rules/detection-rules/suspicious_request_for_quote_or_purchase.yml

morriscode:
> @morriscode it looks like it's been running: https://github.com/sublime-security/sublime-rules/blob/test-rules/detection-rules/suspicious_request_for_quote_or_purchase.yml

It's been running but returned 3 values for the sha. The original, a null which looks like it was getting populated more recently and a third sha with no hits. 

To err on the side of caution I just wanted to reinitiate it. I'm wonder if we close a PR but don't wait for the cleanup job to run before opening it if it actually removes it...

morriscode:
I suspect if we close and reopen a PR without waiting for the remove stale rules job it won't recognize that the PR was close as it's a point in time state check... Gonna close this and test

morriscode:
@jkamdjou I think this is gtg, but it's really hard to discern the test-rule-data in this scenario.

jkamdjou (approved):

PR #516, opened 2023-06-12
Marking Google Translate rule as ASR

aidenmitchell:


morriscode (approved):
This works for me, there are legitimate uses from unsolicited, this is more ASR in practice than we previously thought it would be.

PR #517, opened 2023-06-12
New Rule: Employee impersonation payroll fraud

morriscode:
This supplement our BEC NLU detection for Payroll Fraud.


morriscode:
Testing on this looks to be about what I expected. Removing the $org_display names and validating in Shared Samples, shows high accuracy.  @jkamdjou  Should be ok to go on this, unless you have concerns?

detection-rules/impersonation_employee_payroll_fraud.yml
```diff
@@ -0,0 +1,25 @@
...
+severity: "medium"
+source: |
+  type.inbound 
+  and sender.display_name in $org_display_names
+  and length(attachments) == 0 
+  and length(body.plain.raw) < 300
+  and (
+      sender.email.domain.root_domain not in $org_domains
+      or
+      sender.email.domain.root_domain in $free_email_providers
```
jkamdjou, line 14: could we turn this into [this snippet](https://github.com/sublime-security/sublime-rules/pull/507/files#diff-047d62794629232e9067ab61b70afde24b01d783d597f4b1b35cd1a2c8d0af83R19-R21), or would that miss what you're looking for? am a little concerned about FPs from people's legit personal emails, since we don't have any historical context or other suspicious signals

morriscode: We could but most of these don't bother manipulating reply to. They just straight up send/receive from a external account hoping that people will assume it's a legit personal email.

jkamdjou: gotcha. maybe we could at least include Unsolicited or FTS? or would that not fire either?

morriscode: Yeah either of those in theory should fire, but I was trying to avoid the pitfalls associated. I can toss FTS on it in the interim.

jkamdjou: Unsolicited would at least avoid some of the pitfalls of FTS, and help a bit

morriscode: 💯 Will update!

jkamdjou (approved):

PR #518, opened 2023-06-12
Updating Rule: Marking as ASR

aidenmitchell:


morriscode:
If we add it here we probably want to add it [here](https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/link_new_domain_in_link_first_time_sender.yml) as well? Agree?

morriscode:
@aidenmitchell  ☝️

aidenmitchell:
good idea, added!

morriscode (approved):

PR #523, opened 2023-06-13
Updating Rule: Adjusting HTML detection

aidenmitchell:
Sometimes, HTML files come up as plain text. Added plain text detector to solve this.
image


morriscode (approved):
Nice catch.

PR #524, opened 2023-06-13
New Rule: HTML file with reference to recipient and suspicious patterns

aidenmitchell:


jkamdjou:
nice

morriscode (approved):
👍 Testing looks solid! nice work

PR #527, opened 2023-06-13
Updating Rule: Exempting legitimate Google Maps shortener

aidenmitchell:
This shortener _in this usecase (for Google Maps)_ shouldn't be treated as suspicious.


morriscode (approved):

morriscode:
makes sense to me

PR #529, opened 2023-06-13
Refactoring to use recipient SLD rather than Org SLD

morriscode:


jkamdjou:
@morriscode only seeing a name/description update

morriscode:
> @morriscode only seeing a name/description update

I forgot that last push :)

jkamdjou (approved):

PR #532, opened 2023-06-14
Signals: New Subject and Link based signals

bfilar:


morriscode (commented):
I think @jkamdjou  wanted "description" for all these, I can see the name functioning as a description. So I'm not sure we've rightsized the requirements for signals yet?

bfilar:
Yep. Keeping this in draft until requirements are set.

signals/content/subject_is_upper.yml
```diff
@@ -0,0 +1,4 @@
+name: "Subject: Contains All Capital Letters"
+type: "query"
+source: |
+  regex.match(subject.subject, "^[A-Z]*$")
```
morriscode, line 4:
```suggestion
  regex.match(subject.subject, "^[A-Z ]*$")
```

I think you might want a space in your char set? Otherwise you'll not be able to match things like "TEST SUBJECT"?

signals/links/google_open_redirect.yml
```diff
@@ -0,0 +1,6 @@
+name: "Link: Google Open Redirect"
```
morriscode, line 1:
```suggestion
name: "Subject: Contains All Lowercase Letters""
```

signals/links/google_open_redirect.yml
```diff
@@ -0,0 +1,6 @@
...
+type: "query"
+source: |
+  regex.match(subject.subject, "^[^a-z]*$")
```
morriscode, line 4:
```suggestion
  regex.match(subject.subject, "^[^a-z ]*$")
```
adding space to char set, same as upper letters?

morriscode (commented):
Added some suggestions

signals/content/subject_contains_recipient_addr.yml
```diff
@@ -0,0 +1,9 @@
+name: "Subject: Contains Recipient Email Address"
```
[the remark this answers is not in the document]
bfilar: Yeah definitely. I'll split them out

signals/content/subject_contains_recipient_local_part.yml
```diff
@@ -0,0 +1,4 @@
+name: "Subject: Contains Recipient Local Part"
+type: "query"
+source: |
+  any(recipients.to, strings.icontains(subject.subject, .email.local_part))
```
morriscode, line 4: Without negating an email address pattern match this will fire every time the above rule fires no? 
what we really need is to be able to do like 
`  any(recipients.to, regex.icontains(subject.subject, '.email.local_part[^@]')) but barring that 
`  
Barring that, negating an email address pattern in the subject should prevent the duplicate firing.
`and not regex.icontains(subject.subject, '.+\@.+\..+'))`

Thoughts?

signals/content/subject_suspicious_keywords.yml
```diff
@@ -0,0 +1,195 @@
...
+    "verification( )?-?need",
+    "verify.your?.account",
+    "web ?-?ma[il1]{2}",
+    "web[ -]?ma[il1]{2}",
+    "will.be.suspended",
+    "your (customer )?account .as",
+    "your.office.365",
+    "your.online.access",
+
+    // https://github.com/sublime-security/static-files/blob/master/suspicious_subjects.txt
```
jkamdjou, line 128: we could use $suspicious_subjects now for the non-regex ones

signals/content/subject_is_upper.yml
```diff
@@ -0,0 +1,4 @@
+name: "Subject: Contains All Capital Letters"
+type: "query"
+source: |
+  regex.match(subject.subject, "^[A-Z ]*$")
```
morriscode, line 4: I think we might want to do it this way? 

`^[^a-z]*[A-Z][^a-z]*$`

That way we capture when a subject contains other chars along with ALL CAPS?

like
`#:SCAN ORIGINAL QUOTE SHIPPING DOCS-MSK///2131
`

signals/content/subject_suspicious_keywords.yml
bfilar: Great call

signals/content/subject_contains_recipient_local_part.yml
morriscode: No I'm saying that 
`strings.icontains(subject.subject, .email.local_part`

Would fire on 
`subject: Congrats b_thomas@gmail.com you've won!
recipient.to: b_thomas@gmail.com`

morriscode: Since the local part is contained within the broader email address, you'll always get a duplicate firing of email_local if the full email address in subject. So in the local rule, you'd want to negate if an email pattern is in the subject.

signals/content/subject_is_upper.yml
bfilar: Yeah, that makes sense to me. I wasn't sure if we wanted multiple signals to represent:
- All upper
- Upper w/ symbols
- ... etc

signals/content/subject_contains_recipient_local_part.yml
morriscode: Oh derp yeah that's much smarter.

morriscode (approved):
🚢

signals/content/subject_contains_recipient_local_part.yml
```diff
@@ -0,0 +1,7 @@
+name: "Subject: Contains Recipient Local Part"
+type: "query"
+source: |
+  and any(recipients.to,
```
rw-access, line 4: we need some CI on this repo, this `and` caused issues

PR #533, opened 2023-06-14
Revert "Change validation workflow to run on branches"

rw-access:
Reverts sublime-security/sublime-rules#530

This was running actions in the context of the source repository, but we need to run those in the context of this repository. I think this'll fix some issues we're running to.

And if I'm wrong, we can revert again.


morriscode (approved):
👍

PR #534, opened 2023-06-14
Updating Rule: adding additional signals for detection

aidenmitchell:


aidenmitchell:
Hits a few more TPs in our shared repo and no FPs.

morriscode (approved):
Testing on this looks solid, overlap with other similar content rules, and message actions show a propensity to delete. Approving.

PR #537, opened 2023-06-15
Updating rule: Adding additional NLU signal

aidenmitchell:


morriscode (approved):
Testing looks solid, great addition!

PR #541, opened 2023-06-15
New Rule (Attachment): EML file contains HTML attachment with login portal indicators

aidenmitchell:


morriscode (approved):
We got 48 hits over 2 weeks. I don't see any major concern here.

PR #543, opened 2023-06-15
Update Sharepoint credential phishing rule to use attachments

rw-access:
Our Sharepoint rule uses `beta.message_screenshot()`, but let's pull that out for now while we continue to evaluate the function.


aidenmitchell:
FYSA: This _might_ fail to flag some samples, we've seen some where the logo isn't pulled out as an attachment.
Are we accepting this risk while `beta.message_screenshot()` is being tested?

PR #545, opened 2023-06-17
New rule - Attachment: Embedded Javascript in SVG file (unsolicited)

jkamdjou:
Not sure how common this is in legit email, will def want to test.

Unsolicited should hopefully mitigate some. If we do see FPs, it'll prob be from unsolicited marketing emails pitching some service


jkamdjou:
image

weird. will try again

morriscode:
After reviewing test-rule results these look safe to ship!

PR #546, opened 2023-06-17
New rule - Attachment: Embedded VBScript in MHT file (unsolicited)

jkamdjou:


morriscode:
after reviewing test rule results these look safe to deploy.

morriscode (approved):

PR #547, opened 2023-06-20
New Rule - Brand impersonation: Microsoft (QR code)

morriscode:


detection-rules/attachment_microsoft_image_lure_qr_code.yml
```diff
@@ -0,0 +1,43 @@
...
+          any(file.explode(.),
+              any(.scan.strings.strings, regex.icontains(.,  'scan|camera'))   
+              and any(.scan.strings.strings, regex.icontains(.,  '\bQR\b|Q\.R\.|barcode'))
+          )
+      )
+  )
+  and (
+      not any(headers.hops, .authentication_results.compauth.verdict is not null
+      and .authentication_results.compauth.verdict == "pass"
+      and sender.email.domain.domain == "microsoft.com")
```
jkamdjou, line 26:
```suggestion
      not any(headers.hops,
          .authentication_results.compauth.verdict is not null
          and .authentication_results.compauth.verdict == "pass"
          and sender.email.domain.domain == "microsoft.com"
      )
```

detection-rules/attachment_microsoft_image_lure_qr_code.yml
```diff
@@ -0,0 +1,43 @@
+name: "Attachment: Microsoft image with QR code"
+description: |
+  Detects messages using Microsoft image based lures, referencing a QR code from an Unsolicited sender.
+  These messages often lead users to phishing sites or initiate unwanted downloads upon scanning.
```
jkamdjou, line 4:
```suggestion
  These messages often lead users to phishing sites or initiate unwanted downloads.
```

detection-rules/attachment_microsoft_image_lure_qr_code.yml
rw-access: if only we had... some way... to do this automagically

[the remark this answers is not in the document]
morriscode: I just snagged the snippet from the other MS rule. We can swap it to root domain if you want, it's probably more accurately representative of the goal?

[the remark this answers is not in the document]
jkamdjou: defer to you, i'm good as-is too

PR #548, opened 2023-06-20
Update rules to use .credphish.confidence instead of .brand.confidence

morriscode:
Looks like a couple rules were using .brand.confidence to inspect a credphish.disposition. If we're inspecting a disposition it should be using .confidence, and if we're inspecting a brand it should use .brand.confidence


jkamdjou (approved):
nice find

PR #549, opened 2023-06-20
New Rule: Credential Phishing link with Undisclosed Recipients

morriscode:


morriscode:
No test rules due to linkanalysis usage, but no false flags in testing.  Positive hits in Shared Samples repo, rule should be safe.  Undisclosed recipients carries heavy suspicion.

bfilar (approved):

PR #552, opened 2023-06-21
Updating Rule: using YARA instead of attachments.raw

aidenmitchell:


rw-access (approved):
thanks for being flexible @aidenmitchell!

PR #554, opened 2023-06-21
Signals based on Link metadata

bfilar:


morriscode:
Would a count of links be helpful here?  I'm all about the count today.

bfilar:
Then I'm thinking we'd want to have a signal that is `Link: Total Count` and `Link: Total Unique Count`

signals/links/link_appspot_in_url_path.yml
```diff
@@ -0,0 +1,4 @@
+name: "Link: Appspot in URL Path"
+type: "query"
+source: |
+  length(any(body.links, strings.ilike(.href_url.path, "*appspot.com*")))
```
rw-access, line 4: this isn't valid MQL, you mean `length(filter`

morriscode:
Whew all these tested and validated! Lets go!

morriscode (approved):

PR #556, opened 2023-06-22

vector-sec:
Added OneDrive shortened URL domain


jkamdjou (approved):
thank you @vector-sec !

PR #558, opened 2023-06-22
Initial batch of attachment signals

morriscode:


morriscode:
Just a heads up some of these contain enrichment functions, mostly file explode.

morriscode:
> 

right right! I just wanted to get them up. I'll change this to a draft or I can pluck out the enrichment ones? Any preference?

morriscode:
Should be free of rules with enrichment depends now.

PR #560, opened 2023-06-22
Update rules to use ml.logo_detect

rw-access:
Renaming this function from beta. Still keeping the old name around for backwards compatibility.


morriscode (approved):
:shipit:

PR #563, opened 2023-06-23
Updating Rule: Adding logo_detect check to mitigate FPs

aidenmitchell:


morriscode:
No firings in 3 days, but tis been the weekend. Going to let it bake some more.

morriscode (approved):
No FP's in testing! lgtm

PR #564, opened 2023-06-23
New Rule (Attachment): ICS with embedded document

aidenmitchell:


aidenmitchell:
If this FPs, we might want to look into exploding ICS files beyond strings.

morriscode (approved):
No FP's in testing data. lgtm

PR #565, opened 2023-06-23
New Rule (Attachment): ICS file contains link to executable and suspicious language

aidenmitchell:


morriscode (approved):
No Fp's in testing. lgtm

PR #567, opened 2023-06-23
Update impersonation_paypal.yml

vector-sec:
Excluded paypalcorp.com which is used by Paypal to communicate DMARC reports


jkamdjou (approved):
thank you!

PR #569, opened 2023-06-26
New Rule: Suspicious sender from VPS

morriscode:


morriscode:
This had no hits in testing, and it doesn't retro well due to the changes made to how we parse first hop. I cleaned up the formatting a bit so it should be GTG

detection-rules/suspicious_sender_originating_from_vps.yml
```diff
@@ -0,0 +1,48 @@
+name: "Suspicious Sender originating from a VPS provider"
```
jkamdjou, line 1:
```suggestion
name: "Suspicious sender or recipient pattern originating from VPS provider"
```

detection-rules/suspicious_sender_originating_from_vps.yml
```diff
@@ -0,0 +1,48 @@
...
+  This rule detects messages with specific characteristics indicating they are sent from a Virtual Private Server (VPS) provider.
+  VPS providers offer flexible and anonymous hosting services, making them attractive to malicious actors who want to send malicious emails or engage in other illicit activities.
+  In addition, it also requires at least 1 suspicious sending techniques. 
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and (
+      any(recipients.to, .email.email == sender.email.email) or
+      sender.email.domain.root_domain in $org_domains or
```
jkamdjou, line 12:
```suggestion
```
this one feels a little risky, basically if an org has an automated system sending emails that's cloud hosted right? think we could do without it, or require more than 1 of these signals at once?


or another idea, maybe we could keep this to "commonly abused VPS providers" like OVH and what not -- though i don't know how often we see attacks from DO or Linode

morriscode: 10-4 will see if I can refactor into something more safe feeling.  I'd have expected we'd have picked up on some of that noise in test-rules but we can let it bake a while longer too.

PR #571, opened 2023-06-26
Updating Rule: scan images for PayPal impersonation as well

aidenmitchell:


morriscode (approved):
Testing results look strong.

PR #572, opened 2023-06-26
Add CI validation for signals + insights

rw-access:
In CI, we'll just turn an insight into a rule and hit the existing API. This will be enough to catch errors that we're most interested in.


morriscode (approved):
nice catch!!

jkamdjou:
@rw-access oh you clever boy

PR #573, opened 2023-06-26
New Rule: Fake fax with low reputation link

morriscode:


morriscode:
No FP's in testing. Should be safe to ship!

detection-rules/link_fake_fax_low_reputation.yml
```diff
@@ -0,0 +1,82 @@
...
+  )
+  
+  // any brand logo detected
+  and (
+      any(attachments,
+          .file_type in ('png', 'jpeg', 'jpg', 'bmp')
+          and any(ml.logo_detect(.).brands, .name is not null)
+      )
+      or any(ml.logo_detect(beta.message_screenshot()).brands,
+          .name is not null)
```
jkamdjou, line 35: do we need to exclude legit senders for the supported brands that might be impersonated? and/or want to look for high confidence? it seems a bit broad otherwise due to the tranco/mass mailer checks, but i'll defer to you on whether either of these are necessary

detection-rules/link_fake_fax_low_reputation.yml
```diff
@@ -0,0 +1,82 @@
+name: "Brand Impersonation: Fake fax with low reputation link"
+description: |
+  Detects messages with low reputation links, presence of brand logos and mentions of "fax" in subject, sender or body from a first-time sender.
```
jkamdjou, line 3:
```suggestion
  Detects the presence of known brand logos, mentions of "fax" in the subject or sender's display name, and a low reputation link from a first-time sender.
```

jkamdjou (approved):

PR #576, opened 2023-06-26
Updating rule: re-introducing subject check and adding additional NLU checks

aidenmitchell:


morriscode:
No FP's in testing

morriscode (approved):
no fp's in testing!

PR #577, opened 2023-06-27
New rule - BEC: Employee impersonation with subject manipulation

jkamdjou:


detection-rules/impersonation_employee_subject.yml
```diff
@@ -0,0 +1,37 @@
...
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+
+  // attacks have been observed where the impersonation is
+  // in the subject, and the display name contains what a
+  // subject typically would contain. this is an attempt to
+  // bypass impersonation detections
+  and subject.subject in~ $org_display_names
```
morriscode, line 13:
```suggestion
  and any($fake_org_display_names, strings.contains(subject.subject, .))
```
Do we want this instead? So instead of subject:"sam scholten" it would also match "urgent click here sam scholten"

morriscode (approved):
👍

PR #580, opened 2023-06-27
Updating rule: adding Sharepoint sender domain exemption

aidenmitchell:


jkamdjou:
good with this if you are @morriscode

morriscode (approved):
👍 This validates in testing but we may want to switch to 
`and not regex.contains(headers.message_id, "Comment(Powerpoint|Excel|Word)")`

if/when we start seeing these from sharepoint infra

PR #587, opened 2023-06-28
New Rule: Link: Multi-Factor Authentication phishing via QR code

morriscode:
@jkamdjou feel free to change the name! I wasn't sure about test rules, this uses message_screenshot and scan.qr

### DIFF
--- a/detection-rules/link_multifactor_authentication_QR_code_phishing.yml
+++ b/detection-rules/link_multifactor_authentication_QR_code_phishing.yml
@@ -1,6 +1,6 @@
 name: "Link: Multi-Factor Authentication phishing via QR code"
 description: |
-  This rule detects the verbiage related to Multi-factor authentication in the subject, body or attachments. The rule also checks for the presence of a QR code with a URL that contains the recipients email address, from an unsolicited sender.
+  This rule detects verbiage related to Multi-factor authentication in the subject, body or attachments. The rule also checks for the presence of a QR code with a URL that contains the recipients email address, from an unsolicited sender.
 references:
   
 type: "rule"
--- a/detection-rules/link_multifactor_authentication_QR_code_phishing.yml
+++ b/detection-rules/link_multifactor_authentication_QR_code_phishing.yml
@@ -28,7 +28,7 @@ source: |
   // QR code is present in attachment & recipient email address is present in the URL, a common tactic used in credential phishing attacks
   and (
       any(attachments,
-      .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif')
+          .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif')
           and any(file.explode(.), .scan.qr.data is not null
               and any(recipients.to, strings.icontains(..scan.qr.data, .email.email))
           )
--- a/detection-rules/link_multifactor_authentication_QR_code_phishing.yml
+++ b/detection-rules/link_multifactor_authentication_QR_code_phishing.yml
@@ -16,7 +16,7 @@ source: |
       regex.icontains(body.plain.raw, '(\b2fa\b|\bQ.?R\.?\s?\b|MFA|Muti[ -]?Factor Auth(entication)?)') or
       (
           any(attachments,
-              .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif')
+              .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif', 'pdf')
               and any(file.explode(.), 
                   any(.scan.strings.strings, regex.icontains(.,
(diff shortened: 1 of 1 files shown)


jkamdjou:
> @jkamdjou feel free to change the name! I wasn't sure about test rules, this uses message_screenshot and scan.qr

i think we can test - @rw-access any objections?

jkamdjou:
@morriscode there are some services that do send legit QR codes, eg im pretty sure Microsoft does, which would be unsolicited in most cases. do we need any additional signals like low rep sender or anything like that?

once we support proper URL extraction we can key off a sus link too, but sadly doesn't help us here yet

morriscode:
@jkamdjou  totally agree I wanted to let this run in test and see if we could actually use signals to try to sus out what else to build on.  But maybe a cheaper rule would be better for that?

jkamdjou:
> @jkamdjou  totally agree I wanted to let this run in test and see if we could actually use signals to try to sus out what else to build on.  But maybe a cheaper rule would be better for that?

awesome. i think unsolicited is probably enough to limit for testing, but will let @rw-access or @alexk307 confirm

detection-rules/link_multifactor_authentication_QR_code_phishing.yml
```diff
@@ -0,0 +1,62 @@
...
+      (
+          regex.icontains(subject.subject, '(\b2fa\b|\bQ.?R\.?\s?\b|MFA|Muti[ -]?Factor Auth(entication)?)')
+      ),
+      (
+          regex.icontains(body.html.display_text, '(\b2fa\b|\bQ.?R\.?\s?\b|MFA|Muti[ -]?Factor Auth(entication)?)')
+      ),
+      (
+          regex.icontains(body.plain.raw, '(\b2fa\b|\bQ.?R\.?\s?\b|MFA|Muti[ -]?Factor Auth(entication)?)')
+      ),
+      (
```
rw-access, line 22: idk if this `(` is standard but it seems pretty verbose to me. and `1 of` is technically `or` (gets turned into it anyway)

detection-rules/link_multifactor_authentication_QR_code_phishing.yml
```diff
@@ -0,0 +1,62 @@
...
+  // QR code is present in attachment & recipient email address is present in the URL, a common tactic used in credential phishing attacks
+  and any(attachments,
+      .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif')
+          and any(file.explode(.), .scan.qr.data is not null
+          and any(recipients.to, strings.icontains(..scan.qr.data, .email.email))
+      )
+  
+  // QR code is present in-line & recipient email address is present in the URL, a common tactic used in credential phishing attacks
+      or (
+          any(file.explode(beta.message_screenshot()), .scan.qr.data is not null
```
rw-access, line 42: this should not be inside `any(attachments)`

detection-rules/link_multifactor_authentication_QR_code_phishing.yml
```diff
@@ -0,0 +1,62 @@
...
+              )
+          )
+      )
+  )
+  
+  // QR code is present in attachment & recipient email address is present in the URL, a common tactic used in credential phishing attacks
+  and any(attachments,
+      .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif')
+          and any(file.explode(.), .scan.qr.data is not null
+          and any(recipients.to, strings.icontains(..scan.qr.data, .email.email))
```
rw-access, line 37: indentation is off and I think it's messing up stuff below

rw-access (commented):
I think it's filtered enough before the message screenshot, but there's one bug and some verbose formatting to tidy up

morriscode:
Thanks @rw-access will revise in the morning!

detection-rules/link_multifactor_authentication_QR_code_phishing.yml
morriscode: I've indented and unintended it, it doesn't seem to make any difference in functionality. I'll update so the ands are inline.

morriscode:
I lied tomorrow was so far away, I think it's how it should be now :)

detection-rules/link_multifactor_authentication_QR_code_phishing.yml
```diff
@@ -0,0 +1,58 @@
...
+                    '(\b2fa\b|\bQ.?R\.?\s?\b|MFA|Muti[ -]?Factor Auth(entication)?)'))
+              )
+          )
+      )
+  )
+  
+  // QR code is present in attachment & recipient email address is present in the URL, a common tactic used in credential phishing attacks
+  and (
+      any(attachments,
+      .file_type in~ ('bmp', 'png', 'jpg', 'jpeg', 'gif')
```
jkamdjou, line 31: this should be indented like you have on line 19

morriscode:
Closing this going to resubmit a new variant

PR #594, opened 2023-06-28
Attack surface reduction: EML attachment with HTML file

jkamdjou:
We discovered that the reason https://github.com/sublime-security/sublime-rules/pull/433/files was FPing was because in the following snippet:

```
and any(file.explode(.),
          .flavors.mime == "text/html"
          or .file_extension in ("htm", "html")
          or "html_file" in .flavors.yara
          or any(.scan.strings.strings, regex.contains(., "Content-Type: text/html"))
      )
```

these 2 lines are not only true for HTML attachments, but also EMLs with HTML sections:
```
          or "html_file" in .flavors.yara
          or any(.scan.strings.strings, regex.contains(., "Content-Type: text/html"))
```

this mitigates those FPs and also includes protections against bouncebacks. if we do discover FPs in testing, i've left some placeholders where we can look for additional suspicious signals

### DIFF
--- a/detection-rules/attachment_eml_with_html_attachment.yml
+++ b/detection-rules/attachment_eml_with_html_attachment.yml
@@ -1,8 +1,6 @@
 name: "Attachment: EML file with HTML attachment"
 description: |
-  Detects HTML files in EML attachments from unsolicited senders.
-
-  Reduces attack surface against HTML smuggling.
+  Detects HTML files in EML attachments to reduce HTML smuggling attack surface.
 type: "rule"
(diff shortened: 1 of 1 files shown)


morriscode (approved):
testing looks solid!

jkamdjou:
> testing looks solid!

sweet - i was considering removing Unsolicited here, but i'm a bit worried about the potential for FPs in some envs. so i'm thinking we keep this as-is, and we can add another one without Unsolicited but maybe uncommenting the additional suspicious signals

PR #601, opened 2023-07-07
New Rule: Suspicious password expiration

morriscode:
This adds coverage for a known priority FN, as well as picks up several other previous missed threats. 
No FP's in test environments, but it contains NLU so not ready for test-rules yet.


morriscode:
superseded by 1001

PR #605, opened 2023-07-11
Adding additional Amazon display name to impersonation rule

aidenmitchell:
Credit to @Im-Innocent for the email sample that flagged the need for this change.


morriscode (approved):
👍

PR #608, opened 2023-07-13
Updating Rule: Removing first password string check

aidenmitchell:
The word "password" by itself is FPing a lot, removing.


morriscode (approved):
👍

PR #610, opened 2023-07-18
Updating Rule: Switching to credphish

aidenmitchell:
The rule FPs a considerable amount when using NLU for website evaluation. Switching to credphish should fix this.


morriscode:
The primary pickups on this seem to be Proofpoint urlprotect warnings. Are you seeing FP's beyond that? I think we should try negating urldefense.com from LA.  Thoughts?
image

aidenmitchell:
agree with the negation. The FPs being addressed here were wide-ranging, didn't see any Proofpoint ones come up in my investigations.

morriscode (approved):
👍

PR #612, opened 2023-07-24
Generate static IDs for all feed rules + script

alexk307:
https://www.notion.so/sublimesecurity/PRDV-Backend-aada8800a2b4423ca94b7e5ff06be492?pvs=4

### DIFF
--- a/detection-rules/link_contains_punycode_characters.yml
+++ b/detection-rules/link_contains_punycode_characters.yml
@@ -1,4 +1,3 @@
-id: "74b3698c-d75e-52db-9596-48af93817822"
 name: "Link to a Domain with Punycode Characters"
 description: |
   The body contains a link to a domain with Punycode characters to hide the true URL destination, or contains non-printable ASCII content.
@@ -28,3 +27,4 @@ source: |
   )
 tags:
   - "Suspicious content"
+id: "74b3698c-d75e-52db-9596-48af93817822"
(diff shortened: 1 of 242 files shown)


jkamdjou:
sweet, a couple qs:

1. what do you think of adding the `id` field at the end of the file, instead of at the very top? when people open the file it's nice to see the name and description first IMO

2. has there been any discussion yet on the workflow going forward for adding `id` to new rule contributions, whether it's ones we make or PRs from the community? maybe we could create a GH action to auto-add an `id` if it doesn't exist?

alexk307:
> what do you think of adding the id field at the end of the file, instead of at the very top? when people open the file it's nice to see the name and description first IMO

@cameron-dunn-sublime had the same idea - done!

> has there been any discussion yet on the workflow going forward for adding id to new rule contributions, whether it's ones we make or PRs from the community? maybe we could create a GH action to auto-add an id if it doesn't exist?

Yes - we will need a CI check 👍  Going to add some thoughts to the README for now

detection-rules/link_contains_punycode_characters.yml
```diff
@@ -1,29 +1,30 @@
...
-        and sender.email.email not in $recipient_emails
-    )
-    or (
-        sender.email.domain.root_domain not in $free_email_providers
-        and sender.email.domain.domain not in $recipient_domains
-    )
-  )
-tags:
-  - "Suspicious content"
+name: "Link to a Domain with Punycode Characters"
```
cameron-dunn-sublime, line 1: Just calling out for future viewers these are rewritten because they have `CRLF` line endings instead of `LF`. Maybe we'll clean this up in the future w/ automation in a PR, but not a big deal for now.

cameron-dunn-sublime (approved):

cameron-dunn-sublime:
I'll add an automation to set `id`. I'll probably get to it in the next day or two.

PR #614, opened 2023-07-24
Add GitHub Action to Automatically Add & Validate Rule IDs

cameron-dunn-sublime:
This should ensure rule IDs stay in sync. I'll mark this new workflow as a required check once merged.

I may revisit the path based generation since it's not required for rules going forward, but just getting this out to fill the gap on any new rules which would otherwise be missing `id`.


hugh-sublime (approved):

scripts/generate-rule-ids/main.py
```diff
@@ -18,35 +18,43 @@ def generate_rule_ids():
...
                     with open(full_file_path, 'r') as f:
                         contents = f.read()
 
-                    try:
                         # load the YAML and check if it already has an ID
                         parsed = yaml.safe_load(contents)
                         feed_rule_id = parsed.get('id')
 
                         if feed_rule_id:
+                            # It's expected for this ID not to always match the path generated ID. The purpose of the
```
alexk307, line 41: It'd be sweet if we could check if the PR itself was trying to edit the ID too

cameron-dunn-sublime: I might try and add something like that later, but it's a bit more complex and might have some edge cases with how Git recognizes changes.

detection-rules/new_rule.yml
```diff
@@ -0,0 +1,29 @@
...
+              and sender.email.email not in $recipient_emails
+          )
+          or (
+              sender.email.domain.root_domain not in $free_email_providers
+              and sender.email.domain.domain not in $recipient_domains
+          )
+  )
+tags:
+  - "Suspicious attachment"
+  - "Attack surface reduction"
```
cameron-dunn-sublime, line 29:
```suggestion
  - "Attack surface reduction"
id: "8a84a08c-5576-5ecd-915f-2acdda3e3b28"
```

hugh-sublime (approved):
Tandem review!

PR #624, opened 2023-07-26
Adjust github.ref to head_ref

cameron-dunn-sublime:
Screenshot 2023-07-26 at 4 49 22 PM

In `pull_request_target` workflows the `github.ref` is the base branch, [see here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target).

The effects of it can be seen here:
* https://github.com/sublime-security/sublime-rules/pull/622
   * Rule validation should have failed
* https://github.com/sublime-security/sublime-rules/pull/621
   * https://github.com/sublime-security/sublime-rules/actions/runs/5674758230
   * CI cancelled because both competed for `main`

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -8,8 +8,8 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
 
 jobs:
   tests:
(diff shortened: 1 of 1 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -8,8 +8,8 @@ on:
   workflow_dispatch: {}
 
 concurrency:
-  group: ${{ github.ref }}
-  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
```
cameron-dunn-sublime, line 12: `github.ref` would be empty for on.push.branches events, so this would have already always been `true`. I'd also be fine having this always be `false`. I think it's not very relevant for rules where CI is fast and the rate of merges is low (and now we require CI to run on up to date branches anyways 🤷 )

hugh-sublime (approved):

PR #628, opened 2023-07-27
Change Job Ordering/Make Single Job/Prevent X on main

cameron-dunn-sublime:
Minor adjustments, happy with how everything is working.

https://github.com/sublime-security/sublime-rules/pull/627 if you look at this PR the Checks at the bottom don't show the failure. That's because one job failed, and then the other job pushed a commit so the HEAD commit doesn't have any statuses reported. We can change our ordering to only push after running the tests to fix this. And if we're going to run in that order, we might as well just use a single job and add IDs before running tests.

Also `main` shows failures because the step to set a status only works if the workflow is triggered for a PR, so we'll skip that step if it's not a PR workflow.

https://github.com/sublime-security/sublime-rules/pull/629 is a PR which consumes these changes.

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -91,6 +91,7 @@ jobs:
       # the latest commit, and we can depend on that as a required check.
       - name: "Create a check run"
         uses: actions/github-script@v6
+        if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


hugh-sublime (approved):
🚢

PR #630, opened 2023-07-27
New rule - Spam: URL shortener with short body content and emojis

jkamdjou:


morriscode:
Test rules still isn't consuming rules, this one included. 
I ran the rule in multiple environments and had no FP's or TP's. 
Consider this approved, but wanted to leave open incase we need a rule to test why test-rules is not consuming.

jkamdjou:
> Test rules still isn't consuming rules, this one included. I ran the rule in multiple environments and had no FP's or TP's. Consider this approved, but wanted to leave open incase we need a rule to test why test-rules is not consuming.

@morriscode can i get that sweet ✅

morriscode (approved):

PR #632, opened 2023-07-27
Updating Rule: Fixing FP

aidenmitchell:


response.txt
```diff
@@ -0,0 +1 @@
+{"functions":["length"],"list":["recipient_domains"]}
```
morriscode, line 1: @aidenmitchell is this something tines is adding to your PR's? I've seen this in a couple so far this morning

aidenmitchell: No, this gets added by the ID generator. No idea what’s going on.
cc @rw-access who asked about this as well

morriscode (approved):
🐎

PR #633, opened 2023-07-27
Update attachment_vba_macro_auto_exec_unsolicited.yml

morriscode:
Adding reply negation. This solves all known FP's.

### DIFF
--- a/response.txt
+++ b/response.txt
@@ -1,0 +1,1 @@
+{"functions":["length"],"list":["recipient_domains"]}
--- a/response.txt
+++ b/response.txt
@@ -1,1 +1,0 @@
-{"functions":["length"],"list":["recipient_domains"]}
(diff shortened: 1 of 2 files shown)


response.txt
```diff
@@ -0,0 +1 @@
+{"functions":["length"],"list":["recipient_domains"]}
```
jkamdjou, line 1: accidental commit?

morriscode: Yeah it looks like something is popping that artifact up there.

morriscode: well the file's been deleted but github says it hasn't....

morriscode:
@jkamdjou this should be gtg now.

cameron-dunn-sublime:
@morriscode should be good now. `response.txt` is still shown here but when I checkout the branch it's not there. If it merges to main and includes the file I'll clean it up, but it won't cause problems regardless.

jkamdjou (approved):

PR #615, opened 2023-07-25
Updating NLU rules to use current thread vs coalesce

morriscode:

PR #634, opened 2023-07-27
Updating Rule: Reformatting & changing to inner_text

aidenmitchell:


aidenmitchell:
Semi-conflicts with #615, but I'd like to get this change out first.

response.txt
```diff
@@ -0,0 +1 @@
+{"functions":["length"],"list":["recipient_domains"]}
```
rw-access, line 1: what's this? Seems like it was accidentally committed?

morriscode (approved):
📞

PR #638, opened 2023-08-01
Fix check name, and make title/etc consistent

cameron-dunn-sublime:


madirey (approved):

.github/workflows/update-test-rules.yml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: lewagon/wait-on-check-action@v1.3.1
         with:
           ref: ${{ steps.comment-branch.outputs.head_sha }}
-          check-name: 'Rule Tests and ID Generation'
+          check-name: 'Rule Tests and ID Updated'
```
cameron-dunn-sublime, line 54: This is the correct name (line 110 of `.github/workflows/rule-validate.yml`). I missed that line and thought Title = Name.

I made the title the same now just to avoid future confusion.

PR #642, opened 2023-08-02
Rule Tune: Negating paypal senders with dmarc pass due to FP reports

morriscode:


morriscode:
This exclusion works, rule is good to go.

jkamdjou (approved):

PR #655, opened 2023-08-10
New Rule: BEC with out of band pivot

morriscode:


detection-rules/body_business_email_compromise_out_of_band_pivot.yml
```diff
@@ -0,0 +1,36 @@
...
+source: |
+  type.inbound
+  and length(body.current_thread.text) < 500
+  and regex.icontains(body.current_thread.text,
+                      "personal email|(whatsapp|cell(phone)?).{0,8}(#|number)"
+  )
+  and (
+    any(ml.nlu_classifier(body.current_thread.text).entities,
+        .name == "urgency" or .name == "request"
+    )
```
jkamdjou, line 16: do we need to exclude known senders for this bit of logic? this will basically flag on messages with "cell" and a request in it right?

jkamdjou (commented):
do you wanna kick off rule testing?

detection-rules/body_business_email_compromise_out_of_band_pivot.yml
morriscode: We were probably typing at the same time. Test rules were just kicked off. Right before you commented. It has to contain cell #|number and be short, testing looked good I'm curious how prone it will be without gating.

jkamdjou:
how's this lookin' in testing?

PR #657, opened 2023-08-11
New Rule: Office file loading external template

morriscode:
This is very similar to "attachment_office_remote_doc_template.yml", this rule however leverages  oletools which appears to be more accurate. The previous rule was missing office documents exploiting follina  (CVE-2022-30190)

oleobj considers any of the following .names to be of "high" risk however this was a bit too loose and fires on benign docs. The samples i've seen all exploit `attachedTemplate`, so that is what we've scoped to here. 

-          attachedTemplate
-          externalLink
-          externalLinkPath
-          externalReference
-          frame
-          hyperlink
-          officeDocument
-          oleObject
-          package
-          slideUpdateUrl
-          slideMaster
-          slide
-          slideUpdateInfo
-          subDocument
-          worksheet


detection-rules/attachment_office_external_template.yml
```diff
@@ -0,0 +1,29 @@
...
+  type.inbound
+  and any(attachments,
+          (
+            // office files
+            .file_extension in~ $file_extensions_macros
+            or .file_extension in~ $file_extensions_common_archives
+          )
+          and file.oletools(.).indicators.external_relationships.risk == "high"
+          and any(file.oletools(.).relationships,
+                  .name == "attachedTemplate"
```
aidenmitchell, line 16:
```suggestion
                  .name in ("attachedTemplate", "oleObject")
```
We're currently not firing on [this sample](https://app.docguard.io/b5aafc10c38efe0a91c7ceae4b911810c4a148bd55fae301f5283fc1a2779642/c9b82d1a-790f-4cf5-b719-b4f8201d1881/0/results/dashboard), but we would with this change ([analyzer link](https://analyzer.sublime.security?id=ca290aae-5b2c-445a-8b32-5df825bf7a8e)). Thoughts?

aidenmitchell: It _might_ be too loose... `oleObject` is quite broad. We can run some hunts and see what happens?

aidenmitchell (changes requested):

PR #658, opened 2023-08-11
Updating File_extension to File_type due to misses on files with non standard extensions

morriscode:
Adding an or condition for rar file type check.


detection-rules/attachment_archive_with_exe.yml
```diff
@@ -14,7 +14,7 @@ type: "rule"
 severity: "high"
 source: |
   type.inbound
-  and any(attachments, .file_extension in~ $file_extensions_common_archives
```
rw-access, line 17: porque no los dos? also remember `file_type` values might not intersect with that list how you think, so this might introduce some FNs

rw-access, line 17: File types and file extensions aren't 1:1, so you'll want to manually compare all the known `.file_type` values to `$file_extensions_common_archives` to validate overlap. Probably want an `or` here to be safe too

aidenmitchell (approved):

PR #661, opened 2023-08-14
New Rule: Brand impersonation: Microsoft fake sign-in alert

morriscode:


detection-rules/impersonation_microsoft_fake_sign_in_alert.yml
```diff
@@ -0,0 +1,96 @@
...
+  Detects messages impersonating Microsoft that mimic sign-in security alerts and attempt to solicit a response.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  // Microsoft strings
+  and (
+    strings.contains(subject.subject, "Microsoft")
+    or strings.contains(sender.display_name, "Microsoft")
+    or strings.contains(body.current_thread.text, "Microsoft")
```
jkamdjou, line 12: this one feels a little bit risky. is it necessary for the attacks you've seen?

detection-rules/impersonation_microsoft_fake_sign_in_alert.yml
```diff
@@ -0,0 +1,96 @@
...
+    
+      // sender is not freemail, but the return path email or reply to email is  
+      sender.email.domain.domain not in $free_email_providers
+      and (
+        headers.return_path.domain.root_domain in $free_email_providers
+        or (
+          length(headers.reply_to) > 0
+          and (
+            all(headers.reply_to, .email.domain.root_domain in $free_email_providers)
+            and sender.email.domain.domain not in $free_email_providers
```
jkamdjou, line 46: not necessary, you've got it above

detection-rules/impersonation_microsoft_fake_sign_in_alert.yml
morriscode: We could probably cut body text, the known examples have "Microsoft" in all 3. But I didn't want to make it so strict that it couldn't accommodate permutations. I figured this was safe with the required body text matches + freemail, or non freemail and all the other requirements. Definitely willing to consider cutting if you think we should.

morriscode: Also we could let this bake in test-rules and spin up a slightly more specific approach, I just worry about it's longevity when it's tighter.

jkamdjou: we've got an "or" below this for the brand logo so i'd feel ok cutting body text and merging this in the AM if you are

detection-rules/impersonation_microsoft_fake_sign_in_alert.yml
```diff
@@ -0,0 +1,95 @@
...
+  - "Impersonation: Brand"
+  - "Social engineering"
+detection_methods:
+  - "Computer Vision"
+  - "Content analysis"
+  - "File analysis"
+  - "Header analysis"
+  - "Sender analysis"
+  - "Whois"
+
```
jkamdjou, line 94:
```suggestion
```

detection-rules/impersonation_microsoft_fake_sign_in_alert.yml
morriscode: I tossed the logo stuff in just in case they start using it, right now I haven't observed it, that's why I wanted to check all 3 spots.

PR #669, opened 2023-08-17
New Rule: Headers: Recipient SLD match X-mailer

morriscode:
Would like to run this in test-rules for a while


morriscode:
this hasn't yielded anything significant in testing

PR #671, opened 2023-08-18
Improve usage of .scan.strings.strings to use OCR or `of`

rw-access:
A few things going on in this PR
- Manually updated rules to use `.scan.ocr.raw` instead of `.scan.strings.strings` specifically for images + PDFs when it's clear that's what we want
- Fixed logic in a few rules
- Changed usage of `length(filter(.scan.strings.strings, strings.ilike(...)) >= N` to `N of ( ... )` because that appears to reflect the intention
- Autoformatted a bunch of rules as I made changes.

I recommend hiding whitespace in the diff when reviewing changes!


detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -4,14 +4,13 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  and length(filter(attachments, .file_type not in~ ('png', 'jpeg', 'jpg', 'bmp'))) == 0
-  and length(body.links) >0
+  and all(attachments, .file_type in~ ('png', 'jpeg', 'jpg', 'bmp'))
```
rw-access, line 7: These are logically identical but the other one seemed more confusing for me to understand

detection-rules/attachment_callback_phish_with_img.yml
```diff
@@ -10,48 +10,47 @@ severity: "high"
...
-                "*norton*",
-                "*ebay*",
-                "*paypal*",
-                "*secure anywhere*"
-            ))) >=1)
-            )
+  and any(attachments,
+          .file_type in~ ('png', 'jpg')
+          and any(file.explode(.),
+                  4 of (
```
rw-access, line 26: We chatted about this one in Slack. The main distinction is that we want to know that at least 4 of our substrings were found. The previous approach checks for how many strings return from strelka match at least one.

If `.scan.strings.strings` was `["to purchase an antivirus subscription now, call or cancel now"]`, then we wouldn't match before because everything was in the same string. Alternatively, if we you had `["+1", "+1", "+1", "+1"]`, that used to match before because it didn't check if the same thing was matched.

Ultimately it just means we were counting the wrong thing and it was inverted, so this is one way to flip it.

detection-rules/attachment_docusign_image_suspicious_links.yml
```diff
@@ -4,30 +4,37 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  and length(filter(attachments, .file_type not in~ ('png', 'jpeg', 'jpg', 'bmp', 'gif'))) == 0 
+  and length(attachments) > 0
```
rw-access, line 7: Not necessary, but I figured it improved readability.

detection-rules/attachment_macro_dll_loader.yml
```diff
@@ -6,33 +6,35 @@ references:
...
-              .scan.zip.encrypted == false
-              // zip contains a dll file
-              and any(.scan.zip.all_paths, strings.icontains(., "dll")))
-          and any(file.explode(.),
-              // macro references a dll file
-              any(.flavors.yara, strings.like(., "vb_file"))
-              and any(.scan.strings.strings, strings.icontains(., "dll"))
-             )
-         )
-         or any(file.explode(.), // fallback for encrypted zips
```
rw-access, line 21: This `or` was misplaced, which was revealed by the editor

detection-rules/attachment_macro_dll_loader.yml
```diff
@@ -6,33 +6,35 @@ references:
...
-             .scan.zip.encrypted == true
-             and any(.scan.zip.all_paths, any($file_extensions_macros, strings.icontains(.., .)))
-             // zip contains a dll file
-             and any(.scan.zip.all_paths, strings.icontains(., "dll"))
-         )
-  or any(attachments, .file_extension in~ $file_extensions_common_archives
```
rw-access, line 27: This `any(attachments, ...)` was actually nested inside another `any(attachments, ...)`!

detection-rules/attachment_malicious_onenote_commands.yml
```diff
@@ -12,30 +12,25 @@ severity: "high"
...
   and any(attachments,
-      (
-          .file_extension in~ ("one") or
-          .file_extension in~ $file_extensions_common_archives
-      )
-      and any(file.explode(.),
-      any(
-          .flavors.yara, . == "onenote_file"
-           and
-            any(..scan.strings.strings,
```
rw-access, line 23: This was `..scan.strings.strings` because it was actually nested _inside_ the `any(.flavors.yara, ...)` loop!

detection-rules/body_microsoft_logo_bing_redirect.yml
```diff
@@ -8,57 +8,51 @@ source: |
 
   // Microsoft logo
   and (
-      any(attachments,
-          .file_type in ('png', 'jpeg', 'jpg', 'bmp')
-          and any(ml.logo_detect(.).brands, strings.starts_with(.name, "Microsoft"))
-      )
-      or 
-      any(attachments,
```
rw-access, line 16: Consolidated

detection-rules/impersonation_microsoft_quarantine.yml
```diff
@@ -5,32 +5,34 @@ type: "rule"
...
-          any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft" and .confidence != "low")
-          and any(ml.nlu_classifier(.scan.ocr.raw).entities, .name == "urgency")
-      )
-       and any(file.explode(.),
-          3 of (
-              any(.scan.strings.strings, strings.icontains(., "review")),
-              any(.scan.strings.strings, strings.icontains(., "release")),
-              any(.scan.strings.strings, strings.icontains(., "quarantine")),
-              any(.scan.strings.strings, strings.icontains(., "messages")),
-              any(ml.logo_detect(..).brands, strings.starts_with(.name, "Microsoft"))
```
rw-access, line 20: Nesting logo_detect inside `file.explode` was a little odd, so I refactored this so they are all at the same place. This still keeps the same meaning though! But I can back it out if it's confusing

morriscode:
Do we want to change these to "$file_types_images". Also I'd like to run regressions on this but I'm not sure where that is at or how to invoke it?

rw-access:
Yeah, regression testing is definitely good here. I can update to `$file_types_images` as well. Some of the rules are only looking at 2-3 extensions though, but we're okay opening them up more to others, right?

morriscode:
> Yeah, regression testing is definitely good here. I can update to `$file_types_images` as well. Some of the rules are only looking at 2-3 extensions though, but we're okay opening them up more to others, right?

Yeah that should be fine. I think early on we assumed gif wouldn't be as prevalent as it is.

rw-access:
Dropping this PR and splitting up the remaining work to a few others

PR #672, opened 2023-08-18
Run autoformatting on all rule source

rw-access:
Before I make a few other PRs like #671 where I'll use autoformatting, I figure we can run it first on all rules. If we find anything we don't like, we can manually fix/revert it or add more conditions to the MQL autoformat command.


detection-rules/attachment_dropbox_image_suspicious_links.yml
morriscode, line 1: Do we want to pull JPEG here and replace it with gif? Or should all these be converted to the $file_types_images?

detection-rules/attachment_any_html_new_sender.yml
morriscode, line 1: This is probably personal preference and readability, but the single line or condition might be lost or at the very least isn't as apparent as putting it on a second line.

discovery-rules/attachment_suspicious_macro.yml
morriscode, line 1: Changing the comma delimited lists to from a /s to a \r\n makes the rule bigger but imo easier to read.

morriscode (approved):
This looks great, I left a few comments that are probably just personal preferences or things i'm accustom to, but in review I'm really excited to have a level of consistent formatting throughout the feed rules.

discovery-rules/attachment_suspicious_macro.yml
rw-access: For sure. I've also thought about having them lay out in a grid. Like for example, you could do 4 columns per line, all lined up vertically. It's a little trickier but could be neat.

And after a certain size, it makes sense to put them in a list.

PR #674, opened 2023-08-18
Updating Rule: re-naming bit.ly rule + adding flagged detection

aidenmitchell:
Turns out bit.ly has 2 kinds of gating, complete block (which is what we were detecting before), and a warning page that lets you continue. Adding detection for the latter.

We could also make a new 2nd rule if we want to differentiate in the UI, open to that.

Block page:
a9cc91a6-2041-4742-ba20-3677e7a8c010

Warning page:
52774d5b-5bb6-48f8-b117-0a4104842f95


morriscode (approved):

morriscode:
cc @aidenmitchell conflicts ahoy!

PR #677, opened 2023-08-21
Autoformat a few other rules that were missed

rw-access:
Looks like these ones were autoformatted with an old version. Fixed


aidenmitchell (approved):

morriscode (approved):
💯

PR #678, opened 2023-08-21
Switch rules to use $file_types_images

rw-access:
There are a few cases where we want image OR `pdf`. Does it make sense to add PDF to that list? I think a case could be made either way, I don't have strong opinions.

I also replaced `in~` with `in` because these will always match the case exactly anyway (unlike a file extension)

### DIFF
--- a/detection-rules/inline_image_as_message.yml
+++ b/detection-rules/inline_image_as_message.yml
@@ -17,7 +17,7 @@ source: |
       and any(body.links, not strings.ilike(.href_url.url, "*.png"))
     )
     // cid images are treated as attachments, so we're looking for more than 1
-    or (length(attachments) > 1 and any(attachments, .file_type not in file_types_images))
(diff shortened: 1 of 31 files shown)


detection-rules/inline_image_as_message.yml
```diff
@@ -17,7 +17,7 @@ source: |
       and any(body.links, not strings.ilike(.href_url.url, "*.png"))
     )
     // cid images are treated as attachments, so we're looking for more than 1
-    or (length(attachments) > 1 and any(attachments, .file_type not in ("jpg", "png", "gif")))
+    or (length(attachments) > 1 and any(attachments, .file_type not in file_types_images))
```
morriscode, line 20:
```suggestion
    or (length(attachments) > 1 and any(attachments, .file_type not in $file_types_images))
```

morriscode (commented):
LGTM one suggestion for you @rw-access

morriscode (approved):
👀 Let's go!! Thanks @rw-access

PR #679, opened 2023-08-21
Adding extension-less Office file snippet

aidenmitchell:
Office files without an extension appear as
```
.file_extension is null
and .file_type == "unknown"
and .content_type == "application/octet-stream"
```
in MQL, so adding this snippet to all of our Office file rules to ensure coverage. Size check is a guard rail.


detection-rules/attachment_cve_2023_21716_rtf_fonts.yml
```diff
@@ -15,6 +15,12 @@ source: |
             .file_extension in~ ("rtf", "doc", "docx")
             or .file_extension in~ $file_extensions_common_archives
             or .file_extension in~ $file_extensions_macros
+            or (
+                .file_extension is null
+                and .file_type == "unknown"
```
rw-access, line 20: Not blocking you, but it sounds like we should add more file type matchers based on the contents? That way we can detect pdf(x), ppt(x), doc(x), etc

rw-access: totally, nothing you need to change here. Just making an observation that it sounds like we're short on some file type matchers in the backend

morriscode:
100000 is only .1mb. might want to bump it?

morriscode (approved):

PR #684, opened 2023-08-21
Improve usage of OCR in rules and switch more to `of`

rw-access:
This is basically #671 but after the other refactors to keep it from being as overwhelming.
I tried to keep concerns here focused to one of these two things:

- Manually updated rules to use `.scan.ocr.raw` when it makes sense
- Switch from `length(filter.scan.strings.strings, ...)` to `of` when it makes more sense
- Fixed or simplified logic in a few places in these files when I encountered them


morriscode:
hero! Just tested against a FP i was going to tune and this remediated it>

morriscode (approved):
👍

PR #687, opened 2023-08-22
Update open_redirect_slack.yml

morriscode:
Negating known FP's

### DIFF
--- a/detection-rules/open_redirect_slack.yml
+++ b/detection-rules/open_redirect_slack.yml
@@ -5,6 +5,7 @@ type: "rule"
 severity: "low"
 source: |
   type.inbound
+  and 0 < length(body.links) < 10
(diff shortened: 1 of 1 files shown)


detection-rules/open_redirect_slack.yml
```diff
@@ -5,6 +5,7 @@ type: "rule"
 severity: "low"
 source: |
   type.inbound
+  and 0 < length(body.links) < 10
```
rw-access, line 8: cool to see this usage! (cc @hugh-sublime)

(technically, the `any(body.links, ...)` check will automatically add a `length(body.links) > 0`, so you don't have to think about it. but either way this doesn't hurt anything if it improves your readability.

rw-access (approved):

PR #692, opened 2023-08-22
Regression fixes

morriscode:
Fixing Regression failures.


morriscode:
Fairly minor changes here, dropping a word from one, and adding a "+1" to strings check in the other.

jkamdjou (approved):

rw-access (approved):
thanks for catching these @morriscode!

PR #701, opened 2023-08-24
Update Callback Phishing body - Modifying to use current_thread deprecate HTML version

morriscode:
Previous versions matched 75 messages.
New version matches 98. 

The switch to body.current_thread negates the use for two distinct versions of this rule. Also rule has been updated to carry the same strings as the attachment_pdf counterpart. This is a marketable improvement.


jkamdjou (approved):

morriscode:
> 

haha yes attacks sorry!

PR #711, opened 2023-08-24
Updating Rule: Negate legit senders

aidenmitchell:
Including link check to prevent circumvention by sender spoofing.


morriscode (approved):
👍

PR #712, opened 2023-08-24
Updating Rule: updating regex to avoid FPs

aidenmitchell:


jkamdjou (commented):
lgtm, does it pass regression tests? i can't tell on mobile

aidenmitchell:
Not sure, regression testing does not work on forks yet... we'll do some Shared Samples hunts to be safe.

aidenmitchell:
14 day hunt in Shared Samples shows no regression, only email dropped is the FP.

morriscode (approved):
➕

PR #715, opened 2023-08-25
Tune attached EML with credential phishing

jkamdjou:
Resolves an FN and mitigates some FPs. Further FP mitigation coming soon on the LA/Hydra side by [not analyzing links to images](https://www.notion.so/sublimesecurity/Plumb-ContentType-from-credphish-api-to-MQL-linkanalysis-f9272cc985914642baf0a3d736f7030a?pvs=4). We could remove the Unsolicited snippet, but we should probably wait for the latter to hit first.


morriscode (approved):
🏁

PR #723, opened 2023-08-25
New Rule (Attachment): Fake attachment image lure

aidenmitchell:
Uses `beta.parse_eml`, so can't merge until that's live.


aidenmitchell:
Also added "EML analysis" as a detection method (because of `parse_eml`.)
Thoughts?

morriscode:
@aidenmitchell  blast from the past. 

We could add an additional or `    any(attachments,
        .file_type in $file_types_images
        and any(ml.logo_detect(.).brands, .name =="FakeAttachment"))
    )`

morriscode (approved):

PR #727, opened 2023-08-26
New Rule: Attachment: HTML smuggling with base64 encoded JavaScript function

morriscode:


jkamdjou (approved):
nice, seems pretty safe to me

PR #731, opened 2023-08-27
Tune brand impersonations for Vanta and Barracuda

jkamdjou:


detection-rules/impersonation_vanta.yml
```diff
@@ -8,8 +8,8 @@ authors:
 source: |
   type.inbound
   and (
-    strings.ilike(sender.display_name, '*vanta*')
-    or strings.ilike(sender.email.email, '*vanta*')
+    regex.imatch(sender.display_name, '\bvanta\b')
+    or strings.ilike(sender.email.domain.local_part, '*vanta*')
```
morriscode, line 12:
```suggestion
    or strings.ilike(sender.email.local_part, '*vanta*')
```

jkamdjou: ty

morriscode:
I really like the idea of bolstering these with whois too.  This lets us look for string exact matches in the SLD or root domain, and (most likely) not fire on legitimate domains.  Or loosen levenshtein beyond 1 permutation and gate with age. 

I need to play around with it some more to figure out what the right combination is but i think there's something there. 

eg.
```
strings.ilike(sender.email.domain.root_domain, "*vanta*")
      and beta.whois(sender.email.domain).days_old <= 30
```

morriscode (approved):
👍

PR #733, opened 2023-08-28
Negating FP: impersonation_barracuda.yml

morriscode:
Communications regarding  San Jose Barracudas


morriscode:
Looks like josh did this already

PR #734, opened 2023-08-28
New Rule: HTML file containing the recipients email in JavaScript strings

morriscode:


morriscode:
Testing shows overlap with known malicious HTM(L) files that are only firing ASR rules.

morriscode:
queuing in test rules, but I'd imagine this is only used nefariously... Could pair it with a 'decrypt' string or a `any(.scan.javascript.identifiers, strings.contains(., "CryptoJS"` if we have concerns.

jkamdjou:
> queuing in test rules, but I'd imagine this is only used nefariously... Could pair it with a 'decrypt' string or a `any(.scan.javascript.identifiers, strings.contains(., "CryptoJS"` if we have concerns.

i'd be in favor of adding another signal like this, and/or adding some sus sender/auth signals

morriscode:
Test results already look golden! This is gtg @jkamdjou

detection-rules/attachment_html_recipient_in_javascript_identifiers.yml
```diff
@@ -0,0 +1,39 @@
+name: "Attachment: HTML file containing the recipients email in JavaScript strings"
```
jkamdjou, line 1:
```suggestion
name: "Attachment: HTML with obfuscation and recipient's email in JavaScript strings"
```

PR #736, opened 2023-08-29
Updating Rule: Adding reply/forward negation

aidenmitchell:


morriscode:
Gonna do this another way to maintain some similarities to the static rule

PR #739, opened 2023-08-29
Updating Rule: Adding "shtml" negation

aidenmitchell:
There's an "sh" in "shtml"... fixing.

### DIFF
--- a/detection-rules/attachment_pdf_with_low_reputation_link_to_suspicious_files.yml
+++ b/detection-rules/attachment_pdf_with_low_reputation_link_to_suspicious_files.yml
@@ -13,6 +13,7 @@ source: |
           and any(file.explode(.),
                   any(.scan.pdf.urls,
                       regex.contains(.path, '\.(?:exe|cab|vbs|ps1|rar|iso|dll|one|lnk|sh)')
+                      and not strings.icontains(.path, "shtml")
                       and .domain.root_domain not in $tranco_1m
                   )
           )


detection-rules/attachment_pdf_with_low_reputation_link_to_suspicious_files.yml
```diff
@@ -13,6 +13,7 @@ source: |
           and any(file.explode(.),
                   any(.scan.pdf.urls,
                       regex.contains(.path, '\.(?:exe|cab|vbs|ps1|rar|iso|dll|one|lnk|sh)')
+                      and not strings.icontains(.path, "shtml")
```
rw-access, line 16:
```suggestion
                      and not strings.icontains(.path, ".shtml")
```

detection-rules/attachment_pdf_with_low_reputation_link_to_suspicious_files.yml
```diff
@@ -13,6 +13,7 @@ source: |
           and any(file.explode(.),
                   any(.scan.pdf.urls,
                       regex.contains(.path, '\.(?:exe|cab|vbs|ps1|rar|iso|dll|one|lnk|sh)')
```
rw-access, line 15: maybe?
```suggestion
                      regex.contains(.path, '\.(?:exe|cab|vbs|ps1|rar|iso|dll|one|lnk|sh)\b')
```

morriscode (approved):
\A+ on the \b

PR #740, opened 2023-08-29
Updating Rule: Adding bigcommerce negation

aidenmitchell:


morriscode (approved):
We'll probably end up adding some additional gating to this rule today, or tomorrow too!

PR #746, opened 2023-08-30
FP/FN negations: Update impersonation_netflix.yml

morriscode:
Adding age check or links to a non "com", "org", "net", "ms" domain

### DIFF
--- a/detection-rules/impersonation_netflix.yml
+++ b/detection-rules/impersonation_netflix.yml
@@ -21,10 +21,20 @@ source: |
           .href_url.domain.tld not in ("com", "org", "net", "ms") and .href_url.domain.valid == true
       )
     )
+    or 2 of (
+      strings.icontains(body.current_thread.text, "restart"),
+      strings.icontains(body.current_thread.text, "cancelled"),
+      strings.icontains(body.current_thread.text, "membership"),
+      strings.icontains(body.current_thread.text, "expir"),
+      strings.icontains(body.current_thread.text, "payment")
+      and length(body.links) > 0
+      and not any(body.links,
+                  .href_url.domain.root_domain in ("netflix.com", "dvd.com", "netflixfamily.com")
+      )
+    )
   )
   and sender.email.domain.root_domain not in ('netflix.com', 'dvd.com', 'netflixfamily.com')
   and sender.email.email not in $recipient_emails
-
 attack_types:
   - "Credential Phishing"
 tactics_and_techniques:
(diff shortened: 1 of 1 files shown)


morriscode:
Gonna let regression run on this, but hunts look amazing.

morriscode:
re-queueing mimic

morriscode:
The two regressions on this 

- EML-17381
- EML-993

These regressed because the emails are old and the mdm has no body.current_threat.text. These emails both fire the rule in this PR in manual runs. 

cc @cameron-dunn-sublime  Not sure what the best way to proceed here is, penny for your thoughts!

cameron-dunn-sublime:
@morriscode I see `body.current_thread.text` for both EMLs.

[`17381`](https://github.com/sublime-security/shared-rule-tests/blob/main/test_data/17381/17381_mdm.json#L323)
[`993`](https://github.com/sublime-security/shared-rule-tests/blob/main/test_data/993/993_mdm.json#L61)

MDMs are re-parsed by the test suite anytime a new version of mantis is released, so we'll always have the latest features and changes for parsing.

All that said, I'm not sure why they don't flag here. I'm looking into it!

detection-rules/impersonation_netflix.yml
```diff
@@ -14,6 +14,25 @@ source: |
...
+    )
+    or 2 of (
+      strings.icontains(body.current_thread.text, "restart"),
+      strings.icontains(body.current_thread.text, "cancelled"),
+      strings.icontains(body.current_thread.text, "membership"),
+      strings.icontains(body.current_thread.text, "expir"),
+      strings.icontains(body.current_thread.text, "payment")
+      and length(body.links) > 0
+      and not any(body.links,
+                  .href_url.domain.root_domain in ("netflix.com", "dvd.com", "netflixfamily.com")
```
cameron-dunn-sublime, line 32: Do these checks belong in here (`2 of (...`?

(I don't think this is related the mimic tests failing, just something I noticed)

morriscode: Nice!! Thank you, yeah i think we're just missing a set of parens will update.

cameron-dunn-sublime:
@morriscode tests are passing now. The results didn't show because the tests took longer to run due to `beta.whois` calls. This will resolve itself on the next auto update of the test suite, but I also am bumping the timeout [here](https://github.com/sublime-security/sublime-rules/pull/784).

PR #753, opened 2023-08-31
Update DLL macro attachment rule logic

rw-access:
I forgot to do this in a previous PR. Looks like we had a nested attachments in attachments check. That's fixed now


morriscode (approved):
bless

PR #760, opened 2023-08-31
New Rule: Link: Suspicious display name with image attachment and short or no body text

morriscode:


morriscode:
I think I got a more fun version than this.

PR #761, opened 2023-09-01
Updating Rule: attachment_docusign_image_suspicious_links.yml

aidenmitchell:
FTS fails when internal senders are spoofed, this should fix that.


morriscode (approved):
gorgeous

morriscode (approved):

PR #765, opened 2023-09-01
Updating Rule: headers_voicemail_sendgrid.yml

aidenmitchell:
Adding NLU requirement.


morriscode:
if .intents is 0 it seems to not fire also, not sure if that's intentional or a happy accident.

aidenmitchell:
That's intentional, thanks for checking! Have def done that unintentionally.

morriscode (approved):
🧠

PR #773, opened 2023-09-05
Regression fix: link_credential_phishing_secure_message.yml

aidenmitchell:


morriscode (approved):
I think the length check was initially based on an assumption that these urls would have a minimum length due to hashing/encryption. I couldn't find any concrete evidence this was included for FP negations initially. I ran some hunts just to ensure we wouldn't see unwanted coverage and believe this should be safe.

PR #774, opened 2023-09-05
Moving NLU logic for Extortion rule to current_thread

bfilar:


morriscode (approved):
Good call! I don't know how this one got missed, my bad!

PR #776, opened 2023-09-06
Regression Tuning:  link_fake_thread_nlu_financial_request.yml

morriscode:
This rule is currently in discovery. We have several regressions that it was solving for, but many more FP's. Added new logic and increased the required matches.


morriscode:
Letting mimic run to explore regression.

morriscode:
Tracking mimic new flags [here](https://www.notion.so/sublimesecurity/Tune-and-Re-enable-Fake-message-thread-with-a-suspicious-link-and-engaging-language-from-an-unknown-ac74932d65634ad9b4b3cacb4e1f172c?pvs=4)

23 New FN/TP pickups with the revisions. 
3 messages Mimic says are flagging are FP's. However they do not flag when I run the rule against them directly.

morriscode:
Still seeing a decent amount of marketing, adding some additional logic, and combining request and financial together.

morriscode:
New revisions have quashed marketing noise. Rule efficacy looks strong in hunts and matches where it needs to in regression testing.

jkamdjou (approved):

PR #778, opened 2023-09-07
Updating rule: link_microsoft_low_reputation.yml

aidenmitchell:
Removing link check and editing NLU logic. Holds up well in hunts.

### DIFF
--- a/detection-rules/link_microsoft_low_reputation.yml
+++ b/detection-rules/link_microsoft_low_reputation.yml
@@ -4,34 +4,12 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  // suspicious link
-  and any(body.links,
-          (
-            .href_url.domain.root_domain not in $tranco_1m
-            or .href_url.domain.domain in $free_file_hosts
-            or .href_url.domain.root_domain in $free_subdomain_hosts
-            or .href_url.domain.domain in $url_shorteners
-            or 
-
-            // mass mailer link, masks the actual URL
-            .href_url.domain.root_domain in ("hubspotlinks.com", "mandrillapp.com", "sendgrid.net")
-          )
-
-          // exclude sources of potential FPs
-          and .href_url.domain.root_domain not in (
-            "svc.ms",
-            "sharepoint.com",
-            "1drv.ms",
-            "microsoft.com"
-          )
-  )
-
   // not a reply
   and (
     length(headers.references) == 0
     or not any(headers.hops, any(.fields, strings.ilike(.name, "In-Reply-To")))
   )
-
+  
   // Microsoft logo
   and (
     any(attachments,
@@ -40,7 +18,7 @@ source: |
     )
     or any(ml.logo_detect(beta.message_screenshot()).brands, strings.starts_with(.name, "Microsoft"))
   )
-
+  
   // suspicious content
   and (
     (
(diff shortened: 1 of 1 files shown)


detection-rules/link_microsoft_low_reputation.yml
```diff
@@ -104,19 +102,20 @@ source: |
...
-                   any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft")
+                   any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft" and .confidence != "low")
            )
     )
     or (
       any(ml.nlu_classifier(body.html.inner_text).entities, .name == "urgency")
-      and not any(ml.nlu_classifier(body.current_thread.text).intents,
-                  .name == "benign" and .confidence == "high"
+      and any(ml.nlu_classifier(body.current_thread.text).intents,
+                  .name != "benign"
```
jkamdjou, line 111: could this be "unknown"?

detection-rules/link_microsoft_low_reputation.yml
```diff
@@ -104,19 +102,20 @@ source: |
     or any(attachments,
            .file_type in $file_types_images
            and any(file.explode(.),
-                   any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft")
+                   any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft" and .confidence != "low")
```
jkamdjou, line 105: could we be explicit instead: medium, high? i think that's usually clearer to the user

detection-rules/link_microsoft_low_reputation.yml
bfilar: Or could we say if intent is `None`

detection-rules/link_microsoft_low_reputation.yml
```diff
@@ -110,18 +109,18 @@ source: |
...
            and any(file.explode(.),
-                   any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft")
+                   any(ml.nlu_classifier(.scan.ocr.raw).intents, .name == "cred_theft" and .confidence in~ ("medium", "high"))
            )
     )
     or (
       any(ml.nlu_classifier(body.html.inner_text).entities, .name == "urgency")
-      and not any(ml.nlu_classifier(body.current_thread.text).intents,
-                  .name == "benign" and .confidence == "high"
+      and any(ml.nlu_classifier(body.current_thread.text).intents, .name not in~ ("benign", "unknown")
```
jkamdjou, line 117: @aidenmitchell does this need to be an `all` to match our previous behavior? as written, we could get a `benign:high` and `cred_theft:low` and still flag here

aidenmitchell: I don't think we could get a `benign` _and_ `cred_theft` intent, because we're only evaluating one thing, `body.current_thread.text`?

aidenmitchell:
Closing in favour of #791

detection-rules/link_microsoft_low_reputation.yml
jkamdjou: `.intents` is an array because i think we can return multiple with different confidence levels, but @bfilar can confirm

bfilar: so if the model returns `benign` that would be the only intent we return. If `benign` is not present in intent array would could return something like: `(cred_theft,steal_pii)`

PR #782, opened 2023-09-08
New Rule: Reply-to/Sender Mismatch with suspicious TLD

morriscode:


detection-rules/headers_replyto_mismatch_sus_tld.yml
```diff
@@ -0,0 +1,94 @@
+name: "Reply-to/Sender Mismatch with suspicious TLD"
+description: |
+  This rule detects a mismatch between the reply-to and the sender email addresses, and one or both of them are from suspicious TLDs
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and (
+    any(headers.reply_to,
+        .email.email != sender.email.email
```
jkamdjou, line 10: do we wanna check if the domains don't match too? that'd prevent us from firing on stuff like this:

From: sam@foo.agency
Reply-to: support@foo.agency

morriscode:
This can be reworked now that we have suspicious tld list in a better place

morriscode:
This is firing an unwanted amount of times. Closing

PR #784, opened 2023-09-08
Bump MQL mimic test timeout

cameron-dunn-sublime:
Generally 30 minutes is fine but if a bunch of new external function calls are introduced then the tests can be slower.


morriscode (approved):
🕥

PR #785, opened 2023-09-12
New Rule: Attachment PDF with link to dmg file download

morriscode:
Feel free to tweak name/description. No matter how I worded the descript it didn't sound right to me.

### DIFF
--- a/detection-rules/attachment_pdf_link_to_dmg.yml
+++ b/detection-rules/attachment_pdf_link_to_dmg.yml
@@ -62,3 +62,4 @@ detection_methods:
   - "File analysis"
   - "URL analysis"
   
+id: "2c486fe0-506d-5498-bb19-dbe58904f9dc"
--- a/detection-rules/attachment_pdf_link_to_dmg.yml
+++ b/detection-rules/attachment_pdf_link_to_dmg.yml
@@ -28,7 +28,7 @@ source: |
   
                                   // and the zip contains a dmg file
                                   and any(file.explode(.),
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_pdf_link_to_dmg.yml
```diff
@@ -0,0 +1,65 @@
...
+                      // and downloads a dmg or a zip
+                      and any(beta.linkanalysis(.).files_downloaded,
+                              (
+                                .file_extension == "dmg"
+                                or (
+                                  .file_extension in~ $file_extensions_common_archives
+  
+                                  // and the zip contains a dmg file
+                                  and any(file.explode(.),
+                                          any(file.explode(..), .file_extension =~ "dmg")
```
jkamdjou, line 29: nested file explode

detection-rules/attachment_pdf_link_to_dmg.yml
```diff
@@ -0,0 +1,65 @@
...
+  
+                                          // exif inspection if encrypted
+                                          or strings.ends_with(.scan.exiftool.zip_file_name, ".dmg")
```
jkamdjou, line 32: could do an "is encrypted" check before this but i don't think it would actually matter

morriscode: i could go either way. lmk if you want it added. I don't think it changes the functionality either way.

jkamdjou (approved):

PR #790, opened 2023-09-13
FP fix: Update link_credential_phishing_intent_and_other_indicators.yml

morriscode:
Falsely returns true if recipient is null. Adding a valid check


jkamdjou (approved):

morriscode:
Just wanted to highlight the 5 or so regressions caused here.
Several of them are too old and have current_thread and the rule has since been changed to use that.
The other failures are due to the bug of us matching on recipient found in the url, even though the recipient was null. 

Could add a null recipient check as an or condition but that reintroduces the problem

`     (
          length(recipients.to) == 0
          or all(recipients.to, .display_name == "Undisclosed recipients")
        )
        and length(recipients.cc) == 0
        and length(recipients.bcc) == 0
      ),
  )
)`

PR #791, opened 2023-09-13
FP fix: Update link_microsoft_low_reputation.yml

morriscode:
Adding more MS domain to linked domains, and requiring cred_theft of med/high confidence. Was firing on cred_theft low, on images with only a single word "Microsoft"


morriscode:
These regressions do not contains any microsoft specific content, the rule explicitly looks for a MS logo, and the failed regressions are not Microsoft themed at all. These are acceptable regressions.

aidenmitchell (approved):

jkamdjou (approved):

PR #793, opened 2023-09-15
adding signal whois for sender domain

bfilar:


signals/sender/sender_domain_whois_days_old.yml
```diff
@@ -0,0 +1,4 @@
+name: "Sender: Whois on sender domain"
+type: "query"
+source: |
+  beta.whois(sender.email.domain)
```
rw-access, line 4: I think you need to pick a property or two, right? Or can you handle signals like `{days_old: 3.14, found: true, registrar_name: "GoDaddy"}`

bfilar: To get the correct `days_old` value, I was going to grab `created_at`:
```
root_domain:sublimesecurity.com
created_at:2016-10-15T16:55:59Z
updated_at:2022-10-15T18:55:57Z
expires_at:2023-10-15T16:55:59Z
registrar_name:Google LLC
days_old:2529.7324520814623
found:true
```

bfilar: okay, after thinking it through I can deduce days_old for historic msgs in our training via alternative methods. Will go with `beta.whois(sender.email.domain).days_old` to ensure we return a single value

jkamdjou: we can always have more queries to return other fields too

rw-access: `days_old` is always relative to the MDM `_meta.created_at` field. So just make sure that field is right and you're golden.

bfilar: Awesome. Thanks

morriscode (approved):

PR #795, opened 2023-09-18
Updating rule: impersonation_amazon.yml

aidenmitchell:
Negating listservs


morriscode (approved):
Hey I recognize that snippet! Lgtm

PR #798, opened 2023-09-18
Update impersonation_recipient_domain.yml

morriscode:
Changing recipient domain to use mailbox.email.domain.root_domain to negate instances where the recipient is the sender, and delivery is accomplished via BCC's.


jkamdjou:
this might FN on instances where the recipient is an alias or group

morriscode:
> this might FN on instances where the recipient is an alias or group

yeah i wanted to see what mimic found. I had another way to approach it but letting mimic vet the change gets me a list of targets to ensure I cover :)

PR #805, opened 2023-09-20
Fixing FN:  attachment_html_recipient_in_javascript_identifiers.yml

morriscode:
Fixing issue where the rule would fail to flag if the recipient and the javascript terms were found in separate exploded pieces of the same file.

### DIFF
--- a/detection-rules/attachment_html_recipient_in_javascript_identifiers.yml
+++ b/detection-rules/attachment_html_recipient_in_javascript_identifiers.yml
@@ -7,7 +7,7 @@ source: |
   type.inbound
   and any(attachments,
           (
-            .file_extension in~ ("html", "htm", "shtml", "dhtml")
+            .file_extension in~ ("html", "htm", "shtml", "dhtml", "xhtml")
             or (
               .file_extension is null
               and .file_type == "unknown"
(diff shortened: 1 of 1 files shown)


morriscode:
Mimic showing multiple pickups and no regressions. Many of these were only firing "Any HTML" rules.

New Flag. EML 24000 
New Flag. EML 22757 
New Flag. EML 33987 
New Flag. EML 56868
New Flag. EML 3845 
New Flag. EML 20089 
New Flag. EML 6435 
New Flag. EML 5239 
New Flag. EML 20088 
New Flag. EML 22872 
New Flag. EML 20087 
New Flag. EML 23996 
New Flag. EML 34067

detection-rules/attachment_html_recipient_in_javascript_identifiers.yml
```diff
@@ -16,13 +16,18 @@ source: |
...
           and any(file.explode(.),
                   // suspicious identifiers
-                  any(.scan.javascript.identifiers, strings.like(., "atob", "decrypt"))
+                  any([.scan.strings.strings, .scan.javascript.identifiers],
+                      any(., strings.like(., "*atob*", "*decrypt*"))
+                  )
                   // Recipients address found in javascript
-                  and any(recipients.to,
-                          any(..scan.javascript.strings, strings.icontains(., ..email.email))
+                  and any(file.explode(..),
```
jkamdjou, line 27: this is nested inside the other fileexplode is that intentional?

morriscode: yes. this rule should be firing all the time, but in every sample where the condition is true, it will not fire without the double file explode, since the recipients email is found in a different exploded piece resulting in something like this 

image

rw-access: you still should be able to do it without the file explode inside a file explode

rw-access (approved):

PR #806, opened 2023-09-20
New Rule: HTML smuggling with decimal encoding

morriscode:


jkamdjou (approved):
LGTM if test results check out

PR #815, opened 2023-09-22
Sender Profiles: Prevalence and Past behavior 

morriscode:


jkamdjou:
@morriscode are there real regressions here?

morriscode:
I'm assuming there are regressions due to discrepancies between FTS/Unsolicited and the new logic. The test environments have no context for sender profile. Mimic quits after 10. Here are the regressions. I'll check them later tonight or first thing in the morning!

notion.so/EML-37350
notion.so/EML-30853
notion.so/EML-30303
notion.so/EML-34952
notion.so/EML-34832
notion.so/EML-35067
notion.so/EML-34489
notion.so/EML-34617
notion.so/EML-34189
notion.so/EML-35804

jkamdjou:
> I'm guessing this might be caching at least in SharedEML testing... not sure why Mimic would run into caching issues.
> 
> **notion.so/EML-37350** - Not a regression both original rules still fire? What is interesting is they didn't fire until I sent it to editor, the initial MLV showed no rules firing?
> 
> * Body: Business Email Compromise (BEC) attempt from first-time sender (old)
> * Body: Business Email Compromise (BEC) attempt from first-time sender (new)
> 
> **notion.so/EML-30853** - Similar issue as above, except the old rule fires at the MLV, but the new rule doesn't show matching until sending to editor where both rules show matching.
> 
> * Attachment: DocuSign image lure with no DocuSign domains in links (old)
> * Attachment: DocuSign image lure with no DocuSign domains in links (new)
> 
> **notion.so/EML-30303** - Similar issue - Attachment: Any HTML file (unsolicited) fired in the MLV. But 3 rules show firing after sending to editor. 2 of which are the version in this PR.
> 
> * Attachment: Any HTML file (first-time sender) (new)
> * Attachment: Any HTML file (unsolicited) (new)
> * Attachment: Any HTML file (unsolicited) (old)
> 
> cc @jkamdjou @cameron-dunn-sublime before I go any further I suspect this will be true for the remaining though, we should probably figure out what's going on in the above scenarios?
> 
> notion.so/EML-34952 notion.so/EML-34832 notion.so/EML-35067 notion.so/EML-34489 notion.so/EML-34617 notion.so/EML-34189 notion.so/EML-35804

+ @rw-access in case you have any additional context

detection-rules/attachment_adobe_image_lure_fts.yml
```diff
@@ -20,13 +20,10 @@ source: |
...
   and (
-    (
-      sender.email.domain.root_domain in $free_email_providers
-      and sender.email.email not in $sender_emails
-    )
+    profile.by_sender().prevalence in ("new", "outlier")
     or (
-      sender.email.domain.root_domain not in $free_email_providers
-      and sender.email.domain.domain not in $sender_domains
+      profile.by_sender().any_malicious_messages
```
rw-access, line 25: `any_messages_malicious_or_spam` now

rw-access (approved):
LGTM!
comments could be improved but no need to block on that IMO

PR #817, opened 2023-09-23
New Rule: QR Code with suspicious indicators

morriscode:
1075 messages found 
806 Already Flagged

 This is a massive gap filler. It's been running in sharedemls for a few weeks with no FP's.


morriscode:
Numbers reduced to 618 messages found 532 Already Flagged when you add the sender profile. 
However we don't traditionally take actions, so we haven't properly flagged attacks as TP's.

bfilar (approved):

morriscode:
FYI this is on hold for sender_profiles release.

PR #818, opened 2023-09-24
Signals: New link and sender domains

jkamdjou:


morriscode:
Lgtm! Is there any particular reason to be specific rather than just returning the days_old int and evaluating that way?

bfilar (approved):

rw-access:
🚢

PR #819, opened 2023-09-25
Signals from detection rules, batch 1

jkamdjou:
First batch of extracting signals from detection rules

### DIFF
--- a/signals/headers/headers_bounce_back.yml
+++ b/signals/headers/headers_bounce_back.yml
@@ -1,0 +1,8 @@
+name: "Bounce back"
+type: "query"
+source: | 
+  not strings.like(sender.email.local_part, "*postmaster*", "*mailer-daemon*", "*administrator*")
+  and not regex.icontains(subject.subject, "^(undeliverable|read:)")
+  and not any(attachments, .content_type == "message/delivery-status")
+  // if the "References" is in the body of the message, it's probably a bounce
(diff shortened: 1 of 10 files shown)


signals/headers/headers_bounce_back.yml
```diff
@@ -0,0 +1,8 @@
+name: "Bounce back"
+type: "query"
+source: | 
+  not strings.like(sender.email.local_part, "*postmaster*", "*mailer-daemon*", "*administrator*")
+  and not regex.icontains(subject.subject, "^(undeliverable|read:)")
+  and not any(attachments, .content_type == "message/delivery-status")
+  // if the "References" is in the body of the message, it's probably a bounce
+  and not any(headers.references, strings.contains(body.html.display_text, .))
```
morriscode, line 8: This looks like a snippet to exclude bounce-backs. It'll basically fire on everything that's not a bounceback. I think you want the inverse?

```suggestion
  strings.like(sender.email.local_part, "*postmaster*", "*mailer-daemon*", "*administrator*")
  and regex.icontains(subject.subject, "^(undeliverable|read:)")
  and any(attachments, .content_type == "message/delivery-status")
  // if the "References" is in the body of the message, it's probably a bounce
  and any(headers.references, strings.contains(body.html.display_text, .))
```

morriscode: @jkamdjou fysa

PR #822, opened 2023-09-25
Update Rule: Update link_credential_phishing_voicemail_language.yml

morriscode:
Expanding rule to pick up FN's.

Notable improvement. 
225 messages found 
36 Already Flagged

Will run test rules, but hunts in multiple environments look great.


morriscode:
Todo: I think it makes sense to update this from unsolicited to "new", "outlier" in the coming week.

bfilar (approved):

PR #823, opened 2023-09-26
FN: Update attachment_microsoft_image_lure_qr_code.yml

morriscode:
Adding more robust handling when the QR code url does not contain the recipients email.

### DIFF
--- a/detection-rules/attachment_microsoft_image_lure_qr_code.yml
+++ b/detection-rules/attachment_microsoft_image_lure_qr_code.yml
@@ -25,26 +25,24 @@ source: |
                   // recipient email address is present in the URL, a common tactic used in credential phishing attacks 
                   and any(recipients.to,
                           strings.icontains(..scan.qr.data, .email.email)
-                          or (
                           
-                            // the recipients sld is in the senders display name
-                            any(recipients.to,
-                                strings.icontains(sender.display_name, .email.domain.sld)
-                            )
-                            
-                            // the recipient local is in the body  
-                            or any(recipients.to,
-                                   strings.icontains(body.current_thread.text, .email.local_part)
-                            )
-                            
-                            // or the body is null 
-                            or body.current_thread.text is null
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_microsoft_image_lure_qr_code.yml
```diff
@@ -19,12 +19,35 @@ source: |
...
-                .scan.qr.type == "url"
-                // recipient email address is present in the URL, a common tactic used in credential phishing attacks 
-                and any(recipients.to, strings.icontains(..scan.qr.data, .email.email))
+            or (
+              any(file.explode(.),
+                  .scan.qr.type == "url"
+                  // recipient email address is present in the URL, a common tactic used in credential phishing attacks 
+                  and any(recipients.to,
+                          strings.icontains(..scan.qr.data, .email.email)
+                          or (
```
rw-access, line 28: don't need the parentheses here `x or (y or z)` can be flattened to `x or y or z`

detection-rules/attachment_microsoft_image_lure_qr_code.yml
```diff
@@ -19,12 +19,35 @@ source: |
...
+                                strings.icontains(sender.display_name, .email.domain.sld)
+                            )
+                            
+                            // the recipient local is in the body  
+                            or any(recipients.to,
+                                   strings.icontains(body.current_thread.text, .email.local_part)
+                            )
+                            
+                            // or the body is null 
+                            or length(body.current_thread.text) is null
```
rw-access, line 41: length is null doesn't really make sense and will only be null if the field is also null
```suggestion
                            or body.current_thread.text is null
```

detection-rules/attachment_microsoft_image_lure_qr_code.yml
morriscode: Should be fixed!

rw-access (approved):

PR #828, opened 2023-09-28
MQL Mimic: Change to PR CI, detect changes

cameron-dunn-sublime:
The other side of these changes isn't finished out, but this part appears to be working as needed:
* Changed (modified, deleted, new, etc) are reported via rule ID to the trigger endpoint.
* The git SHA reported is the latest commit, e.g. with the rule ID added.
* Branch and repo are reported correctly
    * This set of changes should also unlock getting MQL Mimic tests working on forks.

Once the other pieces are in place, we'll only run MQL mimic tests for detection rules touched in a PR. The endpoint will always be hit, but the workflow on the other side will succeed pretty quickly (it doesn't need to setup the platform or pull its large source).

https://github.com/sublime-security/sublime-rules/pull/828/commits/e2abbf6999af7a2c72e8deb5eacf531f6291c87a shows this running on push (as it will on `main`)

https://github.com/sublime-security/sublime-rules/pull/829 shows it working as it will on a PR.


.github/workflows/rule-validate.yml
```diff
@@ -92,6 +93,46 @@ jobs:
...
+          
+          echo "Altered Ruled IDs: [$altered_rule_ids]"
+          echo "##[set-output name=rule_ids;]$(echo $altered_rule_ids)"
+          # TODO: This doesn't solve for a modified rule_id. We could merge with any files known on 'main', but changing
+          # a rule ID is a separate problem.
+
+      - name: "Trigger MQL Mimic Tests"
+        env:
+          trigger_url: '${{ secrets.MQL_MOCK_TRIGGER }}'
+          branch: ${{ github.event_name == 'pull_request_target' && github.head_ref || github.ref }}
```
cameron-dunn-sublime, line 123: On `main`/etc `github.ref` will include "refs/heads/". I decided to just handle that downstream.

.github/workflows/rule-validate.yml
```diff
@@ -30,6 +30,7 @@ jobs:
         with:
           ref: ${{ github.head_ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
+          depth: 0
```
cameron-dunn-sublime, line 33: This is what's recommended for change detection. It takes 1 second to checkout, so I don't think this is worth even testing the alternatives (which are some higher number, or maybe API based detection but it's unclear if that's compatible with `pull_request_target` workflows)

.github/workflows/rule-validate.yml
```diff
@@ -92,6 +93,45 @@ jobs:
...
+
+      - name: "Find updated rule IDs"
+        id: find_ids
+        run: |
+          for file in ${{ steps.changed-files.outputs.all_changed_and_modified_files }}; do
+            echo "$file was changed"
+            rule_id=$(yq '.id' $file)
+          
+            echo "$file has rule ID $rule_id"
+            altered_rule_ids=$(echo "$rule_id"" ""$altered_rule_ids")
```
hugh-sublime, line 111: This is so cursed `"$rule_id"" ""$altered_rule_ids"` but I don't know what else I expect from bash.

hugh-sublime (approved):

.github/workflows/rule-validate.yml
cameron-dunn-sublime: Haha `"$rule_id $altered_rule_ids" should work too but I'll extra `"` wrap often in case I'm forgetting some expansion that's going to happen.

PR #830, opened 2023-09-29
New Insight - QR code in attachments

morriscode:
image

This insight will extract the embedded URL from QR codes found in image attachments or PDF attachments.


bfilar (approved):
🚀

morriscode:
Thanks @bfilar  now go rest!

PR #831, opened 2023-09-29
Updating rule: attachment_cve_2023_38831.yml

aidenmitchell:
Adding delivr.to fix.


detection-rules/attachment_cve_2023_38831.yml
```diff
@@ -22,19 +22,19 @@ source: |
                   // zip contains a path with spaces and file extensions 
                   // lure.pdf /lure.pdf .cmd
                   // 
-                  //  /= Initial file name
+                  //  /= Initial file name (including any spaces)
```
rw-access, line 25: Is it okay if this includes dots too? Should we do `.` here or `[\w\s]`?

aidenmitchell: @ajpc500 would know better

ajpc500: The recent Agent Tesla file name is `RFQ 23-301892.pdf /RFQ 23-301892.pdf .cmd`. So `[\w\s]` specifically wouldn't match on the hyphens, nor on brackets. I haven't tested it, but I don't think a double-extension (e.g. `.tar.gz`) would have any impact on the ability to exploit CVE-2023-38831, so we could include `.` here. There's potential to refine this to a subset of symbols, but will defer to you re. any rule performance improvement in doing that!

morriscode (approved):

PR #832, opened 2023-09-29
New Rule: Callback Phishing via DocuSign comment

morriscode:


morriscode:
@jkamdjou This is pretty well constrained, it only had a handful of hits in test-data and they look to be valid. fyi

jkamdjou (approved):
ohhh yeah 🔥

PR #844, opened 2023-10-05
Update VIP impersonation rule to ignore org_domains

rw-access:
Looks like there are cases where we call a message `type.inbound` even when the org domain matches. One case is google calendar.

Now that we went from FTS to a more broad definition of first/outlier/untrusted/unsolicited, we're more FP prone. While investigating the `type.inbound` check, this should help resolve those FPs.


cameron-dunn-sublime (approved):
Is it time to revive discussions around `inbound`?

jkamdjou (approved):

PR #846, opened 2023-10-05

aidenmitchell:


detection-rules/link_microsoft_low_reputation.yml
```diff
@@ -32,6 +32,7 @@ source: |
            )
            or any(body.links, .href_url.domain.domain in $free_file_hosts)
          )
+     and .href_url.domain.root_domain not in $org_domains
```
rw-access, line 35:
```suggestion
         and .href_url.domain.root_domain not in $org_domains
```

rw-access (approved):

PR #852, opened 2023-10-10
Fixing insight naming to match UI and MDM

aidenmitchell:


jkamdjou (approved):

aidenmitchell:
Sorry @jkamdjou, there was 1 more to rename. Thanks!

jkamdjou:
> Sorry @jkamdjou, there was 1 more to rename. Thanks!

how about if we said "Sender (From)"? since folks more familiar with raw EMLs will recognize that?

jkamdjou (approved):

PR #855, opened 2023-10-13
Updating rule: attachment_any_html_unsolicited.yml

aidenmitchell:
Fixing FPs, rule will only flag if the sender isn't known to FP _and_ is unsolicited.

### DIFF
--- a/detection-rules/attachment_any_html_unsolicited.yml
+++ b/detection-rules/attachment_any_html_unsolicited.yml
@@ -12,11 +12,11 @@ source: |
   type.inbound
   and any(attachments, .file_extension in~ ('htm', 'html') or .file_type == "html")
   and (
-    (
+    not profile.by_sender().any_false_positives
+    and (
       profile.by_sender().any_messages_malicious_or_spam
-      and not profile.by_sender().any_false_positives
+      or not profile.by_sender().solicited
     )
-    and not profile.by_sender().solicited
   )
 tags:
   - "Attack surface reduction"
--- a/detection-rules/attachment_any_html_unsolicited.yml
+++ b/detection-rules/attachment_any_html_unsolicited.yml
@@ -13,10 +13,7 @@ source: |
   and any(attachments, .file_extension in~ ('htm', 'html') or .file_type == "html")
   and (
     not profile.by_sender().any_false_positives
-    and (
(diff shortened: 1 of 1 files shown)


jkamdjou (commented):
this will now only fire if there's been malicious messages. i think we want to drop line 16

detection-rules/attachment_any_html_unsolicited.yml
```diff
@@ -12,11 +12,11 @@ source: |
   type.inbound
   and any(attachments, .file_extension in~ ('htm', 'html') or .file_type == "html")
   and (
-    not profile.by_sender().solicited
-    or (
+    (
       profile.by_sender().any_messages_malicious_or_spam
```
rw-access, line 16: this doesn't make sense. You will now only flag on senders that have already flagged before.

rw-access (changes requested):

madirey (approved):

detection-rules/attachment_any_html_unsolicited.yml
```diff
@@ -12,10 +12,10 @@ source: |
   type.inbound
   and any(attachments, .file_extension in~ ('htm', 'html') or .file_type == "html")
   and (
-    not profile.by_sender().solicited
-    or (
+    not profile.by_sender().any_false_positives
```
rw-access, line 15: We need to keep unsolicited as a requirement

aidenmitchell: I think this is fixed? Feel free to just overwrite my changes.

madirey: It looks fixed to me, what am I missing?

aidenmitchell: Nothing, it's just that my brain isn't active yet... so the odds of an error are higher that usual 😛

madirey: Oh sorry, I was just making sure @rw-access doesn't have an additional concern with the above. The rule looks like it's keeping unsolicited as a requirement, so I was confused by the comment.

rw-access: looks good now!

rw-access (approved):
thanks for you patience/endurance!

PR #862, opened 2023-10-17
New rule: impersonation_quickbooks.yml

aidenmitchell:


morriscode:
fired 47 times no other rule overlap for 90% of them. Going to hold off on merging as the impersonation rules are prone to trusted domain firings.

morriscode:
Can we add the highly trusted domain logic here @aidenmitchell

morriscode (approved):
🚕

PR #866, opened 2023-10-18
Update link_google_translate.yml

morriscode:
Updating to use the google_translate_open_redirect encoder.

### DIFF
--- a/detection-rules/link_google_translate.yml
+++ b/detection-rules/link_google_translate.yml
@@ -10,7 +10,10 @@ authors:
 severity: "low"
 source: |
   type.inbound
-  and any(body.links, any(.href_url.rewrite.encoders, strings.contains(., "google_translate_open_redirect")))
+  and any(body.links,
+      .href_url.domain.root_domain == "translate.goog"
+      or "google_translate_open_redirect" in .href_url.rewrite.encoders
+  )
   and (
     not profile.by_sender().solicited
     or (
--- a/detection-rules/link_google_translate.yml
+++ b/detection-rules/link_google_translate.yml
@@ -10,10 +10,7 @@ authors:
 severity: "low"
 source: |
   type.inbound
-  and any(body.links,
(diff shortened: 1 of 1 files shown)


detection-rules/link_google_translate.yml
```diff
@@ -10,7 +10,7 @@ authors:
 severity: "low"
 source: |
   type.inbound
-  and any(body.links, .href_url.domain.root_domain == "translate.goog")
+  and any(body.links, any(.href_url.rewrite.encoders, strings.contains(., "google_translate_open_redirect")))
```
rw-access, line 13: We should still support old ones for backwards compatibility. Also we shouldn't be doing `strings.contains` on an enum and should only be doing `in` or `==`, otherwise you lose extra detection
```suggestion
  and any(body.links,
      .href_url.domain.root_domain == "translate.goog"
      or "google_translate_open_redirect" in .href_url.rewrite.encoders
  )
```

also FYI @hugh-sublime we don't have an editor diagnostic to help with this particular case

morriscode:
@rw-access I don't know if we want to support the backwards compatibility with this one specifically. 
image

The previous logic was prone to false positives even if the query params were clearly not a URL, the encoder seems to take into consideration if the translate.goog contains a url.

detection-rules/link_google_translate.yml
```diff
@@ -10,7 +10,10 @@ authors:
 severity: "low"
 source: |
   type.inbound
-  and any(body.links, .href_url.domain.root_domain == "translate.goog")
+  and any(body.links,
+      .href_url.domain.root_domain == "translate.goog"
+      or "google_translate_open_redirect" in .href_url.rewrite.encoders
+  )
```
rw-access, line 16: sweet! well in that case...
```suggestion
  and any(body.links, "google_translate_open_redirect" in .href_url.rewrite.encoders)
```

rw-access (approved):

PR #867, opened 2023-10-18
New Insight: Open Redirect Destinations

morriscode:
image


insights/links/open_redirect_destinations.yml
```diff
@@ -0,0 +1,7 @@
+name: "Open redirect effective URLs"
+type: "query"
+source: |
+  map(filter(body.links, any(.href_url.rewrite.encoders, strings.ends_with(., "_open_redirect"))), .href_url.url)
```
rw-access, line 4: here's the full list btw. Might want to add a few more for compatibility. Maybe just `adobe`? We can also rename that one to `adobe_open_redirect`. Also `google_amp`

`adobe,barracuda,bing_open_redirect,checkpoint,google_amp,google_open_redirect,google_translate_open_redirect,indeed_open_redirect,microsoft,proofpoint,sublime,trend_micro`

morriscode: it'll pick up any that end in _open_redirect, so as long as the naming standard is proper we should be future proof too?

jkamdjou: we should be _prefixing_ these with `open_redirect_` instead of appending to make it more consistent

cc: @hugh-sublime

morriscode:
converting to draft to wait on naming alignment.

aidenmitchell (approved):

PR #868, opened 2023-10-18
Update regex in rules to remove line-based matches

rw-access:
Line-based tokens `^` and `$` don't work for backtesting. We should be using `regex.match` in these cases anyway.


morriscode (approved):
💲

PR #870, opened 2023-10-19
Add PDF to QR code attachment detection

rw-access:
Some misses otherwise! We still need to get the QR code parsing to work without a protocol for some of the misses, that's a separate thing that doesn't need to happen here


aidenmitchell (approved):

morriscode (approved):
👍

PR #874, opened 2023-10-24
Improved Filtering for Changed Rules

cameron-dunn-sublime:
Old logic didn't understand changes that weren't affecting rule logic, such as renames and tag changes. I moved further off of `tj-actions/changed-files` because I'm not sure how well it's handling `pull_request_target` events and it just generally is more obfuscated than what we're doing now -- mostly just comparing `source`.


morriscode (approved):
👏 This looks great to me.

PR #876, opened 2023-10-25
Update callback_phishing_nlu_body_or_attachments.yml

morriscode:
Changing exiftools to check for "Model" this is present whenever the image is taken from a mobile device.


morriscode:
@bobby adding android screenshot negation and changing .key requirement to just "Model" vs specifically iPhones.

jkamdjou (approved):

PR #882, opened 2023-10-26
New Rule:  Attachment: DocuSign Impersonation (PDF) linking to New Domain <=3d

morriscode:


detection-rules/attachment_pdf_docusign_impersonation_new_domain.yml
```diff
@@ -0,0 +1,38 @@
...
+          and any(ml.logo_detect(.).brands, .name == "DocuSign")
+          and any(file.explode(.), any(.scan.pdf.urls, beta.whois(.domain).days_old <= 3))
+  )
+  and (
+    sender.email.domain.root_domain not in ('docusign.net', 'docusign.com')
+    // check for DMARC fail for spoofs
+    or any(distinct(headers.hops, .authentication_results.dmarc is not null),
+           strings.ilike(.authentication_results.dmarc, "*fail")
+    )
+  )
```
jkamdjou, line 18: trying to wrap my head around this, is this def right?

detection-rules/attachment_pdf_docusign_impersonation_new_domain.yml
```diff
@@ -0,0 +1,38 @@
...
+  // and the sender.display.name contains "via"
```
jkamdjou, line 19:
```suggestion
  // excludes senders that contain "via" in the display name a resilient way
```

jkamdjou (approved):

PR #889, opened 2023-10-27
Include prevalence in login page detection

jkamdjou:


detection-rules/attachment_html_attachment_login_page.yml
```diff
@@ -69,7 +69,10 @@ source: |
           )
   )
   and (
-    not profile.by_sender().solicited
+    (
+      not profile.by_sender().solicited
+      and not profile.by_sender().prevalence in ("new", "outlier")
```
rw-access, line 74: This right? Has to be unsolicited _and old_?

jkamdjou: right

morriscode: so this won't ever fire on a new sender, the sender basically has to be common and unsolicited? I'm not sure i'm following the logic.

jkamdjou: i was clearly sleep deprived when i wrote this. updating

jkamdjou, line 74:
```suggestion
      and profile.by_sender().prevalence in ("new", "outlier")
```

morriscode (approved):
✔️

PR #908, opened 2023-11-03

morriscode:
Adding constant contact to mass mailer links (rs6.net). Also adding an additional or condition to the //suspicious content block to check message screenshot. 

This is required in the event that an image type cannot be rendered. Recent FP's have shown images with compromised integrity, which still render visually, but do not produce OCR.


jkamdjou (approved):

jkamdjou:
to be clear this doesn't resolve any FPs right? just FNs?

morriscode:
> to be clear this doesn't resolve any FPs right? just FNs?

100%

PR #910, opened 2023-11-03
New Rule: Suspicious File share via link or html attachment

morriscode:


detection-rules/credential_phishing_suspicious_file_share_link_or_html.yml
```diff
@@ -0,0 +1,62 @@
...
+        or any([beta.linkanalysis(.)],
+               (
+                 .credphish.disposition == "phishing"
+                 and .credphish.brand.confidence in ("medium", "high")
+               )
+               or any(file.explode(.screenshot),
+                      strings.contains(.scan.ocr.raw, "link has been removed")
+               )
+        )
+        or .href_url.domain.tld in $suspicious_tlds
```
jkamdjou, line 28:
```suggestion
        or (.href_url.domain.tld in $suspicious_tlds and sender.email.domain.tld not in $suspicious_tlds)
```
defer to you if this is worth it

detection-rules/credential_phishing_suspicious_file_share_link_or_html.yml
```diff
@@ -0,0 +1,62 @@
...
+    )
+  )
+  and (
+    (
+      profile.by_sender().prevalence in ("new", "outlier")
+      and not profile.by_sender().solicited
+    )
+  )
+
+
```
jkamdjou, line 48: high_trust_senders here too?

jkamdjou (commented):
couple suggestions otherwise LGTM if test results look good

detection-rules/credential_phishing_suspicious_file_share_link_or_html.yml
morriscode: high trust senders actually breaks a bunch of the uncovered results, from cloud shares like sharepoint or google
This will flag on sharepoint links that have been deleted too. 

Depending on test-results it might have to be added :)

jkamdjou: ahh good point... we could allow those specific ones we've seen if having it otherwise provides value

morriscode:
still not seeing this in the test rules table.

PR #911, opened 2023-11-03
Update impersonation_human_resources.yml

morriscode:
Adding cred_theft intent

### DIFF
--- a/detection-rules/impersonation_human_resources.yml
+++ b/detection-rules/impersonation_human_resources.yml
@@ -8,13 +8,28 @@ source: |
   and regex.icontains(sender.display_name,
                       '(\bh\W?r\W?\b|human resources|hr depart(ment)?|employee relations)'
   )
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_human_resources.yml
```diff
@@ -8,13 +8,28 @@ source: |
...
+      )
+      or any(ml.nlu_classifier(body.current_thread.text).intents,
+             .name in ("cred_theft")
+             and .confidence in ("medium", "high")
+             and (
+               any(ml.nlu_classifier(body.current_thread.text).intents,
+                   .name != "benign"
+               )
+               or length(ml.nlu_classifier(body.current_thread.text).intents) == 0 // not benign but not malicious either
+             )
```
jkamdjou, line 30: what's the intent behind this?

morriscode: Cc @aidenmitchell I think Aiden added that because we saw a null intent, which also doesn't fire !=benign

PR #933, opened 2023-11-07
New Rule: Link Quickbooks Image lure with suspicious link

morriscode:


jkamdjou (approved):
LGTM if testing is good

PR #936, opened 2023-11-07
Update body_extortion.yml

morriscode:
Changing to unsolicited
The DMARC check wasn't checking positionally anyway, so I just yanked it.

I might update this a couple more times before deploying, running some hunts to see if we want to incorporate dmarc checks in a different way.

### DIFF
--- a/detection-rules/body_extortion.yml
+++ b/detection-rules/body_extortion.yml
@@ -16,10 +16,8 @@ source: |
     )
   )
   and (
-    (
       not profile.by_sender().any_false_positives
       and not profile.by_sender().solicited
-    )
   )
 attack_types:
   - "Extortion"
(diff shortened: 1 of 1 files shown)


bfilar (approved):

detection-rules/body_extortion.yml
```diff
@@ -8,22 +8,16 @@ severity: "low"
...
-      and sender.email.domain.domain not in $sender_domains
+    and any(ml.nlu_classifier(body.current_thread.text).entities,
+            .name == "financial"
     )
-    // many extortion emails spoof sender domains and fail sender authentication
-    or any(headers.hops, .authentication_results.dmarc == "fail")
+  )
+  and (
+      not profile.by_sender().any_false_positives
+      and not profile.by_sender().solicited
```
jkamdjou, line 20: i think this opens us up to FPs a little bit, do we wanna add prevalence?

morriscode: Yeah Im not done with this one yet. I started some hunts to flesh it out

PR #957, opened 2023-11-14
New Rule: Callback Phishing Invoice from < 30d domains

morriscode:


morriscode:
No hits in testing in the last 2 days, this was in response to reported FN's so I don't expect high volume, but this works when the very specific situation is encountered. In terms of FP's I'd expect the risk to be low.

detection-rules/bec_invoice_fraud_30d_domains.yml
```diff
@@ -0,0 +1,70 @@
...
+             strings.icontains(.scan.ocr.raw, "lifelock"),
+             strings.icontains(.scan.ocr.raw, "best buy"),
+             strings.icontains(.scan.ocr.raw, "mcafee"),
+             strings.icontains(.scan.ocr.raw, "norton"),
+             strings.icontains(.scan.ocr.raw, "ebay"),
+             strings.icontains(.scan.ocr.raw, "paypal"),
+             strings.icontains(.scan.ocr.raw, "virus"),
+           )
+    )
+  )
```
jkamdjou, line 60: let's add unsolicited + no FPs?

detection-rules/bec_invoice_fraud_30d_domains.yml
```diff
@@ -0,0 +1,70 @@
...
+attack_types:
+  - "BEC/Fraud"
```
jkamdjou, line 62: just confirming, is this not callback phishing given the brands were looking for?

detection-rules/bec_invoice_fraud_30d_domains.yml
```diff
@@ -0,0 +1,70 @@
+name: "BEC/Fraud - Invoice from sender, reply-to or return-path domain less than 30d old."
```
jkamdjou, line 1:
```suggestion
name: "BEC/Fraud: Branded invoice from sender/reply-to domain less than 30 days old"
```

detection-rules/bec_invoice_fraud_30d_domains.yml
jkamdjou: @morriscode this is resolved but i don't see it in the rule, did you mean to do it or nah?

jkamdjou (approved):

detection-rules/bec_invoice_fraud_30d_domains.yml
morriscode: You right. Should be actually resolved now.

morriscode:
Testing on this looks good, only a single hit.

PR #964, opened 2023-11-15
New Rule: Brand Impersonation: Chase with credential phishing indicators

morriscode:


jkamdjou (approved):
could add unsolicited to be safe but defer to you

PR #990, opened 2023-11-17
New Rule: Discovery Credential/Key leakage

morriscode:


detection-rules/discovery_credential_leakage.yml
```diff
@@ -0,0 +1,64 @@
+name: "Discovery - Potential Credential and Key Leakage"
+description: "This rule looks for patterns related to Credentials, or Keys in both inbound and outbound message bodies."
```
jkamdjou, line 2:
```suggestion
description: "This rule looks for patterns related to Credentials or Keys in both internal and outbound message bodies."
```

detection-rules/discovery_credential_leakage.yml
```diff
@@ -0,0 +1,64 @@
...
+        strings.contains(.href_url.url, ".s3.")
+        and regex.contains(.href_url.query_params,
+                           "Credential=((?:A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA)[A-Z0-9]{16})"
+        )
+    )
+  )
+attack_types:
+
+tactics_and_techniques:
+
```
jkamdjou, line 61:
```suggestion
```

PR #992, opened 2023-11-18
New rule - Brand impersonation: Sharepoint fake file share

jkamdjou:


morriscode (approved):
lgtm. I'll approve and we can wait until we have something in test-rules if you want?
I just checked and there's no hits for today yet.

PR #1004, opened 2023-11-20
New Rule: BEC/Fraud: PenPal Scam

morriscode:


morriscode:
This is super specific hunts return the only know bad and no fps

detection-rules/bec_fraud_penpal_scam.yml
```diff
@@ -0,0 +1,57 @@
+name: "BEC/Fraud: PenPal Scam"
```
jkamdjou, line 1:
```suggestion
name: "PenPal Scam"
```

jkamdjou (approved):
ideas in case you want to consider adding, defer to you:
- NLU entity
- no links or attachments

bfilar:
@morriscode perhaps not required, but a:

```
any(ml.nlu_classifier(body.current_thread.text).entities, .name == "request")
```

may work as well

PR #1012, opened 2023-11-21
New Rule: PDF with embedded Javascript - recreation of pr#646

morriscode:


morriscode:
Attached the historic Notion matches above. This rule should be good to go since it never fired in test rules, but does match known samples, historically.

jkamdjou (approved):

PR #1015, opened 2023-11-21
New Rule: Brand impersonation: Microsoft with credential theft language

morriscode:


detection-rules/impersonation_microsoft_credential_theft.yml
```diff
@@ -0,0 +1,56 @@
+name: "Brand impersonation: Microsoft with logo and credential theft language"
```
jkamdjou, line 1:
```suggestion
name: "Brand impersonation: Microsoft with embedded logo and credential theft language"
```

jkamdjou (approved):

PR #1031, opened 2023-11-22
Ross.html screenshot fixes

rw-access:
FYI @aidenmitchell.

I didn't look at any other rules to check for invalid usage of `..` inside `file.explode(.)`.
 Remember that `..` inside a `file.explode` is referring to the _attachment_, not the exploded file. In the archive case, the attachment is the archive file.

### DIFF
--- a/detection-rules/attachment_fake_slack_installer.yml
+++ b/detection-rules/attachment_fake_slack_installer.yml
@@ -12,36 +12,17 @@ source: |
           or .file_type == "html"
           or .content_type == "text/html"
         )
+        any(ml.logo_detect(file.html_screenshot(..)).brands,
+            .name == "Slack" and .confidence in ("medium", "high")
+        )
+        and any(ml.nlu_classifier(file.parse_html(..).display_text).entities,
+                .name == "request" and .text =~ "download"
+        )
         and any(file.explode(.),
-                any(ml.logo_detect(file.html_screenshot(..)).brands,
-                    .name == "Slack" and .confidence in ("medium", "high")
-                )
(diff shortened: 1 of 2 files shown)


detection-rules/attachment_fake_slack_installer.yml
```diff
@@ -12,37 +12,18 @@ source: |
...
           or .content_type == "text/html"
         )
+        any(ml.logo_detect(file.html_screenshot(..)).brands,
+            .name == "Slack" and .confidence in ("medium", "high")
+        )
+        and any(ml.nlu_classifier(file.parse_html(..).display_text).entities,
+                .name == "request" and .text =~ "download"
+        )
         and any(file.explode(.),
-                any(ml.logo_detect(file.html_screenshot(..)).brands,
```
rw-access, line 16: This will evaluate the exact same every time so shouldn't be in the loop.

detection-rules/attachment_fake_zoom_installer.yml
```diff
@@ -12,37 +12,17 @@ source: |
...
-    )
-    or any(attachments,
-           (.file_extension in~ $file_extensions_common_archives)
-           and any(file.explode(.),
-                   (
-                     .file_extension in~ ("html", "htm", "shtml", "dhtml")
-                     or ..file_type == "html"
-                     or ..content_type == "text/html"
-                   )
-                   and any(ml.logo_detect(file.html_screenshot(..)).brands,
```
rw-access, line 35: Removed this entirely. We currently can't look at raw files in archives. This one is referencing the archive attachment.

aidenmitchell (approved):

PR #1041, opened 2023-11-23
Remove .days_known for now

jkamdjou:


rw-access (approved):
Sorry! I'll fix it soon

PR #1049, opened 2023-11-27
FP fix: file_sharing_link_suspicious_subject.yml

aidenmitchell:
Was catching Dropbox communications, fixing.


jkamdjou (approved):
we might also want to add high trust to this

PR #1063, opened 2023-11-30
New Rule: Link to auto-downloaded DMG in archive

morriscode:


morriscode:
Added a bit more gating on this, I'm guessing we see at least some degree of this happening non-maliciously... Probably smart to let test-rules run

jkamdjou (approved):
LGTM if we add the ASR tag

PR #1074, opened 2023-12-01
Update rules, removing prevalence

sumeetjain:


morriscode (approved):
These looks great @sumeetjain 👌  Thank you!
Validated all parse appropriately. 

I see `not profile.by_sender().solicited` added to all of them, when it previously was not part of some, and some used "first time sender". 

Just making sure that's intentional and not a copy/paste cling on of another snippet. 
If it's intentional all good, just wanted to check

jkamdjou (approved):

PR #1075, opened 2023-12-01
New rule - Reconaissance: All recipients bcc'd or undisclosed

jkamdjou:
breaking out part of #1072 to mitigate some FPs and be more targeted

### DIFF
--- a/detection-rules/recon_recipients_bccd_undisclosed.yml
+++ b/detection-rules/recon_recipients_bccd_undisclosed.yml
@@ -34,6 +34,7 @@ source: |
 
 tags:
   - "Attack surface reduction"
+  - "Deliverability testing"
 attack_types:
   - "Spam"
 detection_methods:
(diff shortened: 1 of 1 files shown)


detection-rules/recon_recipients_bccd_undisclosed.yml
```diff
@@ -0,0 +1,43 @@
+name: "Reconaissance: All recipients bcc'd or undisclosed"
+description: |
+  Recon messages are used to validate whether a recipient address is valid or not, potentially preceding an attack.
```
morriscode, line 3:
```suggestion
  Recon messages, a form of deliverability testing, are used to validate whether a recipient address is valid or not, potentially preceding an attack.
```

Just trying to work in the buzzwords, could adopt here or in the tags. I'll suggest both

detection-rules/recon_recipients_bccd_undisclosed.yml
```diff
@@ -0,0 +1,43 @@
...
+        any(distinct(headers.hops, .authentication_results.dmarc is not null),
+            strings.ilike(.authentication_results.dmarc, "*fail")
+        )
+      )
+    )
+    or sender.email.domain.root_domain not in $high_trust_sender_root_domains
+  )
+
+tags:
+  - "Attack surface reduction"
```
morriscode, line 36:
```suggestion
  - "Attack surface reduction"
  - "Deliverability testing"
```

morriscode (approved):

PR #1081, opened 2023-12-04
Update body_extortion.yml

morriscode:
Adding Compauth failures to cover known FN's

### DIFF
--- a/detection-rules/body_extortion.yml
+++ b/detection-rules/body_extortion.yml
@@ -8,10 +8,12 @@ severity: "low"
 source: |
   type.inbound
   and (
-  any(ml.nlu_classifier(body.current_thread.text).intents,
-      .name == "extortion" and .confidence == "high"
-  )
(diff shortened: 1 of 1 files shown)


jkamdjou (approved):

detection-rules/body_extortion.yml
```diff
@@ -22,8 +24,12 @@ source: |
       sender.email.domain.root_domain not in $free_email_providers
       and sender.email.domain.domain not in $sender_domains
     )
+  
     // many extortion emails spoof sender domains and fail sender authentication
-    or any(headers.hops, .authentication_results.dmarc == "fail")
+    or any(headers.hops,
+           .authentication_results.dmarc == "fail"
+           or .authentication_results.compauth.verdict not in ("pass", "softpass")
```
jkamdjou, line 31: @morriscode what happens if there's no SPF/DMARC configured in this case? we could check the explicit failure conditions instead if we're worried about that

PR #1083, opened 2023-12-05
Changes to Docusign rule to reduce false positives.

aslefhewqiwbepqwefbpqsciwueh:
Hi,

This rule was causing false positives with docusign marketing emails. I suggest two changes. 

Check all the links for docusign domain instead of any:
```
and all(body.links,
          not strings.ilike(.href_url.domain.root_domain, "docusign.*")
  )
``` 

Check the sender is not Docusign itself:
```
and not strings.like(sender.email.domain.root_domain, "docusign.com")
```

Many thanks!

### DIFF
--- a/detection-rules/credential_phishing_docusign_embedded_image_lure.yml
+++ b/detection-rules/credential_phishing_docusign_embedded_image_lure.yml
@@ -34,10 +34,6 @@ source: |
       and not profile.by_sender().any_false_positives
     )
   )
-
-  //negate sender from Docusign
-  and not strings.like(sender.email.domain.root_domain, "docusign.com")
-
   // negate highly trusted sender domains unless they fail DMARC authentication
   and
   (
--- a/detection-rules/credential_phishing_docusign_embedded_image_lure.yml
+++ b/detection-rules/credential_phishing_docusign_embedded_image_lure.yml
@@ -34,6 +34,10 @@ source: |
       and not profile.by_sender().any_false_positives
     )
   )
+
+  //negate sender from Docusign
+  and not strings.like(sender.email.domain.root_domain, "docusign.com")
+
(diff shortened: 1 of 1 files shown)


detection-rules/credential_phishing_docusign_embedded_image_lure.yml
```diff
@@ -34,6 +34,10 @@ source: |
       and not profile.by_sender().any_false_positives
     )
   )
+
+  //negate sender from Docusign
+  and not strings.like(sender.email.domain.root_domain, "docusign.com")
+
```
morriscode, line 40:
```suggestion
```

I've added "docusign.com" to our $high_trust_sender_root_domains. Which should accomplish the intent yet still incorporates a dmarc failure check to prevent spoofed docusign senders.

detection-rules/credential_phishing_docusign_embedded_image_lure.yml
```diff
@@ -5,7 +5,7 @@ severity: "high"
 source: |
   type.inbound
   and length(attachments) <= 1
-  and any(body.links,
+  and all(body.links,
```
morriscode, line 8: ### 
```suggestion
  and length(body.links) > 0
  and all(body.links,
```

Adding a length check to precede the `all(body.links` check to prevent a false truthin the event the array is empty 


all returns true if all of the values in the original array match an expression, which is provided by the second argument. If the array is empty, then all is vacuously `true. [Reference](https://docs.sublimesecurity.com/docs/functions#all)

morriscode (approved):
Thanks @aslefhewqiwbepqwefbpqsciwueh  
Looks great, much appreciated!

aslefhewqiwbepqwefbpqsciwueh:
Many thanks! 🙏

PR #1084, opened 2023-12-05
Fixing $org_domains case sensitivity

aidenmitchell:
Fixing case-sensitivity. $org_domains can sometimes contain capital letters.


rw-access:
Before this one merges, let's investigate "$org_domains can sometimes contain capital letters.". That shouldn't be the case. Might be, but if it is, we just need to fix it there. We always lowercase domains otherwise

morriscode (dismissed):

PR #1086, opened 2023-12-05
Update open_redirect_youtube.yml

morriscode:
Changing this to a more robust negation to mitigate FP's


aidenmitchell (commented):
LGTM, but would like a 2nd pair of eyes.

jkamdjou (approved):
this could probably use high trust + the unsolicited snippet too

morriscode:
> this could probably use high trust + the unsolicited snippet too

Good call! Added

PR #1087, opened 2023-12-05

morriscode:
Swapping prevalence for unsolicited


jkamdjou (approved):

jkamdjou:
we should run some Hunts w/ this change if we haven't already

morriscode:
> we should run some Hunts w/ this change if we haven't already

Had issues with the hunts timing out,  I'm guessing we'll get maybe 3d hunts to complete if we're lucky.

PR #1088, opened 2023-12-05

morriscode:
wrapping 'legal' in boundaries. Should probably consider doing this to all the single words in the list.


rw-access (commented):
> Should probably consider doing this to all the single words in the list.

definitely. `HR` and a few other short ones could use it too

morriscode:
> > Should probably consider doing this to all the single words in the list.
> 
> definitely. `HR` and a few other short ones could use it too

Exactly I know we did HR in another rule using this list, it's tough to keep them in sync.

jkamdjou (approved):

PR #1093, opened 2023-12-06
Update link_suspicious_message_unscannable_cloudflare.yml

morriscode:
limiting links to avoid FP's


jkamdjou (approved):
we wanna also exclude bounce backs?

morriscode:
> we wanna also exclude bounce backs?

For sure, was just doing quick updates while tuning, combined both tunes to this PR

PR #1099, opened 2023-12-06
Update link_suspicious_message_unscannable_cloudflare.yml

aidenmitchell:
Adding additional Cloudflare blocking detection.


morriscode:
Looks like this rule has since been expanded. I don't think this is needed any longer if you disagree please reopen @aidenmitchell

PR #1104, opened 2023-12-06
Update inline_image_as_message.yml

morriscode:
Moving the link req outside of the or'd condition. This is done to mitigate FP's. Would like to run some comparative hunts before merging


morriscode:
After reviewing hunts, the original version of this rule had a lot of unintended collateral, the proposed version seems to reflect the intent of a hyperlinked image, while also mitigating FP's.

jkamdjou (approved):

PR #1114, opened 2023-12-08
New rule: open_redirect_newegg.yml

aidenmitchell:


detection-rules/open_redirect_newegg.yml
```diff
@@ -0,0 +1,18 @@
+name: "Open redirect: Newegg"
+description: |
+  Message contains use of the Newegg open redirect, but the sender is not Newegg. This has been exploited in the wild.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and any(body.links, .href_url.domain.root_domain == 'newegg.com' and .href_url.path =~ '/rts/go2.aspx')
```
zoomequipd, line 8: looks like the `x` url param is what is used for the redirect website. might include that value?

zoomequipd (changes requested):
Doesn't look like redirects currently work (I assume NewEgg fixed it)  but could still provide value should it be used in the future, or another redirect is discovered. 

consider adding the `x=` to the query_params for higher confidence.

zoomequipd (changes requested):
Doesn't look like redirects currently work (I assume NewEgg fixed it)  but could still provide value should it be used in the future, or another redirect is discovered. 

consider adding the `x=` to the query_params for higher confidence.

zoomequipd (approved):

PR #1120, opened 2023-12-12

morriscode:
Adding negation for bounceback content_types being the only attachment


aidenmitchell (approved):

jkamdjou:
i think we can close this since #1147 just merged

PR #1124, opened 2023-12-13
FP fix: attachment_office365_image.yml

aidenmitchell:
We'll want Mimic to run here, to be sure this check is ok.


aidenmitchell:
All the Mimic flags are not as a result of this change, so I'm ok with pushing this.
My guess is that, because Mimic hasn't been flagging for a while, these regressions are from an older change to this rule.

PR #1125, opened 2023-12-13
FP fix: attachment_html_smuggling_atob.yml

aidenmitchell:
Our `ilike` check was a little too broad, requiring only `=`, `atob`, and `;` _anywhere_ in a string. This fix now checks to ensure the `=` and `atob` are close to one another.

Also adds high trust for good measure.


aidenmitchell:
One miss, and it fires 5 other rules, I think that's ok.

morriscode (approved):

PR #1136, opened 2023-12-13
FP fix: impersonation_microsoft.yml

aidenmitchell:
Adding distance check to regex.


morriscode:
Yo! I think is 5 is pretty conservative. 

For example via + surrounding spaces would be 5 chars. I'd probably expand to 10-20ish just to account for word substitutions or padding, etc.

jkamdjou (approved):

PR #1137, opened 2023-12-14
Updating recipient.to checks to ensure domain is valid

aidenmitchell:


bfilar (approved):
Nice work!

morriscode (approved):

morriscode:
Couple mimic failures after the last push. If you review those i think this is gtg. 
@aidenmitchell

morriscode:
cc @aidenmitchell  looks like we got some conflicts now :/

PR #1141, opened 2023-12-14
Negating replies & requiring subdomain

aidenmitchell:


detection-rules/link_login_or_captcha.yml
```diff
@@ -26,6 +32,11 @@ source: |
           and beta.linkanalysis(.).effective_url.domain.domain not in ("login.squarespace.com")
   )
 
+  // negate legit replies
```
morriscode, line 35: Could we avoid this if we swapped prevalence for pure unsolicited checks?

PR #1148, opened 2023-12-15
Adding link count check

aidenmitchell:


morriscode (approved):

aidenmitchell:
/mql-mimic-exempt 178682

we don't need to fire on this.

PR #1149, opened 2023-12-15

aidenmitchell:
"team" is too loose, removing.


jkamdjou (approved):
we use this snippet in a few other rules as well, we should prob update them all to keep em in sync

PR #1158, opened 2023-12-15
New insight: body_qr_code_link.yml

aidenmitchell:


morriscode:
image

morriscode:
This appears to work as intended, however it could cause duplicate entries as there's no attachment length check. 
In addition, I've ran into some inconsistencies with the QR code being identified in screenshots.  I'll approve this for now, but I think we may need to refine.

PR #1160, opened 2023-12-19
New Rule: Malware Bazaar hash in archive

morriscode:


detection-rules/attachment_malwarebazaar_hash_in_archive.yml
```diff
@@ -0,0 +1,32 @@
+name: "MalwareBazaar: Malicious attachment hash in archive (trusted reporters)"
+description: "Detects if an arhive attachments contains a file that matches a SHA256 hash reported as malware on MalwareBazaar by trusted reporters from first-time senders."
```
jkamdjou, line 2:
```suggestion
description: "Detects if an arhive attachments contains a file that matches a SHA256 hash reported as malware on MalwareBazaar by trusted reporters from unknown senders."
```

jkamdjou (approved):

PR #1164, opened 2023-12-19
Update impersonation_sharepoint_body_credential_theft.yml

aidenmitchell:
2nd idea, adding suspicious subject check.


morriscode:
cc @aidenmitchell  mimic fails here :/

PR #1168, opened 2023-12-19
Update link_content_credential_phishing.yml

morriscode:
Various legitimate docusign negations

### DIFF
--- a/detection-rules/link_content_credential_phishing.yml
+++ b/detection-rules/link_content_credential_phishing.yml
@@ -39,7 +39,20 @@ source: |
       and not profile.by_sender().any_false_positives
     )
   )
-
+  
+  // negate docusign 'via' messages
+  and not (
+      any(headers.hops,
+          any(.fields, .name == "X-Api-Host" and strings.ends_with(.value, "docusign.net"))
+      )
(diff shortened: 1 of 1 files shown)


detection-rules/link_content_credential_phishing.yml
```diff
@@ -39,7 +39,20 @@ source: |
...
   )
-
+  
+  // negate docusign 'via' messages
+  and not (
+      any(headers.hops,
+          any(.fields, .name == "X-Api-Host" and strings.ends_with(.value, "docusign.net"))
+      )
+      and strings.contains(sender.display_name, "via")
+    )
```
jkamdjou, line 49: i think indentation is wrong here

jkamdjou (approved):
🙌 do we want this on the other docusign rules too?

detection-rules/link_content_credential_phishing.yml
morriscode: Does it look better now? This is how the formatter formatted it :)

morriscode:
> 🙌 do we want this on the other docusign rules too?

The via component is in a couple other docusign rules already. I'll see if it's needed to add the received server raw component.

PR #1169, opened 2023-12-19
Create callback_phishing_google_group.yml

aidenmitchell:


detection-rules/callback_phishing_google_group.yml
```diff
@@ -0,0 +1,60 @@
...
+        and any(file.explode(.),
+  
+                // exclude images taken with mobile cameras and screenshots from android
+                not any(.scan.exiftool.fields,
+                        .key == "Model"
+                        or (
+                          .key == "Software"
+                          and strings.starts_with(.value, "Android")
+                        )
+                )
```
morriscode, line 21:
```suggestion
              // exclude images taken with mobile cameras and screenshots from Apple
              and not any(.scan.exiftool.fields,
                          .key == "DeviceManufacturer"
                          and .value == "Apple Computer Inc."
              )
```

morriscode: Figured we probably wanted to exclude IOS too

morriscode (approved):

PR #1176, opened 2023-12-20
FP Fix attachment_html_smuggling_atob

morriscode:
Curious what mimic results look like. Alternative FP fix to https://github.com/sublime-security/sublime-rules/pull/1125


morriscode:
this performs worse closing

PR #1178, opened 2023-12-20
Updating Knowbe4 insight & signal

morriscode:


insights/headers/phishing_simulation_knowbe4.yml
```diff
@@ -1,7 +1,24 @@
 name: "KnowBe4 phishing simulation"
 type: "query"
 source: |
-  any(headers.ips, regex.match(.ip, '147\.10\.167\.([0-9]|[0-5][0-9]|6[0-3])', '23.21.109.(197|212)', '52.49.(201.246|235.189)'))
+  type.inbound
```
aidenmitchell, line 4: Do we know if KB4 phish injections (not traditional mailed ones) come up as `inbound` in Platform?

morriscode: they seem to in FanDuels env, this logic was copied directly from their exception

aidenmitchell: Confirmed as inbound in customer env. Approving.

aidenmitchell (approved):

PR #1182, opened 2023-12-21
Update impersonation_okta.yml

aidenmitchell:
Negating replies.

### DIFF
--- a/detection-rules/impersonation_okta.yml
+++ b/detection-rules/impersonation_okta.yml
@@ -1,5 +1,5 @@
 name: "Brand impersonation: Okta"
-description: "Impersonation of Okta an Identity and access management company."
+description: "Impersonation of Okta, an identity and access management company."
 type: "rule"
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_okta.yml
```diff
@@ -9,7 +9,11 @@ source: |
     or strings.ilike(sender.email.domain.domain, '*Okta*')
     or strings.ilike(subject.subject, '*Okta*')
   )
-  and not(
+  and (
+    length(headers.references) > 0
+    or any(headers.hops, any(.fields, strings.ilike(.name, "In-Reply-To")))
+  )
```
jkamdjou, line 15: isn't this requiring it to be a reply?

aidenmitchell: oops, good catch!

morriscode (approved):

PR #1184, opened 2023-12-21
Signals: Org-specific signals

bfilar:


signals/content/subject_contains_org_display_name.yml
```diff
@@ -0,0 +1,4 @@
+name: "Subject: Matches entry in the $org_display_names list"
+type: "query"
+source: |  
+  subject.subject in~ $org_display_names
```
morriscode, line 4:
```suggestion
  any($org_display_names, strings.icontains(subject.subject, . ))
```
The existing logic would have to be an exact char match (case insens). I'm guessing by the name that's what you want, but I'd wager a contains would be more helpful?

signals/headers/headers_large_num_unknown_recipients.yml
```diff
@@ -0,0 +1,11 @@
+name: "Headers: Large unknown recipient list"
+type: "query"
+source: |  
+  (
```
morriscode, line 4:
```suggestion
```
Don't think these are needed

signals/headers/headers_large_num_unknown_recipients.yml
```diff
@@ -0,0 +1,11 @@
...
+    length(recipients.to) > 10
+    and length(filter(recipients.to,
+                      .email.domain.domain not in $org_domains
+                      and .email.email not in $recipient_emails
+               )
+    ) >= 10
+  )
```
morriscode, line 11:
```suggestion
```

morriscode (approved):

signals/content/subject_contains_org_display_name.yml
```diff
@@ -0,0 +1,4 @@
+name: "Subject: Matches entry in the $org_display_names list"
+type: "query"
+source: |  
+  any($org_display_names, strings.icontains(subject.subject, . ))
```
rw-access, line 4: This list is probably too large for this to be performant for `any` checks like this

rw-access: I would say stick with `in`, `in~` with large lists for now

bfilar: Awesome! Yeah, I wasn't sure the best way to do that

PR #1185, opened 2023-12-21
Signals: Common legitimate mail services

bfilar:


signals/sender/sender_legitimate_domain_contain_recipient_addr.yml
```diff
@@ -0,0 +1,4 @@
+name: "Sender: Legitimate domain where recipient in sender's email address local part"
```
aidenmitchell, line 1: Should we be checking for the "recipient in sender's email address local part"?

signals/sender/sender_legitimate_domain_negation.yml
```diff
@@ -0,0 +1,27 @@
+name: "Sender: Legitimate message senders"
```
aidenmitchell, line 1: What's the idea here, to fire on secure mailers? Right now, it looks like this is just "not a known secure mailer".

aidenmitchell (changes requested):

signals/sender/sender_legitimate_domain_negation.yml
bfilar: yeah. i need to `or` these and remove the `not` logic

aidenmitchell, line 1:
```suggestion
name: "Sender: Legitimate secure message senders"
```

signals/sender/sender_legitimate_domain_negation.yml
```diff
@@ -0,0 +1,27 @@
...
+type: "query"
+source: |  
+  (
+    sender.email.domain.root_domain in ("protectedtrust.com", "medallia.com", "icims.com", "workday.com")
+    or any(body.links,
+            .href_url.domain.root_domain == sender.email.domain.root_domain
+    )
```
aidenmitchell, line 8:
```suggestion
```

I'm not sure these senders _always_ send secure messages... thoughts?
as for L6-8, it's probably a remnant from pulling these signals out

signals/sender/sender_legitimate_ups_domain.yml
```diff
@@ -0,0 +1,4 @@
+name: "Sender: Legitimate UPS domain"
+type: "query"
+source: |  
+  sender.email.domain.root_domain in ("ups.com", "upsemail.com")
```
aidenmitchell, line 4: May want to add a DMARC check here, unless risk score will weigh a DMARC failure more/a failure would be accounted for elsewhere.

signals/sender/sender_is_known_msft_domain copy.yml
```diff
@@ -0,0 +1,14 @@
+name: "Sender: Known MSFT domain"
+type: "query"
+source: |
```
aidenmitchell, line 3: May want to add a DMARC check here, unless risk score will weigh a DMARC failure more/a failure would be accounted for elsewhere.

signals/sender/sender_is_known_boa_domain.yml
```diff
@@ -0,0 +1,4 @@
+name: "Sender: Known Bank of America domain"
+type: "query"
+source: |
```
aidenmitchell, line 3: May want to add a DMARC check here, unless risk score will weigh a DMARC failure more/a failure would be accounted for elsewhere.

signals/sender/sender_is_known_fax_service.yml
```diff
@@ -0,0 +1,4 @@
+name: "Sender: Known fax mailers"
+type: "query"
+source: |
```
aidenmitchell, line 3: May want to add a DMARC check here, unless risk score will weigh a DMARC failure more/a failure would be accounted for elsewhere.

aidenmitchell (changes requested):

signals/sender/sender_legitimate_ups_domain.yml
bfilar: Yeah we have signals already for DMARC failures

aidenmitchell (approved):

PR #1189, opened 2023-12-21
Update image_as_content_open_redir.yml

aidenmitchell:


aidenmitchell:
The 1 regression was a lucky catch by this rule, it's not image as content. Still catches on a DocuSign rule, so not concerned. Override will be needed.

morriscode:
cc @aidenmitchell  there's a few more regressions here if you want to review and accept them
176038,188966, 146374, 101127

aidenmitchell:
/mql-mimic-exempt 176038, 188966, 146374, 101127
176038, 188966, 146374 are not image as content.
101127 is not an open redirect.

morriscode (approved):

PR #1191, opened 2023-12-22
Update link_login_or_captcha.yml

morriscode:
Huge FP mitigation by not analyzing the free subdomain links if they're all images

### DIFF
--- a/detection-rules/link_login_or_captcha.yml
+++ b/detection-rules/link_login_or_captcha.yml
@@ -6,26 +6,37 @@ severity: "medium"
 source: |
   type.inbound
   and any(body.links,
-
+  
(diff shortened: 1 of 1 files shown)


morriscode:
This is an ASR Rule. The regressions here are covered by higher fidelity higher accuracy rules. The regressions on this rule in particular were a lucky catch, as the intent of the rule is not to detect free subdomain hosted images, but content. 

Other rules firing are: 
Brand impersonation: American Express (AMEX)
Credential phishing content and link (first-time sender)
Credential phishing link (unknown sender)


This needs a mimic override

jkamdjou (approved):

PR #1195, opened 2023-12-26
Trigger via Comment + Always Post "Rule Tests and ID Updated" Check

cameron-dunn-sublime:
First step for adding the ability to comment to accept certain MQL Mimic failures. I have to basically develop on `main` because workflows triggered by comments only run for actions defined on `main`. If this gets too painful I can test on a scratch repo though.

Also this change moves the `Rule Tests and ID Updated` check up to earlier & it'll always post a status. This just means that PRs that fail early non-mimic validations will show this check with a red X instead of being pending.

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -6,6 +6,8 @@ on:
   pull_request_target:
     branches: [ "**" ]
   workflow_dispatch: {}
+  issue_comment:
+    types: [ created ]
 
 concurrency:
   # For pull_request_target workflows we want to use head_ref -- the branch triggering the workflow. Otherwise,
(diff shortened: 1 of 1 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -96,8 +114,45 @@ jobs:
...
+      # When we add a commit, GitHub won't trigger actions on the auto commit, so we're missing a required check on the
+      # HEAD commit.
+      # Various alternatives were explored, but all run into issues when dealing with forks. This sets a "Check" for
+      # the latest commit, and we can depend on that as a required check.
+      - name: "Create a check run"
+        uses: actions/github-script@v6
+        if: github.event_name == 'pull_request_target' && ${{ always() }}
+        env:
+          parameter_url: '${{ github.event.pull_request.html_url }}'
+          conclusion: "${{ steps.final_basic_validation.outcome == 'success' && 'success' || 'failure' }}"
```
cameron-dunn-sublime, line 129: Added this logic + usage on L147 -- otherwise this block is just moved.

morriscode (approved):

PR #1197, opened 2023-12-26
Expand Workflow Permissions

cameron-dunn-sublime:
Follow up from https://github.com/sublime-security/sublime-rules/pull/1195


cameron-dunn-sublime:
Some other comment

PR #1198, opened 2023-12-27
Two Follow Ups: Expand Permissions + Hopefully fix check logic

cameron-dunn-sublime:
Two follow ups from https://github.com/sublime-security/sublime-rules/pull/1195

1. Expand workflow permissions (needed to read comments)
2. CI failed on `main` because we tried to post the "Rule Tests and ID Updated" check (we shouldn't)

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -2,7 +2,7 @@ name: Rules PR CI
 
 on:
   push:
-    branches: [ "main", "test-rules", "cd.test-comment" ]
+    branches: [ "main", "test-rules" ]
   pull_request_target:
     branches: [ "**" ]
   workflow_dispatch: {}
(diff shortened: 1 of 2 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -123,7 +125,7 @@ jobs:
       # the latest commit, and we can depend on that as a required check.
       - name: "Create a check run"
         uses: actions/github-script@v6
-        if: github.event_name == 'pull_request_target' && ${{ always() }}
+        if: github.event_name == 'pull_request_target' && always()
```
cameron-dunn-sublime, line 128: I tested this by expanding L5 to

```
on:
  push:
    branches: [ "main", "test-rules", "cd.test-comment ]
```

and it skipped this step like it should have (I didn't have a PR open on the branch): https://github.com/sublime-security/sublime-rules/actions/runs/7333566023/job/19969174991

BTW: `always()` in this context just means ignore previous step failures.

morriscode (approved):

PR #1199, opened 2023-12-27
Collect mql-mimic-exempt Comments & Sent to MQL Mimic

cameron-dunn-sublime:
With some luck, this will be the last change on the `sublime-rules` side to support a system where we can make exceptions to MQL Mimic by commenting:

```
/mql-mimic-exempt: 12234, 556
```

That would cause 12234 & 556 to be skipped in the test suite. This will allow bypassing when the team decides that a apparent regression is actually a good thing, e.g. the EML shouldn't have matched the rule in the first place.

The behavior is that:
1. Whenever a commit is pushed to a PR, or a comment with `/mql-mimic-exempt` is added by an org member, the test suite will be re-triggered and should update the existing checks.
    * Updating of the existing checks isn't tested yet, but I think it will work (I can't test till earlier issues are resolved on `main`)
2. All comments on the PR are inspected for `/mql-mimic-exempt`, and any numbers that appear on the same line are considered.
    * To be clear, this means that the comments accumulate rather than overwrite.
    * The syntax is pretty flexible, e.g. all of these are allowed:
        * `/mql-mimic-exempt     :123;456`
        * `/mql-mimic-exempt: 123 456`
        * `/mql-mimic-exempt: 123,#456`
        * `/mql-mimic-exempt 123/456`
        * etc.. `#` is ignored, `,`, `;`, `:`, `/` and any whitespace are allowed as separators

**Downstream support in MQL Mimic has not been added yet!** It's mostly done, but not merged.

Tested via https://github.com/sublime-security/sublime-rules/pull/1202, resulting trigger request body is `{"branch":"cd.test-2","repo":"sublime-security/sublime-rules","token":"***","sha":"de8c38262ed8cda7c55a087c9d0e802f7d462458","only_rule_ids":"2fc36c6d-86a2-5b12-b5a4-5d8744858381","skip_eml_ids":"123 456 789 123 456 789 444"}` which is correct for the comments I've put on that PR.

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -61,33 +61,33 @@ jobs:
           pip install -r scripts/generate-rule-ids/requirements.txt
           python scripts/generate-rule-ids/main.py
 
-#      - name: Validate Rules
-#        run: |
-#          for f in *-rules/*.yml
-#          do
-#            echo "Processing $f"
-#            http_code=$(yq -o=json eval 'del(.type)' "$f" | curl -H "Content-Type: application/json" -X POST --data-binary @- -o response.txt -w "%{http_code}" --silent https://playground.sublimesecurity.com/v1/rules/validate)
-#            echo '' >> response.txt
-#            cat response.txt
-#            if [[ "$http_code" != "200" ]]; then
-#              echo "Unexpected response $http_code"
-#              exit 1
-#            fi
-#          done
-#
-#      - name: Validate Insights and Signals
-#        run: |
-#          for f in {insights,signals}/**/*.yml
-#          do
-#            echo "Processing $f"
-#            http_code=$(yq eval 'del(.type) | .source = "length([\n\n" + .source + "\n]) >= 0"' "$f" -o=json | curl -H "Content-Type: application/json" -X POST --data-binary @- -o response.txt -w "%{http_code}" --silent https://playground.sublimesecurity.com/v1/rules/validate)
-#            echo '' >> response.txt
-#            cat response.txt
-#            if [[ "$http_code" != "200" ]]; then
-#              echo "Unexpected response $http_code"
-#              exit 1
-#            fi
-#          done
+      - name: Validate Rules
+        run: |
+          for f in *-rules/*.yml
+          do
+            echo "Processing $f"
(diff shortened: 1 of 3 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -39,8 +39,8 @@ jobs:
         uses: actions/checkout@v3
         if: github.event_name == 'issue_comment'
         with:
-          repository: ${{ steps.comment-branch.outputs.head_owner }}/${{ steps.comment-branch.outputs.head_repo }}
-          ref: ${{ steps.comment-branch.outputs.head_ref }}
+          repository: ${{ steps.comment_branch.outputs.head_owner }}/${{ steps.comment_branch.outputs.head_repo }}
```
cameron-dunn-sublime, line 42: Another follow up from https://github.com/sublime-security/sublime-rules/pull/1198/https://github.com/sublime-security/sublime-rules/pull/1195

 Still not seeing the comment triggering behavior work as expected, and hopefully this is the last issue. I can't test this until it's merged to `main`, so no guarantees (but we have a near identical workflow for test rules, so it should be close if this doesn't fix it).

detection-rules/attachment_adobe_image_lure_qr_code.yml
```diff
@@ -22,6 +22,7 @@ source: |
             or .file_type == "pdf"
             or .file_type in $file_extensions_macros
           )
+
```
cameron-dunn-sublime, line 25:
```suggestion
```

.github/workflows/update-test-rules.yml
```diff
@@ -45,12 +45,12 @@ jobs:
 
       - name: Get PR branch
         uses: alessbell/pull-request-comment-branch@v1.1 # Fork of xt0rted/pull-request-comment-branch, see https://github.com/xt0rted/pull-request-comment-branch/issues/322
-        id: comment-branch
+        id: comment_branch
```
cameron-dunn-sublime, line 48: Changed it here for consistency. `_` preferred so that it's considered a single word in most contexts.

morriscode:
This looks great! I'm excited to hear how the testing goes.

morriscode (approved):

PR #1201, opened 2023-12-27
New Rule: Spam Emoji Filter (Unsolicited and Untrusted senders)

morriscode:


morriscode:
This looks great in manual hunts, will add test rules when CI finishes

detection-rules/spam_emoji_filter_untrusted_unsolicited.yml
```diff
@@ -0,0 +1,35 @@
+name: "Spam: Emoji Filter (Untrusted and Unsolicited Senders)"
```
jkamdjou, line 1:
```suggestion
name: "Spam emojis in subject or display name (unknown sender)"
```

does this look for any emoji, or is it particular ones that are commonly associated with spam?

morriscode: Smileys & Emotion (😀 - 😏)
People & Body (👋 - 🤏)
Animals & Nature (🐀 - 🌿)
Food & Drink (🍇 - 🍾)
Travel & Places (🚀 - 🛸)
Activities (🎃 - 🎿)
Objects (📔 - 🧾)
Symbols (💀 - 🧿)
Flags (🏁 - 🏳️‍🌈)

It covers a massive range. But their presence in a subject or display name is typically indicative of unwanted email. I'm hoping to learn more from test-rule results

jkamdjou: word. i get legit newsletters with emojis in the subject, that i never reply to from time to time. is the # of emojis a factor too?

morriscode: Same. I don't want to cater to the exception though, it prevents more spam than newsletters and occasionally order receipts. We could deploy it as a discovery rule or ASR? I think there's an appetite for a more aggressive approach for spam, but it may not align with everyone's posture.  

 I've been running it for months, and I've permitted two sender exclusions. The rest is garbage in the thousands.

morriscode: we're at 550000 matches but only like 2000 actions. Regardless of how well i think this works, I'm going to probably close this

PR #1204, opened 2023-12-27
Update attachment_adobe_image_lure.yml

morriscode:
FN coverage. Adding message screenshot


morriscode:
Testing looks great, this is gtg.

detection-rules/attachment_adobe_image_lure.yml
```diff
@@ -9,25 +9,26 @@ source: |
...
-                  )
-                  and (
-                    (length(body.links) > 0)
-                    or (length(.scan.url.urls) > 0 or length(.scan.pdf.urls) > 0)
-                  )
-          )
+  and (
+    any(file.explode(beta.message_screenshot()),
+        any(ml.logo_detect(beta.message_screenshot()).brands, .name == "Adobe")
+        and 0 < length(body.links) < 10
```
jkamdjou, line 15: this feels a little broad -- basically flagging on a message with an Adobe logo + a link? am i understanding correctly?

morriscode: and there's no display text on the links.
It seems broad, but it's fired a collective 180 times in testing, and there's only 4 hits with no other rule overlap. 

Of those 180 firings, 73 of them have actions confirming TP, (quarantine, trash, webhook) with 0 actions indicating FP's.

aidenmitchell (approved):

aidenmitchell (approved):

PR #1206, opened 2023-12-27
MQL Mimic Exemption Follow Ups

cameron-dunn-sublime:
1. Fix an issue with the wrong branch being used for testing
2. Fix an issue preventing MQL Mimic tests from actually running

Described more in line. These issues only affected comment triggers (what I added earlier today).


.github/workflows/rule-validate.yml
```diff
@@ -314,12 +340,12 @@ jobs:
             -d "$body"
 
       - name: Wait for MQL Mimic check to be completed
-        uses: fountainhead/action-wait-for-check@v1.1.0
-        id: wait-for-build
+        uses: sublime-security/action-wait-for-check@master
```
cameron-dunn-sublime, line 343: I forked this to add support for `ignoreIDs`. This is pretty simple so I'm hopeful it'll work.

The issue I saw was that we triggered the new MQL Mimic run, but immediately saw the previous result and finished the workflow (it needs to stay running for MQL Mimic to post back results).

Ignoring the existing check ID should avoid this issue. I validated that new checks show up with a new ID, even when the same name is posted to the same ref/commit.

I'd probably do this differently if I was starting from scratch -- have the pending check be created here (vs. from MQL mimic) and pass along the check ID to be updated, but that would require juggling changes in a couple spots at once.

.github/workflows/rule-validate.yml
```diff
@@ -295,12 +303,30 @@ jobs:
...
+        id: find_mql_mimic_results
+        env:
+          sha: '${{ steps.get_head.outputs.HEAD }}'
+        with:
+          debug: ${{ secrets.ACTIONS_STEP_DEBUG || false }}
+          script: |
+            const result = await github.rest.checks.listForRef({
+              check_name: "MQL Mimic Tests",
+              owner: "sublime-security",
+              repo: "sublime-rules",
```
cameron-dunn-sublime, line 317: This is correct -- the checks are always on `sublime-security/sublime-rules`. Sending along the repo from forks just supports checking out the rule code.

.github/workflows/rule-validate.yml
```diff
@@ -35,21 +35,43 @@ jobs:
...
 
-      - name: Checkout (from comment)
-        uses: actions/checkout@v3
-        if: github.event_name == 'issue_comment'
-        with:
-          repository: ${{ steps.comment_branch.outputs.head_owner }}/${{ steps.comment_branch.outputs.head_repo }}
-          ref: ${{ steps.comment_branch.outputs.head_ref }}
-          fetch-depth: 0
+      - name: Get Refs
+        id: get_refs
```
cameron-dunn-sublime, line 39: I refactored this to handle some more logic. When a comment was triggering, we were running MQL Mimic with a ref of `main`, not the PR branch (see line 302 of the old code)

PR #1208, opened 2023-12-28
Another Workflow Fix

cameron-dunn-sublime:
Close! Commenting is working as expected, but CI that runs on pushes to `main` was broken, and I had one other incorrect reference.

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -36,18 +36,42 @@ jobs:
         id: comment_branch
 
       - name: Get Refs
-        id: get_refs
+        id: get_head_ref
         run: |
+          # Accurate for push events, other events require changes.
           head_ref="${{ github.head_ref }}"
           repo="${{ github.repository }}"
-          # Either of these must be set in block below
+
+          if [[ "${{ github.event_name }}" == 'pull_request_target' ]]; then
+            repo="${{ github.event.pull_request.head.repo.full_name }}"
+          elif [[ "${{ github.event_name }}" == 'workflow_dispatch' ]]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -36,43 +36,56 @@ jobs:
...
-          run_all=""
-          base_ref=""
 
           if [[ "${{ github.event_name }}" == 'pull_request_target' ]]; then
-            # Detect changes based on whatever we're merging into.
-            base_ref="${{ github.base_ref }}"
             repo="${{ github.event.pull_request.head.repo.full_name }}"
-          elif [[ "${{ github.event_name }}" == 'push' ]]; then
-            # Detect changes based on the previous commit
-            base_ref="$(git rev-parse HEAD^)"
```
cameron-dunn-sublime, line 53: This can't be done until after the Checkout step, so I had to split this into two steps.

.github/workflows/rule-validate.yml
```diff
@@ -188,15 +201,15 @@ jobs:
 
       - name: Checkout base
         uses: actions/checkout@v4
-        if: ${{ steps.get_refs.outputs.run_all != 'true' }}
+        if: ${{ steps.get_base_ref.outputs.run_all != 'true' }}
         with:
-          ref: ${{ steps.get_refs.outputs.head_ref }}
+          ref: ${{ steps.get_base_ref.outputs.base_ref }}
```
cameron-dunn-sublime, line 206: This was a mistake from before -- it should be base_ref

morriscode (approved):

PR #1218, opened 2024-01-04

morriscode:
Making this more tolerant of fake password expiration notices by changing the NLU requirements to be OR'd conditions. This allows the suspicious text to stand on it's own. Seeing how it performs in mimic


morriscode:
Performance looks great. We have a lot of overlap with customer rules, and other feed rules plus a 5% gain in net new coverage.

aidenmitchell (approved):

PR #1220, opened 2024-01-04
Update body_callback_phishing_no_attachment.yml

morriscode:
Loosening rule to accommodate non-freemail senders.  Adding message screenshot


morriscode:
Volume looks acceptable, i see no actions indicative of FP's. Large overlap with custom rules.

morriscode:
archiving this, rule had updates today that should help

PR #1222, opened 2024-01-05
Update attachment_html_smuggling_atob.yml

morriscode:
Removing a duplication of the strings check, and swapping the remaining strings check to a regex with distance.


morriscode:
/mql-mimic-exempt: 26407, 25914, 25177, 15723, 224, 62823

morriscode:
The above exclusions were matching on incorrect atob overlap. Tightening the distance prevents such matches by closer anchoring the match to the surrounding words. (FP's such as 'AtObject' for example)

jkamdjou (approved):
nice

PR #1227, opened 2024-01-09
New Rule: DoorDash Impersonation

morriscode:


morriscode:
This is pretty safe to active, no hits in testing yet, but matches on the known samples shared.

aidenmitchell (approved):

PR #1231, opened 2024-01-09
New Rule: Dropbox comment phishing

morriscode:


morriscode:
Testing results look good, no egregious match quantity, rule is pretty specific, low concerns with merge.

detection-rules/dropbox_credential_phishing_via_comment.yml
```diff
@@ -0,0 +1,78 @@
...
+  and regex.contains(body.current_thread.text,
+                     "[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}"
+  )
+  
+  // and it's likely a freemail
+  and any($free_email_providers, strings.icontains(body.current_thread.text, .))
+  
+  // unsolicited and no FP's
+  and not profile.by_sender().solicited
+  and not profile.by_sender().any_false_positives
```
jkamdjou, line 65:
```suggestion
```
we're filtering only to dropbox as a sender, so i don't think we want this here

detection-rules/dropbox_credential_phishing_via_comment.yml
```diff
@@ -0,0 +1,75 @@
...
+  
+  // there's an email in the body
+  and regex.contains(body.current_thread.text,
+                     "[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,}"
+  )
+  
+  // and it's likely a freemail
+  and any($free_email_providers, strings.icontains(body.current_thread.text, .))
+  
+
```
jkamdjou, line 63:
```suggestion
```

jkamdjou (approved):

PR #1236, opened 2024-01-10
New Rule: Fake Storage Alert

morriscode:


morriscode:
Testing looks good here!

aidenmitchell (changes requested):
Are we going to account for legit "Approaching SharePoint Site Storage Limit" emails? I see those surfacing in some envs.

morriscode:
> Are we going to account for legit "Approaching SharePoint Site Storage Limit" emails? I see those surfacing in some envs.

Yeah I should probably do that. I saw a couple in but couldn't tell if they were legit or not

aidenmitchell (approved):

morriscode:
Testing looks good, distribution is pretty even, some overlap with brand impersonation rules, this is net new coverage primarily.

PR #1247, opened 2024-01-19
Update link_microsoft_impersonation_using_hosted_png.yml

myenerguler:
Hello Team,

First of all, I would like to thank you for this project, which I have admired.

In the part of the rule that I marked with an asterisk, is the goal to detect a logo image embedded in the HTML mail content with an img tag? If so, I guess the URL referenced in the src attribute of the img tag is not included in the 'link' array. The best solution at this point would be to pass the output of the function that takes a screenshot of the message to the ml.logo_detection function. I've noticed in some other rules that only attachment images are subject to some detection; I think it might be a good practice to include images embedded in the message with the img tag in the HTML content in the detection rule, by subjecting the message screenshot to the same procedure. Actually, more generally, it would be great to design a function that outputs the raw HTML value parse with beautifulsoup.

### DIFF
--- a/detection-rules/link_microsoft_impersonation_using_hosted_png.yml
+++ b/detection-rules/link_microsoft_impersonation_using_hosted_png.yml
@@ -14,7 +14,7 @@ source: |
   and sender.email.domain.root_domain not in~ ('microsoft.com', 'microsoftsupport.com', 'office.com')
 
   // logo hosted on microsoft.com
-  and any(body.links,
+  **and any(body.links,
           regex.icontains(.display_url.url, '.{0,50}microsoft\.com\/.{0,70}logo.{0,25}\.png')
   )
 


morriscode:
Thanks @myenerguler 

This rule is specifically crafted to look for a very particular set of events.

1. Type: Inbound - The rule applies to incoming messages.
2. Display Name in All Caps - The sender's display name is entirely in uppercase letters, which is a common tactic in phishing emails to grab attention.
3. Sender's Email Domain - The sender's email domain is not one of the official Microsoft domains (`microsoft.com`, `microsoftsupport.com`, `office.com`).
4. Microsoft Logo URL - The email contains a link to a URL that seems to host a Microsoft logo (`logo.png`), and the URL includes `microsoft.com`. This is used to make the email appear legitimate.
5. Link Display Text in All Caps - The email contains at least one link with display text in all caps, such as "VIEW", "CLICK", "DOWNLOAD", "CHECK". This is another method to grab the recipient's attention.
6. Request Being Made - The email contains a request, which could be an action the sender wants the recipient to take. This is identified using natural language understanding techniques.
7. Sender Domain and Link Domain Mismatch - None of the domains in the links match the sender's email domain, and they are not part of the top million domains, indicating they might be less reputable or unfamiliar.
8. Organization Domain in Email Subject - The subject of the email contains a domain related to the recipient's organization, suggesting the email is trying to appear relevant to the recipient's work.

We have other rules that use various approaches to detect Microsoft logos and HTML smuggling. I will be closing this PR but however I'd encourage you to continue this discussion in our[ community slack channel. ](https://sublimecommunity.slack.com/join/shared_invite/zt-1hhwosroy-LvflKNVE3HEtgIcbHdB1sw#/shared-invite/email)

PR #1248, opened 2024-01-19
New Rule: Risk Score

bfilar:


detection-rules/risk_score.yml
```diff
@@ -0,0 +1,18 @@
...
+description: |
+  Aggregates a series of MQL-based signals to determine risk of a message
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and beta.risk_score().confidence in ("medium", "high")
+  and (
+    (
+      profile.by_sender().prevalence in ("new", "outlier")
```
jkamdjou, line 11: can we drop this? we'll likely run into the FTS issues from the past

bfilar: @jkamdjou does this seem better to you?

morriscode:
The above test rules add never took. The rule is not in the table. Going to try again will monitor actions.

PR #1253, opened 2024-01-22

morriscode:
Adding image size restrictions


morriscode:
/mql-mimic-exempt: 112340,111602,114206,114016,114015,111578,113081,113550,115144

These fire other tighter scoped QR code rules

PR #1256, opened 2024-01-24
Update impersonation_microsoft_quarantine.yml

aidenmitchell:
Adding detection for body (not an attachment).


morriscode:
@aidenmitchell conflict here too, if you can take a peak.

morriscode:
This looks like it needs tuning, I'm guessing loosening the scope without the corresponding negations like we had in place for MS is causing some issues. See slack for deets

PR #1261, opened 2024-01-28
New Rule: Attachment: Docx with embedded binary

morriscode:


morriscode:
The logic here is really specific, matches known bad samples, but no hits in testing or live hunts. Should be gtg.

detection-rules/attachment_docx_embedded_binary.yml
```diff
@@ -0,0 +1,61 @@
...
+                  // The malicious file to be downloaded and run with the data URI may not always be portable executable
+                  or any(.scan.strings.strings,
+                         strings.ilike(., "*.createObjectURL(*)*")
+                  )
+                  or any(.scan.strings.strings,
+                         strings.ilike(., "*.msSaveOrOpenBlob(*)*")
+                  )
+          )
+  )
+
```
jkamdjou, line 52:
```suggestion
tags:
  - "Attack surface reduction"
```

jkamdjou (approved):
one suggestion

PR #1262, opened 2024-01-29
New Rule: Impersonation Suspected supplier lookalike with suspicious content

morriscode:


morriscode:
This matches 2 known samples of supplier impersonation attempting to intercept or redirect payments. 
It has not fired at all in the last day of testing. It's a relatively over tightened rule and because of that it should be fairly safe to deploy. I don't expect high volume but I do expect when it fires, it will be worth having. 

I'm also working on other variation using this logic that are perhaps slightly looser. lmk what you want to do with this one.

aidenmitchell (approved):

detection-rules/impersonation_suspected_supplier_payment_request.yml
```diff
@@ -0,0 +1,72 @@
+name: "Impersonation: Suspected supplier impersonation with suspicious content"
+description: "This rule detects supplier impersonation by checking for: similar but not identical linked domains to the sender, non-freemail senders using freemail infrastructure, sender domains less than 90 days old (via whois), unsolicited communication or no prior interaction with the reply-to address, and either payment interception language or a very short email body with only image attachments."
```
aidenmitchell, line 2:
```suggestion
description: "This rule detects supplier impersonation by checking for: similar linked domains to the sender, non-freemail senders using freemail infrastructure, sender domains less than 90 days old, unsolicited communication or no prior interaction with the reply-to address, and a suspicious body."
```

aidenmitchell (changes requested):

aidenmitchell (approved):

PR #1266, opened 2024-01-30
New Signals: Impersonation set

bfilar:


signals/content/impersonation_msft.yml
```diff
@@ -0,0 +1,36 @@
+name: "Impersonation: Microsfot"
```
morriscode, line 1:
```suggestion
name: "Impersonation: Microsoft"
```

bfilar: I was showing the power of levenshtein distance :trollface:

PR #1273, opened 2024-01-31
Update credential_phishing_suspicious_subject_nlu_financial_urgent.yml

aidenmitchell:
Changing negation to case-insensitive.


zoomequipd:
Looks like this can be closed - change was contained within 
https://github.com/sublime-security/sublime-rules/commit/5354b1309932196afcb3db463c919a21d61e112f

PR #1278, opened 2024-02-01
New Signals

bfilar:


morriscode (approved):

signals/headers/headers_no_recipients_in_org_domain.yml
```diff
@@ -1,4 +1,5 @@
-name: "Headers: No recpients in $org_domains list"
```
rw-access, line 1: FYI this needs to be changed in both Mantis and Hydra (we were matching the typo before)

signals/sender/sender_new_domain_L90D.yml
```diff
@@ -0,0 +1,4 @@
+name: "Sender: Domain registered less than 90 days ago"
+type: "query"
+source: |
+  beta.whois(sender.email.domain).days_old < 90
```
rw-access, line 4: Why not just make the signal on `.days_old` instead?

signals/headers/headers_no_recipients_in_org_domain.yml
bfilar: yep... I'll pick it up during model training, and then PR an updated config in Hydra

signals/sender/sender_new_domain_L90D.yml
bfilar: yes... that would work too :|

PR #1279, opened 2024-02-05
New Rule: Credential phishing: Suspicious link traversing Russian infrastructure

morriscode:
Feel free to tweak the name, I was drawing a blank. 
We saw about 2000 of these earlier with the only rule firing being "Russia return-path TLD (untrusted sender)" thought they could use something of higher fidelity.

### DIFF
--- a/detection-rules/link_russian_traversed_credential_phishing.yml
+++ b/detection-rules/link_russian_traversed_credential_phishing.yml
@@ -50,3 +50,4 @@ detection_methods:
   - "Natural Language Understanding"
   - "URL analysis"
 
+id: "a5203e3b-979f-5feb-bc6c-d62de08a2823"
--- a/detection-rules/link_russian_traversed_credential_phishing.yml
+++ b/detection-rules/link_russian_traversed_credential_phishing.yml
@@ -1,4 +1,4 @@
-name: "Link: Credential Phishing via messages with Russian TLD traversal"
+name: "Link: Credential Phishing traversing Russian infrastructure"
 description: "This rule detects credential phishing attempts in emails traversing Russian TLDs by aggressively analyzing links for signs of phishing, including suspicious keywords, login prompts, or links flagged for credential theft, excluding emails from trusted domains unless they fail DMARC verification."
 type: "rule"
 severity: "high"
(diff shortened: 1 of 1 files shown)


detection-rules/link_russian_traversed_credential_phishing.yml
```diff
@@ -0,0 +1,53 @@
+name: "Link: Credential Phishing via messages with Russian TLD traversal"
```
aidenmitchell, line 1:
```suggestion
name: "Link: Credential Phishing via Russian sending infrastructure"
```

detection-rules/link_russian_traversed_credential_phishing.yml
```diff
@@ -0,0 +1,53 @@
...
+  and any(headers.domains, .tld in ("ru", "su"))
+  and any(ml.nlu_classifier(body.current_thread.text).intents,
+          .name == "cred_theft" and .confidence in ("medium", "high")
+  )
+  and 0 < length(body.links) < 5
+  and any(body.links,
+          beta.linkanalysis(., mode="aggressive").credphish.disposition == "phishing"
+          or (
+            strings.icontains(beta.linkanalysis(., mode="aggressive").final_dom.raw,
+                              "Pàsswórd"
```
aidenmitchell, line 16: We might want more permutations here (perhaps with `replace_confusables`?)

aidenmitchell (changes requested):

detection-rules/link_russian_traversed_credential_phishing.yml
morriscode: I thought about that too, sending infrastructure seems like we're saying origination when this is really just any route traversal.

detection-rules/link_russian_traversed_credential_phishing.yml
morriscode: For sure since it wasn't accessible globally yet I didn't include it

aidenmitchell: Oh that's not GA... hm.

aidenmitchell (approved):

PR #1283, opened 2024-02-06
New Rule: Detect PDF attachments with links created with Python 

padey:
The goal of this detection rule is to ensure that PDF files generated with a Python and with one or more links in the PDF are detected. These techniques were used by PikaBot (they used ReportLab), among others.


morriscode:
thanks @padey. Will initiate some testing on this right now

padey:
@morriscode thanks! My testpool is pretty limited / small. I'm excited what you think!

morriscode:
Testing looks great here, I don't see any FP's. Going to approve

PR #1288, opened 2024-02-06

morriscode:
Adding gating for newsletters with an exorbitant amount of links


jkamdjou (commented):
we should add high trust here too?

morriscode:
> we should add high trust here too?

I thought that at first but pivoted to this because I was afraid it may cause failures from other cloud hosted senders we do trust. 

I'll try adding high trust and seeing what mimic says.

PR #1289, opened 2024-02-06
Create mismatched_sender_reply_return_domains.yml

aidenmitchell:


insights/headers/mismatched_sender_reply_return_domains.yml
```diff
@@ -0,0 +1,10 @@
+name: "Sender, reply-to, and return path domains are distinct"
```
jkamdjou, line 1:
```suggestion
name: "Mismatched sender, reply-to, and return path"
```
this feels more consistent with other insight names

morriscode (approved):

PR #1290, opened 2024-02-06
New Rule: Link: QR code in EML attachment with credential phishing indicators

morriscode:


morriscode:
FN coverage, rule uses LA fyi.

jkamdjou (approved):

PR #1293, opened 2024-02-07
Updating rules to use attack_types: Reconnaissance

morriscode:


morriscode:
pending FE and BE adoption of the recon attack_type.

PR #1309, opened 2024-02-12
Update credential_phishing_suspicious_subject_nlu_financial_urgent.yml

morriscode:
FP reduction. Dropping casing requirements for "REGISTER NOW"


morriscode:
Mimic failure (269504) is a benign sample that we're trying to avoid with this change.

jkamdjou (approved):

PR #1313, opened 2024-02-12
Update impersonation_wells_fargo.yml

morriscode:
I'm not a huge fan of the entire first block in this rule, but adding regex for wellsfargo for now. Need to rework this entire rule.


morriscode:
This is good to go

bfilar (approved):

PR #1320, opened 2024-02-15
New Rule: Callback Phishing via extensionless rfc822 attachment

morriscode:


morriscode:
This is pretty specific and gated akin to the other callback phishing messages. I would bet this is safe to ship.

bfilar (approved):

detection-rules/callback_phishing_extensionless_rfc822_image.yml
```diff
@@ -0,0 +1,80 @@
+name: "Callback Phishing via extensionless rfc822 attachment "
+description: "This rule detects messages with an unknown file_type (extensionless) and a content_type of 'message/rfc822' that do not contain an email, but rather an image file with Callback Phishing indicators."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(attachments,
+          .file_type == "unknown"
+          and .content_type == "message/rfc822"
+          and not any(file.explode(.),
```
jkamdjou, line 10: is this `not` supposed to be here?

morriscode: yep

morriscode: I can move it to the line below if that makes it more clear
doesn't change the functionality

morriscode: that should be better?

PR #1324, opened 2024-02-19
New Rule: PikaBot Malicious Office Document 

padey:
PikaBot changed its attack pattern from links/urls in the mail body/pdf to classic malicious office documents. This rule is quite simple, but works in our environment without any false positives. 

Example string:
`file://///85[.]195[.]115[.]20/share/reports_02[.]15[.]2024_2[.]js`


morriscode:
thanks @padey, will start testing today!

PR #1326, opened 2024-02-21
New Rules: Attachment and Link PikaBot Februar 2024

padey:
@morriscode - wanted to create a second PR for another rule, deleted my commits and closed the latest PR, sorry :D that was not my plan.

The link rule covers the smb link as a link in the body which bypasses some protections from windows/office. Opens the zip directly, exe is inside. This is my first rule which uses linkanalysis, not sure if everything is correct...

The attachment rule covers the office docs with smb links to a javascript file.


morriscode:
Put some notes, and also DM'd you some issues with using LA against SMB shares.

padey (commented):
@morriscode maybe its to better idea to go for a a simpler detection? :) If I understood you correct, linkanalysis can't handle smb and so we can "file.explode" the exe?

PR #1331, opened 2024-02-22
Add mime flavor check for resiliency

jkamdjou:


detection-rules/attachment_archive_with_exe.yml
```diff
@@ -19,7 +19,12 @@ source: |
             .file_extension in~ $file_extensions_common_archives
             or .file_type == "rar"
           )
-          and any(file.explode(.), any(.flavors.yara, . == "mz_file"))
+          and any(file.explode(.),
+                  // the YARA scanner alone can be a bit unreliable,
+                  // it's matched on MZ strings in a text file before
+                  any(.flavors.yara, . == "mz_file")
```
rw-access, line 25: WDYT about just `.flavors.mime == "application/x-dosexec"`?

jkamdjou: there's more than just that, eg x-msdownload. we could try to make it an exhaustive list but i wasn't sure id have them all covered (maybe it's just those? not sure). this felt safe enough since combined with the yara flavor

morriscode (approved):

PR #1336, opened 2024-02-26
Updating rules to ensure display name contains a space and ~ was included in the check

morriscode:


jkamdjou (approved):
nice one. do we wanna update VIP impersonation too?

morriscode:
> nice one. do we wanna update VIP impersonation too?

I haven't seen evidence it's needed. Much more likely that the org_display names will contains process accounts than the VIP group. But if we run into it, we definitely can update.

PR #1351, opened 2024-02-29
New Insight Yara Rule Match

morriscode:
image

### DIFF
--- a/insights/attachments/attachment_yara_rule_match.yml
+++ b/insights/attachments/attachment_yara_rule_match.yml
@@ -1,0 +1,11 @@
+name: "Yara Rule Match"
+type: "query"
+source: |
+  map(filter(attachments,
+           .file_type in $file_extensions_common_archives
+           or .file_type in $file_extensions_executables
+           or .file_type == "unknown"
+    ),
+    map(file.explode(.), distinct(map(.scan.yara.matches, .name), .))
+  )
+severity: "medium"


insights/attachments/attachment_yara_rule_match.yml
```diff
@@ -0,0 +1,11 @@
+name: "Yara rule match"
```
jkamdjou, line 1:
```suggestion
name: "Yara signature match"
```

insights/attachments/attachment_yara_rule_match.yml
```diff
@@ -0,0 +1,11 @@
...
+type: "query"
+source: |
+  map(filter(attachments,
+           .file_type in $file_extensions_common_archives
+           or .file_type in $file_extensions_executables
+           or .file_type == "unknown"
+    ),
+    map(file.explode(.), distinct(map(.scan.yara.matches, .name), .))
+  )
+severity: "medium"
```
jkamdjou, line 11: this is interesting... the severity of the yara sig could be anything. but idk what else we would put here

jkamdjou (approved):

insights/attachments/attachment_yara_rule_match.yml
```diff
@@ -0,0 +1,11 @@
+name: "Yara signature match"
```
aidenmitchell, line 1:
```suggestion
name: "YARA signature match"
```

aidenmitchell (changes requested):

insights/attachments/attachment_yara_rule_match.yml
morriscode: I figure if you have yara installed, you probably want to know about any matches. Informational didn't seem strong enough, and critical seemed like overkill.

aidenmitchell (approved):

PR #1372, opened 2024-03-06
Exclude org_domains from spam_onmicrosoft.yml

vector-sec:
We have some instances where mailboxes for automation might send with our .onmicrosoft.com domain, adding this line excludes those messages from this rule.


aidenmitchell (approved):

aidenmitchell:
Thank you Eric!

PR #1377, opened 2024-03-08

morriscode:
Adding tolerance for emails with image attachments that may have other attachments


morriscode:
Test rules looks great

aidenmitchell (approved):

PR #1383, opened 2024-03-12
New Rule: Fake DocuSign HTML table not linking to DocuSign domains

morriscode:
Alternative name: Brand Impersonation: DocuSign HTML Table Forgery

Open to suggestions.


morriscode:
Not seeing this in the test-rules table.....

morriscode:
Test rules are running, but I don't think this should be a blocker. This is pretty tightly scoped and has performed well in hunts already. FN coverage is solid.

morriscode:
Testing looks solid, also matching user reports.

morriscode:
Every test rule match has an AS of malicious, couldn't ask for better!

aidenmitchell (approved):

PR #1386, opened 2024-03-15

aidenmitchell:
Negating documents already signed with Docusign.

### DIFF
--- a/detection-rules/attachment_docusign_image_suspicious_links.yml
+++ b/detection-rules/attachment_docusign_image_suspicious_links.yml
@@ -36,6 +36,10 @@ source: |
                                     "(wichtig|dringend|sofort)"
                  )
                )
+               and not (
+                   strings.ilike(.scan.ocr.raw, "*DocuSigned By*") and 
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_docusign_image_suspicious_links.yml
```diff
@@ -36,6 +36,10 @@ source: |
                                     "(wichtig|dringend|sofort)"
                  )
                )
+               and not (
+                   strings.ilike(.scan.ocr.raw, "*DocuSigned By*") and 
+                   strings.ilike(.scan.ocr.raw, "*DocuSign Envelope ID*")
+              )
```
jkamdjou, line 42: is there anything we could use that might be harder to bypass, like docusign in the headers or exiftools? if not, i think we could roll with this

aidenmitchell: I wish... I looked at the sample and didn't find anything. I think we'll cross that bridge if we get there.

jkamdjou (approved):

PR #1393, opened 2024-03-18
Update credential_phishing_docusign_embedded_image_lure.yml

morriscode:
Finishing tightening this thing up. This knocks out the remainder of the Fp's and a couple I've seen today. There's also a [static file update ](https://github.com/sublime-security/static-files/pull/166)that will take care of the rest from infotrack.co.uk aka SignIT which is an official docusign partner (https://www.docusign.com/en-gb/blog/interview-with-infotrack-and-the-partnership)


morriscode:
/mql-mimic-exempt: 262118, 315071, 323179, 167254, 167258, 185788, 181157

These are all FP's with the exception of 185788 which is a QR code/Docusign lure.

PR #1405, opened 2024-03-21
New Rule: Attachment File scheme link containing executable filtype

morriscode:


detection-rules/attachment_file_scheme_link_to_executable_filetype.yml
```diff
@@ -0,0 +1,43 @@
+name: "Attachment: File scheme link containing executable filetype"
```
jkamdjou, line 1:
```suggestion
name: "Attachment: OLE external relationship containing file scheme link to executable filetype"
```

jkamdjou (approved):

PR #1406, opened 2024-03-22
New Rule: Attachment File scheme link ontaining IP address

morriscode:


detection-rules/attachment_file_scheme_link_containing_ip_address.yml
```diff
@@ -0,0 +1,42 @@
+name: "Attachment: File scheme link containing IP address"
```
jkamdjou, line 1:
```suggestion
name: "Attachment: OLE external relationship containing file scheme link to IP address"
```

jkamdjou (approved):

PR #1443, opened 2024-04-02
New Rule: Attachment: Callback Phishing solicitation via text file with a large unknown recipient list

morriscode:


morriscode:
this feels pretty safe to deploy sans test-rule
Hunt results are all FN TP's.

aidenmitchell (approved):

PR #1454, opened 2024-04-04
Update spam_onmicrosoft.yml

padey:
Added a legit German reply and also added the "X-Auto-Response-Suppress" header to find auto response mails from Exchange. 

**X-Auto-Reponse-Suppress:**
https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxcmail/ced68690-498a-4567-9d14-5c01f974d8b1

### DIFF
--- a/detection-rules/spam_onmicrosoft.yml
+++ b/detection-rules/spam_onmicrosoft.yml
@@ -8,7 +8,11 @@ source: |
     and length(recipients.to) < 2
     and length(body.links) > 0
     // bounce-back negations
-    and not strings.like(sender.email.local_part, "*postmaster*", "*mailer-daemon*", "*administrator*")
+    and not strings.like(sender.email.local_part,
+                         "*postmaster*",
+                         "*mailer-daemon*",
+                         "*administrator*"
+    )
     and not any(attachments,
                 .content_type in (
                   "message/rfc822",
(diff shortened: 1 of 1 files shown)


detection-rules/spam_onmicrosoft.yml
```diff
@@ -3,40 +3,43 @@ description: "The default Microsoft Exchange Online sender domain, onmicrosoft.c
...
+        or regex.imatch(subject.subject, '[^:]?automat.[^:]{0,20}:.*$')
+        // german legit replies
+        or strings.istarts_with(subject.subject, "Abwesen")
+      )
+      and (
+        length(headers.references) > 0
+        and any(headers.hops, any(.fields, strings.ilike(.name, "In-Reply-To")))
+        or any(headers.hops, any(.fields, strings.ilike(.name, "X-Auto-Response-Suppress")))
+      )
     )
```
morriscode, line 34:
```suggestion
      and (
      	 any(headers.hops, any(.fields, strings.ilike(.name, "In-Reply-To")))
         or any(headers.hops,
             	    any(.fields, strings.ilike(.name, "X-Auto-Response-Suppress"))
         )
      )
   )
)
```

padey: makes sense. :) 
hunt was successful and looking good!°

morriscode:
@padey I'm seeing some TP's that are no longer flagging with the X-Auto-Response-Suppress inclusion :/

PR #1461, opened 2024-04-08
New insight: Salesforce HTML attachment

aidenmitchell:
To call out attachments sent from Salesforce (example):
```
<html><head><meta http-equiv="Content-Type" content="text/html; charset=UTF-8"><meta http-equiv="Refresh" content="0; URL=https://sublime.my.salesforce.com/sfc/p/2b111156a/a/Vo0000002a/dskalfjiwaejfiwofjdklsanvdasjfl"></head><body><div>Attachment not opening? Click this link: <a href="https://sublime.my.salesforce.com/sfc/p/2b111156a/a/Vo0000002a/dskalfjiwaejfiwofjdklsanvdasjfl">testdocument.pdf</a></div></html>
```

### DIFF
--- a/insights/attachments/salesforce_html.yml
+++ b/insights/attachments/salesforce_html.yml
@@ -1,0 +1,25 @@
+name: "Salesforce HTML attachment"
+type: "query"
+source: |
+  type.inbound
+  and length(attachments) == 1
+  and any(attachments,
+          (
+            .file_extension in~ ("html", "htm", "shtml", "dhtml")
+            or .file_extension in~ $file_extensions_common_archives
+            or .file_type == "html"
+          )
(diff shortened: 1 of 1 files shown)


morriscode:
Putting on my analyst hat.  What should I infer from seeing this insight?

aidenmitchell:
The attached HTML file is from Salesforce, so it's probably safe. Nice indicator with any HTML rules.

aidenmitchell:
> then it should be a negation in the HTML rules yeah?

usually these trigger `Any HTML file*` rules, do we want to negate there? Seems incongruent to those rules.

Maybe a better name is good? Something like "Salesforce marketing attachment"?

PR #1463, opened 2024-04-08
New Rule: Spam Attendee List solicitation

morriscode:


detection-rules/spam_attendee_list_solicitation.yml
```diff
@@ -0,0 +1,28 @@
...
+severity: "low"
+source: |
+  type.inbound
+  and length(body.current_thread.text) < 2000
+  and length(body.links) < 5
+  and any([body.html.display_text, body.current_thread.text],
+          regex.icontains(body.current_thread.text,
+                          "(Attendee|Member|Participants) (list|database)"
+          )
+  )
```
aidenmitchell, line 13: I'm a little confused about this part, is it supposed to be `regex.icontains(.`?
Also, why both `body.html.display_text` and `body.current_thread.text`?

morriscode: It is supposed to be icontains, I just didn't lower the strings since it doesn't matter anyway :) 
The aggregation of display_text and body.current_thread.text is both an artifact from joshes original rule, but also because in testing without it, we miss some. :/ 

image

aidenmitchell (approved):

PR #1469, opened 2024-04-10
Update SPF and DMARC insights

morriscode:


jkamdjou (approved):

aidenmitchell (commented):
Does `not pass` equal a `fail` though? Just want to check that the name of the insight will still match the value, as we're not directly looking for `*fail` anymore.

morriscode:
> Does `not pass` equal a `fail` though? Just want to check that the name of the insight will still match the value, as we're not directly looking for `*fail` anymore.

yes sir that's the idea!

PR #1482, opened 2024-04-22
Create mismatched_sender_reply_same_name.yml

aidenmitchell:
image


insights/headers/mismatched_sender_reply_same_name.yml
```diff
@@ -0,0 +1,12 @@
+name: "Mismatched sender and reply to address with same display name"
```
morriscode, line 1:
```suggestion
name: "Matching display names, mismatched emails (sender/reply-to)"
```

Our insight names are getting longer which is causing truncation. I tried to shorten it but don't think I succeeded. Consume or disregard if you don't like this.

morriscode (approved):

PR #1485, opened 2024-04-24

morriscode:
Was firing on invalid URL's in body links. Also added negations for office365.com sending addresses.


morriscode:
/mql-mimic-exempt: 308820, 315818 

these are not regressions. The EML's are too large to be processed.

aidenmitchell (approved):

PR #1511, opened 2024-05-06
Update vip_impersonation.yml

aidenmitchell:


detection-rules/vip_impersonation.yml
```diff
@@ -14,7 +14,12 @@ source: |
   type.inbound
   and (
     any($org_vips, .display_name =~ sender.display_name)
-    or any($org_vips, strings.contains(subject.subject, .display_name))
+    or (
```
rw-access, line 17: no need for these parentheses

rw-access (approved):

PR #1513, opened 2024-05-06
Create vip_impersonation_fake_thread.yml

morriscode:


detection-rules/vip_impersonation_fake_thread.yml
```diff
@@ -0,0 +1,46 @@
+name: "VIP Impersonation: Fake thread with display name match, email mismatch"
```
aidenmitchell, line 1:
```suggestion
name: "VIP impersonation: Fake thread with display name match, email mismatch"
```

aidenmitchell (approved):

PR #1518, opened 2024-05-08
Create credential_phishing_hyperlinked_image_with_suspicious_file_sha…

morriscode:
Plan to run in test-rules for a bit. Have a few different ideas to tackle this.


morriscode:
only 1 hit so far.

zoomequipd:
🧹

PR #1539, opened 2024-05-13
Create spam_single_malformed_pdf_romance_scam.yml

morriscode:


morriscode:
Good to go here test rules only showing high AS matches!

aidenmitchell (approved):

PR #1552, opened 2024-05-16
Update body_bec_mobile_solicitation.yml

morriscode:


morriscode:
the regex on line 12 looks incorrect but was purposely crafted to catch the "personalcell" no spaces.

morriscode:
/mql-mimic-exempt: 124728, 140442, 197683, 197680, 189975


These are not the target set.

aidenmitchell (approved):

PR #1557, opened 2024-05-20
Create credential_phishing_fake_email_quarantine_notification.yml

morriscode:


morriscode:
> Is this in favour of, or working together with [`main`/detection-rules/impersonation_microsoft_quarantine_release_in_body.yml](https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/impersonation_microsoft_quarantine_release_in_body.yml?rgh-link-date=2024-05-21T21%3A40%3A20Z) ?

Neither, I've seen enough of these in the last week or so with no branding.

PR #1560, opened 2024-05-21
Create spam_single_recipient_duplicated_in_cc.yml

morriscode:


detection-rules/spam_single_recipient_duplicated_in_cc.yml
```diff
@@ -0,0 +1,95 @@
...
+  and length(recipients.to) == 1
+  and length(recipients.cc) == 1
+  
+  // unsolicited
+  and not profile.by_sender().solicited
+  
+  // recipients to equals cc
+  and any(recipients.to, any(recipients.cc, .email.email == ..email.email))
+  
+  // body is short with spam keywords
```
jkamdjou, line 18: this is for the block on line 20

detection-rules/spam_single_recipient_duplicated_in_cc.yml
```diff
@@ -0,0 +1,95 @@
...
+  and (
```
jkamdjou, line 19: am a little worried about FPs since it only needs to match 1 of these conditions. we should probably test rules it?

detection-rules/spam_single_recipient_duplicated_in_cc.yml
morriscode: not sure what you mean

detection-rules/spam_single_recipient_duplicated_in_cc.yml
morriscode: unsolicted + 1 recipient and 1 cc and they match is almost good enough to ship on it's own.

morriscode: Rule has 150+ matches in with an AS of 90+

aidenmitchell (approved):

PR #1576, opened 2024-05-28

aidenmitchell:


detection-rules/impersonation_quickbooks.yml
```diff
@@ -15,12 +15,39 @@ source: |
...
+    'quickbooksonline.com'
+  )
+  and (
+    not profile.by_sender().any_false_positives
+    and not profile.by_sender().solicited
+  )
   and (
-      not profile.by_sender().any_false_positives 
-      and not profile.by_sender().solicited
+    length(filter(body.links,
```
jkamdjou, line 32: i think a comment for this big block would be helpful. its a little convoluted

PR #1584, opened 2024-05-29
Update body_bec_mobile_solicitation.yml

aidenmitchell:


detection-rules/body_bec_mobile_solicitation.yml
```diff
@@ -10,7 +10,7 @@ source: |
                      "(email|send|shoot|give|provide|attach|confirm|) (me|your)"
   )
   and regex.contains(body.current_thread.text,
-                     "(cell(/s?phone)?|mobile.{0,10}(phone|number|#|no)|whatsapp)"
+                     "(cell(/s?phone)?|mobile.{0,10}|contact.{0,10}(phone|number|#|no)|whatsapp)"
```
jkamdjou, line 13:
```suggestion
                     "(cell(/s?phone)?|(mobile.{0,10}|contact.{0,10})(phone|number|#|no)|whatsapp)"
```
does this need to be this to maintain current parity? assuming this is for FNs

aidenmitchell: Don't think so, it's supposed to be something like `(cellphone OR mobile... OR contact...)` AND `(phone|number|#|no)|whatsapp`

morriscode: Both regex's were looking a bit wonky. Updated both

morriscode:
FWIW this rule was super busted due to a floating pipe. Going to take a minute to fix it.

morriscode (approved):

PR #1586, opened 2024-05-29
Update body_callback_phishing_no_attachment.yml

aidenmitchell:


detection-rules/body_callback_phishing_no_attachment.yml
```diff
@@ -20,6 +20,10 @@ source: |
     or sender.email.domain.tld in $suspicious_tlds
     or network.whois(sender.email.domain).found == false
     or headers.mailer in~ ("Microsoft CDO for Windows 2000")
+    or (
+      length(recipients.to) == 1
+      and all(recipients.to, .email.domain.domain not in $org_domains)
+    )
```
jkamdjou, line 26: this doesn't feel like a "suspicious" signals like the others, any concerns for FPs?

aidenmitchell: I can run hunts with it, but a single recipient that isn't an org domain is weird to me (in conjunction with the rest of the rule)

morriscode: This looks fine to me since the rest of the message is pretty specific. It's definitely weird in this context that the only designated recipient is not a member of the org

morriscode (approved):

PR #1602, opened 2024-05-31
Update credential_phishing_fake_email_quarantine_notification.yml

aidenmitchell:


morriscode:
Assuming this is to negate any collisions with docusign rules?

morriscode (approved):

PR #1603, opened 2024-05-31

aidenmitchell:


morriscode (approved):

morriscode:
/mql-mimic-exempt: 514509

This actually fires but I'm seeing failures due to what I'm guessing is message screenshot

PR #1625, opened 2024-06-07
Create credential_phishing_one_drive_impersonation.yml

morriscode:


morriscode:
Testing looks good here, AS on all matches in the 90s. Matching shared EML's all bad.

zoomequipd (approved):

PR #1641, opened 2024-06-10
Update attachment_office365_image.yml

morriscode:
# Description

accommodating tags preceding the RE:|FWD in the subject

### DIFF
--- a/detection-rules/attachment_office365_image.yml
+++ b/detection-rules/attachment_office365_image.yml
@@ -77,7 +77,7 @@ source: |
             )
     )
   )
-
+  
   // negate replies
   and (
     (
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_office365_image.yml
```diff
@@ -96,6 +96,7 @@ source: |
           or strings.istarts_with(subject.subject, "AW:")
           or strings.istarts_with(subject.subject, "TR:")
           or strings.istarts_with(subject.subject, "FWD:")
+          or regex.imatch(subject.subject, '(\[[a-z]+\]\s?){0,3}(re|fwd?)\s?:')
```
zoomequipd: In reviewing this PCRE with regex101, found that it did NOT match where it should have due to a space in one of the tags `[Action Required]`
https://regex101.com/r/WWlrhx/1 (when accounting for the anchored behavior of the `match`)

Updated the regex to account for anything between `[` and `]` 
https://regex101.com/r/WWlrhx/2

reference to delete the regex101 entry if ever needed: 
delete/Wn49RS5RUFulhHAd8UZ5lWfb71nKASDYZJnZ

Will update this regex to account for it: `(\[[^\]]+\]\s?){0,3}(re|fwd?)\s?:`

aidenmitchell (approved):

PR #1679, opened 2024-06-20
GitOps changes

aidenmitchell:
# Description

Various QoL changes for GitOps.

### DIFF
--- a/.github/workflows/update-test-rules.yml
+++ b/.github/workflows/update-test-rules.yml
@@ -151,3 +151,17 @@ jobs:
           
           git commit --allow-empty -F ../message.txt
           git push origin test-rules
+  label-pr:
+    name: Label PR
+    needs: add-to-test-branch
+    runs-on: ubuntu-20.04
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v3
+
+      - name: Add label
+        uses: actions-ecosystem/action-add-labels@v1
+        with:
+          labels: in-test-rules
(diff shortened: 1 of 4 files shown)


.github/workflows/update-test-rules.yml
```diff
@@ -151,3 +151,17 @@ jobs:
...
+    permissions:
+      pull-requests: write
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v3
+
+      - name: Add label
+        uses: actions-ecosystem/action-add-labels@v1
+        with:
+          labels: in-test-rules
```
cameron-dunn-sublime, line 167: I _think_ this won't know which PR to comment on. If you have tested or read documentation that suggests otherwise, we can give it a go but IIRC:

```
on:
  issue_comment:
    types: [created]
```

which is what triggers this workflow is always run with a context for the `main` branch, not a PR. The "Get PR branch" step above solves for this (although it only gets the branch, not the PR).

Maybe "actions-ecosystem/action-add-labels@v1" is smart about this though?

Also any reason for a separate job not just a separate step?

aidenmitchell: It might be smart enough? From their README:

> This action extract the number from an issue or a pull request which has triggered this by default. It means you don't need to care about something annoying like whether you should use `${{ github.event.issue.number }}` or `${{ github.event.pull_request.number }}`.

> Also any reason for a separate job not just a separate step?

Not really, more just for visual distinction in the GHA monitoring UI. Doesn't matter to me either way!

cameron-dunn-sublime: > From their README: ...

Both of the example usages they show wouldn't require the hoops we're jumping through for getting the branch, so I'm still a little doubtful it'll work but we can try it for sure.

cameron-dunn-sublime (approved):

PR #1696, opened 2024-06-27
Update fake_thread_suspicious_indicators.yml

morriscode:
# Description

Moving the bail out logic for fake thread to apply to the entire condition. Also expanding tag tolerance in the subject line to include re|fwd? and automatic*


morriscode:
/mql-mimic-exempt: 191778, 254213, 252066, 343227, 367853, 361869, 367968, 479127, 518792

zoomequipd (approved):

PR #1718, opened 2024-07-10
Update link_cyrillic_substitutions_unsolicited.yml

morriscode:
# Description
Negating atlassian emails from eastern european senders

The sender.email is dynamically generated which is problematic when trying to use it to make exemptions to this rule
image


morriscode:
/mql-mimic-exempt: 356776, 358541, 358542, 358552, 358546

ghosts.

aidenmitchell (approved):

PR #1732, opened 2024-07-12
Migration of Advance Fee to intent

bfilar:
# Description
Migrating all references of `advance_fee` as a `tag` to an `intent`. Also including `advance_fee` as an intent in a few `bec`-themed rules


bfilar:
/mql-mimic-exempt: 124728, 140442, 197683, 197680, 189975, 477824

PR #1745, opened 2024-07-17
Update impersonation_sharepoint_fake_file_share.yml

zoomequipd:
# Description
Negate message groups which are from SharePoint actual based on the defined message-id. 

The "Fake" file share was triggering on a subset of message group that were not fake. 

## Associated hunts

Message Groups that will no longer fire this rule
-[ Hunt 1](https://platform.sublimesecurity.com/hunts/df0b65fd-ce5e-4088-8ea3-cce32eb3fc9c)


zoomequipd:
/mql-mimic-exempt: 258752, 324490, 329556, 144088, 144089, 144090, 144091, 144092, 343349

morriscode (approved):

PR #1753, opened 2024-07-18

aidenmitchell:
# Description

Correcting `any_false_positives` logic to apply in all cases, and adding numerous negations.

# Samples

- https://platform.sublimesecurity.com/messages/9a8162ceaddaa575c5cc605b814654796a6ed9d491313e38e93973b57af5ccb7
- https://platform.sublimesecurity.com/messages/fbb4bd6dcb760c64bc17830de81293695a5b50202e7b5809d1942b9e26c21110

### DIFF
--- a/detection-rules/qr_code_suspicious_indicators.yml
+++ b/detection-rules/qr_code_suspicious_indicators.yml
@@ -34,7 +34,7 @@ source: |
                   )
                   and not (
                     .scan.exiftool.image_height > 3000
-                    or .scan.exiftool.image_width > 3000
+                    and .scan.exiftool.image_width > 3000
                   )
           )
       )
--- a/detection-rules/qr_code_suspicious_indicators.yml
+++ b/detection-rules/qr_code_suspicious_indicators.yml
@@ -300,9 +300,10 @@ source: |
     )
     or (
       profile.by_sender().any_messages_malicious_or_spam
-      and not profile.by_sender().any_false_positives
     )
   )
+  
+  and not profile.by_sender().any_false_positives
   
   // negate highly trusted sender domains unless they fail DMARC authentication
   and (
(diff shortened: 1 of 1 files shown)


morriscode:
/mql-mimic-exempt: 29518,57918,112340,111602,114015,114016,111578,113081,143516,113550,115144,110786,116335,110367,112301,129455,139681,178600,136615,136614,136751,136750,140518,136752,140517,143892,159340,178582,178585,178583,178589,178587,178586,178588,178591,178592,178590,178593,178595,178596,178594,178597,178599,178598,184836,184840,184837,187510,187512,187511,187513,187514,187518,187516,187517,187515,187519,187520,187521,187522,187523,187524,187525,187526,194812,245996,246012,246003,246238,233791,268380,268386,282175,268414,304666,188275,188917,280268,294686,251033,352269,294781,352307,167058,140048,438966,326499,183062,568911

detection-rules/qr_code_suspicious_indicators.yml
```diff
@@ -53,7 +55,35 @@ source: |
           strings.icontains(sender.display_name, .email.domain.sld)
       )
       or length(body.current_thread.text) is null
-      or body.current_thread.text == ""
+      or (
```
zoomequipd, line 58: just to make sure i'm understanding this logic correctly, we're adding a new condition that when there is at least one reference, the suspicious indicator is that it doesn't have "In-Reply-To" header and the subject doesn't indicate a reply/forward, which would normally be included when a message includes references. 

New logic breakdown: 
// current thread is empty (could be evaded by "Sent with Microsoft Outlook" auto signatures, warning banners, or actual signatures)
AND 
  // it has references
  // but doesn't have a "In-Reply-To" header
  // and the subject doesn't indicate a reply/forward
  OR
  // it has no references.

aidenmitchell: Correct, except that `current thread is empty` isn't actually new, it's just the git diff being silly. I was adding extra logic to that check

zoomequipd (approved):

PR #1771, opened 2024-07-24
Update fake_thread.yml

zoomequipd:
# Description

Better account for forwards and tags in the subject.

## Associated hunts

New matching messages:
- [Hunt 1](https://platform.sublimesecurity.com/hunts/fcc589bc-d9fb-4c4a-b917-890b1ce8c396)


zoomequipd:
going to combine with other related work into a single PR

PR #1772, opened 2024-07-24
Update Fake Threads logic 

zoomequipd:
# Description

Update Fake Thread rules with the objectives of:

- Better account for forwards and emails with tags in subjects
- Use consistent logic
- Add additional conditions to find the forward/reply "headers" in emails

## Associated hunts

- [Insight - New Matches](https://platform.sublimesecurity.com/hunts/fcc589bc-d9fb-4c4a-b917-890b1ce8c396)
- [impersonation_fake_msg_thread_mismatched_from_freemail_replyto.yml - New Matches](https://platform.sublimesecurity.com/hunts/1f7ac530-2a2e-442c-a42d-8f1b54a7e1c2)
	I had to exclude a the `any(header.hops, any(.fields, ...` logic to get the hunt to complete
- vip_impersonation_fake_thread.yml
	No delta in Shared EMLs, even without the $org_vip domains

### DIFF
--- a/detection-rules/fake_thread_suspicious_indicators.yml
+++ b/detection-rules/fake_thread_suspicious_indicators.yml
@@ -30,11 +30,15 @@ source: |
           or strings.istarts_with(subject.subject, "FWD:")
         )
       )
-      and 3 of (
-        strings.icontains(body.html.display_text, "from:"),
-        strings.icontains(body.html.display_text, "to:"),
-        strings.icontains(body.html.display_text, "sent:"),
-        strings.icontains(body.html.display_text, "subject:")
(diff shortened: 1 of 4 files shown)


zoomequipd:
/mql-mimic-exempt: 191778, 254213, 252066, 343227, 367853, 361869, 367968, 479127, 518792

previously excluded here: https://github.com/sublime-security/sublime-rules/pull/1696#issuecomment-2195211557
this isn't an issue with the change made in this commit.

zoomequipd:
changed up the logic to handle header.in_reply_to as an OR instead of an AND

morriscode (approved):

morriscode (approved):

PR #1773, opened 2024-07-24
Create impersonation_gmail_av_scan.yml

zoomequipd:
# Description

Detect fake Google Attachment language within the body of a message. 

## Associated hunts

- [Hunt 1](https://platform.sublimesecurity.com/hunts/2eb9d4ad-9373-400b-8e38-93533296d1ff)


zoomequipd:
Been in test rules for a couple weeks with only a single hit. low volume, but seems to be high confidence

detection-rules/attachment_fake_gmail_attachment.yml
```diff
@@ -0,0 +1,30 @@
+name: "Impersonation: Fake Gmail Attachment"
+description: "Message detects fake Gmail attachments by inspecting the body of a message for elements found within Gmail's user interface for attachment. In expected use, these elements only appears within the gmail WebUI and not within the body of message . The presence of this within message indciates a fake attachment."
```
morriscode, line 2:
```suggestion
description: "Message detects fake Gmail attachments by inspecting the body of a message for elements found within Gmail's user interface for attachment. In expected use, these elements only appears within the gmail WebUI and not within the body of message. The presence of this within message indicates a fake attachment."
```

detection-rules/attachment_fake_gmail_attachment.yml
```diff
@@ -0,0 +1,30 @@
...
+  
+  // negate highly trusted sender domains unless they fail DMARC authentication
+  and (
+    (
+      sender.email.domain.root_domain in $high_trust_sender_root_domains
+      and not headers.auth_summary.dmarc.pass
+    )
+    or sender.email.domain.root_domain not in $high_trust_sender_root_domains
+  )
+  and not profile.by_sender().any_false_positives
```
morriscode, line 20: Apparently this should only be used to short circuit any_messages_malicious_or_spam, and was never intended to stand on it's own, however we've been using it as a standalone probably incorrectly since the false_positive flag isn't anchored to the afflicting rule. Not sure if we want to start correcting that here? 

eg.
```
(
  not profile.by_sender().solicited
  or (
    profile.by_sender().any_messages_malicious_or_spam
    and not profile.by_sender().any_false_positives
  )
)
```

zoomequipd: I don't believe including `not profile.by_sender().solicited` based on the sample should be included. 

It introduces FNs of spoofed messages where the org has received message with the actual sender (unspoofed sender)
  
If `.any_false_positives` is not intended to stand on its own, then in this case perhaps a modification of the one you've proposed will be acceptable.  
```
  and not (
      profile.by_sender().any_messages_malicious_or_spam
      and profile.by_sender().any_false_positives
  )
```  
or perhaps no scoping logic at all should be included in this rule.

morriscode: I think you want +any_messages_malicious_or_spam, but not -any_false_positives. 

So to properly do that

`and (
  (
    profile.by_sender().any_messages_malicious_or_spam
    and not profile.by_sender().any_false_positives
  )
  or not profile.by_sender().any_messages_malicious_or_spam
)`

morriscode (approved):

PR #1651, opened 2024-06-12
Update attachment_callback_phish_with_pdf.yml

aidenmitchell:
# Description

Negating legitimate replies.

# Associated samples

https://platform.sublimesecurity.com/messages/9f53393c381e9e78cc5f4d16e7f6ef36aebe5fe899908f309faf89bd5f36a840

PR #1783, opened 2024-07-27
Update attachment_callback_phish_with_pdf.yml

zoomequipd:
# Description

- Reduce FPs by enforcing that the conditions for matching within the pdf all occur with in the same file.explode() object.
- Increase TP matches (and thus satisfying mimic failures) by adding variations of mcafee with the logic.

# Associated samples

FP Sample that no longer matches
- [Sample 1](https://platform.sublimesecurity.com/messages/a6ce79edef70dd0120320585528af4154e91382c84892b6d961a085c40cdf067)

### DIFF
--- a/detection-rules/attachment_callback_phish_with_pdf.yml
+++ b/detection-rules/attachment_callback_phish_with_pdf.yml
@@ -14,22 +14,25 @@ source: |
       and not profile.by_sender().any_false_positives
     )
   )
-
+  
   // single attachment
   and length(attachments) == 1
-
+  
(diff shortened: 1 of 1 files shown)


zoomequipd:
/mql-mimic-exempt: 36067, 56613, 120761, 36012

36067 = FP
56613, 120761, 36012  = a Forwarded message - these were previously exempted in https://github.com/sublime-security/sublime-rules/pull/1651#issuecomment-2166491986

morriscode (approved):

PR #1784, opened 2024-07-28
Create link_webflow_unsolicited.yml

zoomequipd:
# Description

Match on messages from unsolicited senders with at least one link to webflow.io

# Associated samples

- [Sample 1](https://platform.sublimesecurity.com/messages/8de67507028f2581e7d549412b22bbb38eaed9b0371c6f3dd2576273016298f7)
- [Sample 2](https://platform.sublimesecurity.com/messages/408afbf8b6ecf0f7379cac60d0bbd2bb20f36d45932bd4c71e0ca168bdae3cab)
- [Sample 3](https://platform.sublimesecurity.com/messages/61c91112dd08ad68c29f1879f9862e5d69cc0f04a0bc7430a3fd7f0426ad0fdd)

## Associated hunts

- [Hunt 1](https://platform.sublimesecurity.com/hunts/bc4e4897-482c-48be-b844-45690ba7a967)

### DIFF
--- a/detection-rules/link_webflow_unsolicited.yml
+++ b/detection-rules/link_webflow_unsolicited.yml
@@ -11,7 +11,7 @@ source: |
           and .href_url.domain.subdomain != "www"
           and .href_url.path == "/"
   )
-  // not solicited or from malicious/spam user with now FPs
+  // not solicited or from malicious/spam user with no FPs
   and (
     not profile.by_sender().solicited
     or (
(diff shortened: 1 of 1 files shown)


zoomequipd:
low volume while in test rules. Shared EML in the last 14 had FPs, but they were only new sender in shared emls. Prior to that all hits look good. 

Official webflow communications are sourced from webflow.com and not webflow.io. and the little `and .href_url.path == "/"` section actually pulls a lot of weight in removing benign matches.

I had considered added some logic that matched the "default" publish domain which always ends with `-<6 hex chars>` but [urlscan.io](https://urlscan.io/search/#domain%3Awebflow.io) reveled that would FN too much.

morriscode:
Do you think this is ASR?

zoomequipd:
It seems to be doing well enough, I suppose if we get a bunch of FP we can make an ASR rule.

morriscode (approved):

PR #1793, opened 2024-07-30
Update attachment_pdf_malicious_creator_juliep.yml

zoomequipd:
# Description

Match additional samples by opening up the page count

# Associated samples

- [Sample 1](https://platform.sublimesecurity.com/messages/99033ace1b8e237cf48811d616bf223ee1219a88919a6a3bfdc45e1e7c15d045)
- [Sample 2](https://platform.sublimesecurity.com/messages/663fa1fee261fe7324d32209b55bf5ecd8d446be4a1269610b1f657b33d30a1a)

### DIFF
--- a/detection-rules/attachment_pdf_malicious_creator_juliep.yml
+++ b/detection-rules/attachment_pdf_malicious_creator_juliep.yml
@@ -16,7 +16,6 @@ source: |
                            and strings.icontains(.value, 'DynaPDF')
                     )
                   )
-                  and .scan.exiftool.page_count == 2
                   and (
                     any(.scan.exiftool.fields,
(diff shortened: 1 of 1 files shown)


morriscode:
I think this is good to go. 

One question do you think it's worth moving the (Julie P) section of the title like.

`"Attachment: PDF Author Associated with Malicious Emails (Julie P.)"`
to 
`""Attachment: PDF Author (Julie P.) Associated with Malicious Emails"`

When I see the name at the end It feels like that could be the title of the author of the rule, or the sender of the malicious emails? Moving it closer to the Author section makes it work in my head. Thoughts?

morriscode (approved):

PR #1796, opened 2024-07-30
Update impersonation_dropbox.yml

aidenmitchell:
# Description

Adding link check to flag on display URL == dropbox.com but with a mismatch.

# Associated samples

- https://platform.sublimesecurity.com/messages/3dca3712fbb6994e5a0401f9330ed444e640c9a32bfd3feadc0d0bbb2fea984d

### DIFF
--- a/detection-rules/impersonation_dropbox.yml
+++ b/detection-rules/impersonation_dropbox.yml
@@ -10,7 +10,9 @@ source: |
     or strings.ilevenshtein(sender.display_name, 'dropbox') <= 1
     or strings.ilike(sender.email.domain.domain, '*dropbox*')
     or any(body.links,
-       .display_url.domain.root_domain == "dropbox.com" and .mismatched
+           .display_url.domain.root_domain == "dropbox.com"
+           and .mismatched
+           and not .href_url.domain.root_domain == "mimecast.com"
     )
   )
   and sender.email.domain.root_domain !~ 'dropbox.com'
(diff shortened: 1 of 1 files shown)


morriscode:
Added to test rules, I'm guessing this is fine but just wanted to double check since a mismatch could be for a few reasons?

zoomequipd:
looks OK in test rules 3 new matches compared to existing version (4 matches total) but, I think there's a chance of FPs on URL rewriters for which there are not decoders. I did run a hunt, which turned up nothing though, so maybe not as big of a concern as I think. 

https://platform.sublime.security/hunts/8e094495-eec7-4e0c-9902-eef441df9f5b

aidenmitchell:
Just in case, I added a Mimecast negation.

detection-rules/impersonation_dropbox.yml
```diff
@@ -9,6 +9,11 @@ source: |
     strings.ilike(sender.display_name, '*dropbox*')
     or strings.ilevenshtein(sender.display_name, 'dropbox') <= 1
     or strings.ilike(sender.email.domain.domain, '*dropbox*')
+    or any(body.links,
+           .display_url.domain.root_domain == "dropbox.com"
+           and .mismatched
+           and not .href_url.domain.root_domain == "mimecast.com"
```
zoomequipd, line 15: `mimecastprotect.com`?

zoomequipd (changes requested):

zoomequipd (approved):

PR #1808, opened 2024-08-03
Create open_redirect_pirlsandiego.yml

zoomequipd:
# Description
Match message containing an open redirect from PIRL San Diego
# Associated samples

- [Sample 1](https://platform.sublimesecurity.com/messages/ce5e11d10c4b26493f4e00c851dce757b40ca047470924b033a88b721361933e)

## Associated hunts

- [Hunt 1](https://platform.sublimesecurity.com/hunts/e33f67e7-0877-447a-ace8-4b5d887b6ee7)

### DIFF
--- a/detection-rules/open_redirect_pirlsandiego.yml
+++ b/detection-rules/open_redirect_pirlsandiego.yml
@@ -33,3 +33,4 @@ tactics_and_techniques:
 detection_methods:
   - "Sender analysis"
   - "URL analysis"
+id: "a057d64c-71a1-554a-8ff1-7016d5cae8cf"
--- a/detection-rules/open_redirect_pirlsandiego.yml
+++ b/detection-rules/open_redirect_pirlsandiego.yml
@@ -1,0 +1,35 @@
+name: "Open Redirect: PIRL San Deigo"
+description: |
+  Message contains use of the PIRL San Deigo open redirect. This has been exploited in the wild.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and any(body.links,
+          .href_url.domain.root_domain == "pirlsandiego.net"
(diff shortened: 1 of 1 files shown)


zoomequipd:
Low volume hits while deployed in test.  Ready for review

detection-rules/open_redirect_pirlsandiego.yml
```diff
@@ -0,0 +1,36 @@
...
+  Message contains use of the PIRL San Deigo open redirect. This has been exploited in the wild.
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and any(body.links,
+          .href_url.domain.root_domain == "pirlsandiego.net"
+          and strings.icontains(.href_url.path, '/LinkClick.aspx')
+          and strings.icontains(.href_url.url, 'link=')
+  )
```
aidenmitchell, line 12:
```suggestion
  )
  and sender.email.domain.root_domain != 'pirlsandiego.net'
```

zoomequipd: `pirlsandiego.net` doesn't have an MX record, so they must not be sending email from that.

Based on what I could find on their website, this is used to redirect users out of pirlsandiego.net, lots of examples on this page: http://pirlsandiego.net/Npirl/ContactInfo/tabid/541/Default.aspx

I'm confident it'll be abnormal to see this redirect in messages.

If we want to add confidence, we might be able to negate links where `tabid` and `mid` are present within the URL query params, which hasn't been observed in malicious use but is present on the website.  If we start seeing FPs, I'll get that added in quickly.

aidenmitchell: Makes sense, thanks for the context 🫡

aidenmitchell (approved):

PR #1812, opened 2024-08-06
Update names.yml

morriscode:
# Description

Updating the insight to produce "unnamed attachment" when a file name is null. This keeps the array count in line with the message details.


zoomequipd:
confirmed in manual testing
With `morriscode-patch-40` branch: 
image

With `main` branch: 
image

zoomequipd (approved):

PR #1818, opened 2024-08-08

zoomequipd:
# Description

Adjust regex to not match on image OCRs which include "DocuSign" and "Graphics Design"

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/33f6d1f4a27e30de3dd3917ac52aebc7e08826eaa4b895bb3eb6eda90e4e1fdf)

### DIFF
--- a/detection-rules/attachment_docusign_image_suspicious_links.yml
+++ b/detection-rules/attachment_docusign_image_suspicious_links.yml
@@ -19,8 +19,8 @@ source: |
   )
   and (
     any(attachments,
-    .file_type in $file_types_images
-        (
+        .file_type in $file_types_images
+        and (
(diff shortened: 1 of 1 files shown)


zoomequipd:
Rule is actually scanning more than just images, not sure if it's intended, but was the result of many mimic failures.  Several of them had no match on other rules.

morriscode:
> Rule is actually scanning more than just images, not sure if it's intended, but was the result of many mimic failures. Several of them had no match on other rules.

I don't think it was intended, looks like we didn't scope the post file explode inspection. 
Might be an accidental win? I'm good with shipping as is for now.

morriscode (approved):

PR #1821, opened 2024-08-09
Update impersonation_human_resources.yml

zoomequipd:
# Description

Negate benign messages from HR Interests
# Associated samples

- [Sample 1](https://platform.sublime.security/messages/d1a809aaa7b758b94ef25efacf707f18335844a6246ffcc48a07b76632312a80)

### DIFF
--- a/detection-rules/impersonation_human_resources.yml
+++ b/detection-rules/impersonation_human_resources.yml
@@ -21,7 +21,7 @@ source: |
   )
   // Negate common marketing mailers
   and not regex.icontains(sender.display_name,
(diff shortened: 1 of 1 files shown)


zoomequipd:
/mql-mimic-exempt: 181406, 413523, 413524

181406 - hasn't alerted since adding high_trust_sender exclusion added in #877 and has been excluded in other runs since #1710
413523, 413524 - these two actually match in analyzer.  However, they do not match in shared-emls due to the profile.by_sender().prevalence not being "new" or "outlair" but instead "uncommon". -- documented this in the tracking doc

PR #1822, opened 2024-08-09
Create brand_impersonation_ms_planner.yml

aidenmitchell:
# Description

Seeing an increase in MS Planner impersonation.


morriscode:
Relatively low matches in test-rules, no rule overlap, and no attack score hits. 
I think this is safe to push

morriscode (approved):

PR #1870, opened 2024-08-30
Update link_credential_phishing_voicemail_language.yml

aidenmitchell:
# Description
Negating "listen to... podcast"

# Associated samples
- https://platform.sublime.security/messages/c2862cb2c964dddefd7f1cfc9a26b6849cea3d758b5a4dc37083aebc43c74524

PR #1875, opened 2024-09-03

vector-sec:
# Description

Removed false positives related to audio related emails, such as podcasts, that were being caught by this rule. Also added verizonwireless.com as an exclusion because they send wav files, but the content-type is PDF for some reason, so they were also flagged at times.

### DIFF
--- a/detection-rules/link_credential_phishing_voicemail_language.yml
+++ b/detection-rules/link_credential_phishing_voicemail_language.yml
@@ -103,7 +103,8 @@ source: |
     "magicjack.com",
     "unitelvoice.com",
     "voipinterface.net",
-    "ringcentral.biz"
+    "ringcentral.biz",
+    "verizonwireless.com"
   )
   and not any(attachments, strings.starts_with(.content_type, "audio"))
   // filter out topics that mention audio and listening, but are not voicemails
--- a/detection-rules/link_credential_phishing_voicemail_language.yml
+++ b/detection-rules/link_credential_phishing_voicemail_language.yml
@@ -106,6 +106,12 @@ source: |
     "ringcentral.biz"
   )
   and not any(attachments, strings.starts_with(.content_type, "audio"))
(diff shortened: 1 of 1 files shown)


detection-rules/link_credential_phishing_voicemail_language.yml
```diff
@@ -103,9 +103,17 @@ source: |
...
     "voipinterface.net",
-    "ringcentral.biz"
+    "ringcentral.biz",
+    "verizonwireless.com",
+    "t-mobile.com"
   )
   and not any(attachments, strings.starts_with(.content_type, "audio"))
+  // filter out topics that mention audio and listening, but are not voicemails
+  and not any(["podcast", "tune", "album", "episode", "study", "series", "coach"],
+              any([subject.subject, sender.display_name, body.current_thread.text],
```
zoomequipd, line 113: I like this idea, but this current implementation causes FNs when any of `subject.subject, sender.display_name, body.current_thread.text` are null (display_name and body.current_thread.txt both observed causing FNs)

It might be worth adding this logic to the existing check of these fields as seen/proposed in: https://github.com/sublime-security/sublime-rules/pull/1870/files.  This avoids looping the same details twice and allows for more specific inspection of the otherwise matching field.

zoomequipd: Just to be clear, is that something you plan on doing or would you like me to make the changes to your PR?

vector-sec: @zoomequipd just updated, what do you think of this revised version?

zoomequipd: 🎯  that's _exactly_ what I was thinking.

zoomequipd:
/mql-mimic-exempt: 319824, 409756

319824, 409756 = both actually fire on analyzer

aidenmitchell (approved):

PR #1882, opened 2024-09-06
Create attachment_pdf_extortion_leveraging_breach_data.yml

morriscode:
# Description

Covers recent extortion permutations leveraging breach data, such as addresses, phone numbers and often pulling streetview images to further bolster confidence and fear. 

- Sample 1
- Sample 2

## Associated hunts

https://platform.sublime.security/hunts/f02f02b8-7cec-475a-9fb0-66b228bb784f

### DIFF
--- a/detection-rules/attachment_pdf_extortion_leveraging_breach_data.yml
+++ b/detection-rules/attachment_pdf_extortion_leveraging_breach_data.yml
@@ -1,0 +1,46 @@
+name: "Extortion / Sextortion - PDF attachment leveraging breach data from freemail sender"
+description: "Detects sextortion attempts leveraging breach data, including names, addresses, phone numbers and frequently using Google Maps/Bing Maps streetview images to bolster confidence and fear."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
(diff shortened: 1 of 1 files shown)


detection-rules/attachment_pdf_extortion_leveraging_breach_data.yml
```diff
@@ -0,0 +1,50 @@
...
+    or subject.subject == body.current_thread.text
+  )
+  
+  // there's a PDF attachment with an image at a depth of one, measuring 148x148 and containing a QR code that is a BTC address
+  and any(attachments,
+          .file_type == "pdf"
+          and any(file.explode(.),
+                  (.depth == 1 and .flavors.mime == "image/jpeg")
+                  and .scan.exiftool.image_height == 148
+                  and .scan.exiftool.image_width == 148
```
aidenmitchell, line 31: Do you feel this is too specific?

morriscode: I don't think so, it seems to be whatever they're generating the QR code with has a specific size.

aidenmitchell (approved):

PR #1891, opened 2024-09-10
Create credential_phishing_zoho_sign_template_abuse.yml

morriscode:
# Description

New rule exploring Zoho Signs template for credential phishing 

# Associated samples

[- Sample 1](https://platform.sublime.security/messages/7f89931be3bd93abd32817f149d8fad87042324b8e74216bcf5aadf8917ffe37)


zoomequipd:
ran a wide hunt today, this rule looks good, might be a bit narrowly scoped, but until we have more samples, I think it's a good starting point. 

https://platform.sublime.security/hunts/160f9301-34d1-4085-bacb-6fd703af0400

@morriscode thoughts?

zoomequipd (approved):

PR #1936, opened 2024-09-20
Create Impersonation_vip_fake_thread_not_recip.yml

zoomequipd:
# Description

Add coverage for additional VIP Impersonation Fake Thread campaigns.


zoomequipd:
due to limitations on list nesting, this rule produces too many FPs

PR #1941, opened 2024-09-23

zoomequipd:
# Description

Adjust parameter to catch a FN
 
# Associated samples
Samples that will match after this change.  
- [Sample 1](https://platform.sublime.security/messages/3ef5615a12054541a356b3a1cf9acf4c0584993f76d38289e05f742b8fca07e0) - length of "urgency" w/ request display text
- [Sample 2](https://platform.sublime.security/messages/b211bc2144925c33057b55979dd0937905167d35ed75de1630b2e536e8f1d9a6)  - redirect length and sender display name

### DIFF
--- a/detection-rules/link_credential_phishing_intent_and_other_indicators.yml
+++ b/detection-rules/link_credential_phishing_intent_and_other_indicators.yml
@@ -329,7 +329,7 @@ source: |
       length(filter(ml.nlu_classifier(body.current_thread.text).entities,
                     .name == "urgency"
              )
-      ) >= 3
+      ) >= 2
       // and any body links
       and any(body.links,
               // display text contains a request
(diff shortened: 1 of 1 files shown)


aidenmitchell:
How do we feel about test-rules'ing this one for a bit? I feel like we already see FPs from this rule as-is, and this change could make that worse. Thoughts?

zoomequipd:
I think this change is ok because it still has to be combined with an NLU detected "request" within the display text. 

```
      // multiple entities displaying urgency 
      length(filter(ml.nlu_classifier(body.current_thread.text).entities,
                    .name == "urgency"
             )
      ) >= 2
      // and any body links
      and any(body.links,
              // display text contains a request
              any(ml.nlu_classifier(.display_text).entities, .name == "request")
      )
```      

Run me see if I can get a hunt showing the newly matched messages over L14D

aidenmitchell (approved):

morriscode (approved):

PR #1942, opened 2024-09-23
Update link_multistage_microsoft_forms.yml

zoomequipd:
# Description

1) Rule optimization: move sender profile check outside of the "any" and within a simple `and`
2) avoid FP when a form contains more than one page 

# Associated samples
Resolved FP
- [Sample 1](https://platform.sublime.security/messages/a4dde0ecdf4731110e71c04d6d59e0299d51d91f1382fcc585c0becfbebc4b44)


zoomequipd:
100% coverage compared to existing rule.  this should be good to go.

zoomequipd:
updated PR to address FP found today as well. I still believe this is good to go.

PR #1957, opened 2024-09-27
Create brand_impersonation_docusign_pdf_with_suspicious_links.yml

morriscode:
# Description

New rule to detect PDF's with DocuSign logos, a CTA and no links to docusign.


morriscode:
I see no hits in testing, should be safe to push, but i readded to test rules to validate. This was FN coverage.

PR #1974, opened 2024-10-04
Update link_multistage_google_drive.yml

zoomequipd:
# Description

add additional logic to trim down the number of LA clicks by:
1) Ensuring the Sender Display Name does not appear within $org_display_names
2) Matches on observed file name patterns before continuing to LA

This should also cut down on the number of FNs as a result of LA not being able to fetch the page
# Associated Hunts
- [Hunt 1](https://platform.sublime.security/hunts/ceed635e-7459-4c05-91e6-2ad544a56e49)

### DIFF
--- a/detection-rules/link_multistage_google_drive.yml
+++ b/detection-rules/link_multistage_google_drive.yml
@@ -26,6 +26,9 @@ source: |
       or strings.istarts_with(sender.display_name, strings.concat(., " (Google "))
   )
   
+  // threat actors dont want others to edit the share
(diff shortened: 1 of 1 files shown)


zoomequipd:
LA rule so no test-rules will let this bake a bit in shared EMLs

zoomequipd:
tightened more after feedback

zoomequipd:
looks good, ran L30d hunt in a few places.

morriscode (approved):

PR #1996, opened 2024-10-09
Create link_adobe_share_unsoliticed_sender.yml

zoomequipd:
# Description

Add new coverage from observed abused Adobe message which link to an Adobe hosted file which links to cred phishing. 

# Associated samples

Sample was provided privately


zoomequipd:
looking good, very low volume based on telemetry data

detection-rules/link_adobe_share_unsoliticed_sender.yml
```diff
@@ -0,0 +1,31 @@
+name: "Link: Adobe Share from Unsolicited Sender"
+description: "This attack surface reduction rule matches on message from Adobe which was sent by an email address (as determined by the sender display name) which doesn't appear to have a relationship with the recipient organization."
```
aidenmitchell, line 2:
```suggestion
description: "This attack surface reduction rule matches on messages from Adobe which were sent by an email address (as determined by the sender display name) which doesn't appear to have a relationship with the recipient organization."
```

aidenmitchell (changes requested):

zoomequipd:
going to rework this based on customer feedback

aidenmitchell (approved):

PR #2016, opened 2024-10-16
Update impersonation_sharepoint_fake_file_share.yml

aidenmitchell:
# Description

Added additional logic to catch FNs.

# Associated samples

- https://platform.sublime.security/messages/506efaa24f4ea591cf9f72441416b9c79a4cc0a422a7a89fdbee753d85c7ea45

### DIFF
--- a/detection-rules/impersonation_sharepoint_fake_file_share.yml
+++ b/detection-rules/impersonation_sharepoint_fake_file_share.yml
@@ -16,7 +16,10 @@ source: |
                         "*invited you to access a file*"
           )
       )
-      and strings.ilike(subject.subject, "*shared*", "*updated*")
+      and (
+        strings.ilike(subject.subject, "*shared*", "*updated*")
+        or subject.subject is null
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_sharepoint_fake_file_share.yml
```diff
@@ -183,6 +186,15 @@ source: |
         regex.icontains(body.html.raw, 'rgb\(([0-4]),\s?(18[0-1]),\s?(24[0])\)'),
         regex.icontains(body.html.raw, 'rgb\((25[0-5]),\s?(20[0-2]),\s?([0-7])\)')
       )
+      or (
```
zoomequipd, line 189: At first this new logic gave me concern because it would match on messages that were sent from the org_domain which might actually contain the recipient domain as well.  But there is already logic to negate where the sender domain is in the $org_domains.

zoomequipd (approved):

PR #2022, opened 2024-10-17
Update spam_attendee_list_solicitation.yml

morriscode:
Massive revamp of this rule, tighter matching, removal of titles and matching on threads


morriscode:
/mql-mimic-exempt: 482456,621421,373655,482489,283703,482427,456956,283707,283698,628075,482462,628079,505784,283719,414992,482487,482472,283716

zoomequipd (approved):

PR #2029, opened 2024-10-18
Create link_adobe_share_suspicious.yml

zoomequipd:
# Description

Add additional coverage to malicious (with higher confidence) Adobe shares, regardless of sender relationship to the recipient domain

# Associated samples

Samples shared privately


zoomequipd:
ran hunts in several environments, no hits. ready for review

morriscode (approved):

PR #2047, opened 2024-10-24
Update attachment_adobe_image_lure_qr_code.yml

aidenmitchell:
# Description
Adding Adobe keywords in case LogoDetect fails.

# Associated samples

- https://platform.sublime.security/messages/b6a3c3800382728d042c5c2ca7b43b786a6eb1f06877ab0e91d44525d7fd4777

### DIFF
--- a/detection-rules/attachment_adobe_image_lure_qr_code.yml
+++ b/detection-rules/attachment_adobe_image_lure_qr_code.yml
@@ -15,6 +15,9 @@ source: |
           or any(ml.logo_detect(beta.message_screenshot()).brands,
                  .name == "Adobe"
           )
+          or any(file.explode(.),
+               any(.scan.strings.strings, regex.icontains(., "adobe (acrobat|sign)"))
+          )
         )
     )
     or any(attachments,


detection-rules/attachment_adobe_image_lure_qr_code.yml
```diff
@@ -15,6 +15,9 @@ source: |
           or any(ml.logo_detect(beta.message_screenshot()).brands,
                  .name == "Adobe"
           )
+          or any(file.explode(.),
```
zoomequipd, line 18: Given it's against images or pdfs, do you need leverage the OCR output?

aidenmitchell: We should be ok here, OCR output of the root attachment becomes a deeper `text/plain` object in the FileExplode output that has the OCR output in `.scan.strings.strings`

aidenmitchell: So I guess, this _is_ using the OCR output, just not directly.

zoomequipd (approved):

PR #2061, opened 2024-10-25
Update credential_phishing_esign_document_notification.yml

aidenmitchell:
# Description

Adding OR'd check for HR impersonation.

# Associated samples

- https://platform.sublime.security/messages/afa7340701d8385231da439eba6c746fc46a171f30d537e12aa4ded237bab615

### DIFF
--- a/detection-rules/credential_phishing_esign_document_notification.yml
+++ b/detection-rules/credential_phishing_esign_document_notification.yml
@@ -50,7 +50,7 @@ source: |
           )
   )
   and (
-    // unusal repeated patterns in HTML 
+    // unusual repeated patterns in HTML 
     regex.icontains(body.html.raw, '((<br\s*/?>\s*){20,}|\n{20,})')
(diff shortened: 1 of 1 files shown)


detection-rules/credential_phishing_esign_document_notification.yml
```diff
@@ -67,6 +67,8 @@ source: |
     or regex.icontains(body.html.raw, '>Docus[1l]gn<')
     // common greetings via email.local_part
     or any(recipients.to, strings.icontains(body.current_thread.text, .email.local_part))
+    // HR impersonation
+    or strings.ilike(sender.display_name, "H?R", "*Human Resources*")
```
zoomequipd, line 71: `H?R` within strings.ilike _requires_ a character at between `H` and `R`.  Might be worth including `HR` has it's own string to make sure it matches without a separating character
```suggestion
    or strings.ilike(sender.display_name, "HR", "H?R", "*Human Resources*")
```

aidenmitchell: Whoops, good call, added

zoomequipd (approved):

PR #2090, opened 2024-11-05
standardize docusign auth check

zoomequipd:
# Description

standardize auth check for docusign - source : https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/callback_phishing_docusign_comment.yml#L14

### DIFF
--- a/detection-rules/callback_phishing_docusign_comment.yml
+++ b/detection-rules/callback_phishing_docusign_comment.yml
@@ -8,11 +8,20 @@ source: |
   and length(attachments) == 0
   
   // Legitimate Docusign sending infratructure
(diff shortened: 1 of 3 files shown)


zoomequipd:
This looses up the docusign auth detection

aidenmitchell:
Looks like `docusign.com` was pulled out of one rule, and is missing from the other 2. Is that intentional?

morriscode (dismissed):

zoomequipd:
> Looks like `docusign.com` was pulled out of one rule, and is missing from the other 2. Is that intentional?

Yeah, docusign.com appears to be used more for corporate communications than product communications.

morriscode (approved):

PR #2122, opened 2024-11-14
Update impersonation_ledger.yml

zoomequipd:
# Description
avoid alerting on ledger actual domains if they pass dmarc

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/03d6c1038fefa1f4496b99be86d033ea28a458528ddbfbfeb3305f6257505940)

### DIFF
--- a/detection-rules/impersonation_ledger.yml
+++ b/detection-rules/impersonation_ledger.yml
@@ -19,7 +19,7 @@ source: |
     )
     or (
       (
-        // only match ledger actual domains if dmarc filas
+        // only match ledger actual domains if dmarc fails
         not (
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_ledger.yml
```diff
@@ -18,7 +18,14 @@ source: |
       )
     )
     or (
-      sender.email.domain.root_domain != 'ledger.com'
+      (
+        // only match ledger actual domains if dmarc filas
```
aidenmitchell, line 22:
```suggestion
        // only match ledger actual domains if dmarc fails
```

PR #2160, opened 2024-11-22

zoomequipd:
# Description

remove solicited logic to avoid FNs, results will be compared to #2161 

# Associated samples
N/A
## Associated hunts
N/A


zoomequipd:
results of this will be compared to #2161

PR #2177, opened 2024-12-02
Create link_sharepoint_attached_eml.yml

zoomequipd:
# Description

Add coverage for suspicious SharePoint links in attached EMLs

## Associated hunts

- [Hunt 1](https://platform.sublime.security/hunts/01936ee9-a1b1-74a5-b3bb-d14ea7659cdf)


zoomequipd:
in testing, this rule did not produce unique TP hits and was rather noisy with FPs

PR #2186, opened 2024-12-04
Create abuse_google_drive_new_sender_domain.yml

zoomequipd:
# Description

Add ASR coverage for Google Drive share notifications from newly registered reply-to addresses

## Associated hunts

- [Hunt 1](https://platform.sublime.security/hunts/01938fd0-e53e-7c78-8992-512d72ceb042)


zoomequipd:
seems low volume, no telemetry data.

morriscode (approved):

PR #2218, opened 2024-12-13
Update abuse_dropbox_sus_names.yml

zoomequipd:
# Description
add  untitled.paper

# Associated Hunt
https://platform.sublime.security/hunts/0193be2a-2d7a-73da-8dac-e850c5db1180


morriscode (approved):
brilliant

PR #2251, opened 2024-12-27
Create link_azurecomm.yml

zoomequipd:
# Description

add ASR coverage for abused Azure Communications

## Associated hunts

- [Hunt 1](https://platform.sublime.security/hunts/01940950-6c6a-7bcf-8c5a-aaaf5ae220b7)


zoomequipd:
i think a url decoder is the way to go here.

PR #2255, opened 2024-12-31
Create impersonation_chrome_web_store_policy.yml

zoomequipd:
# Description
Add specific coverage for observed patterns related to Chrome Web Store phishing


zoomequipd:
low volume, tuned based on observed FP.  It's still a pretty wide net here, thus the change to severity low until we see more feedback.

morriscode (approved):

PR #2260, opened 2025-01-03
Update link_multistage_docusign.yml

morriscode:
Adding org_domains to the filter

# Description

Adding org_domains to the filter

# Associated samples

-[ Sample 1](https://platform.sublime.security/messages/37943ec9f54ffcdcdc27d996ee74685ba71aa6bc8982e219ff509157d65d2564)

-[Sample 2](https://platform.sublime.security/messages/b4f77716b71e92d326d0fd1fd1b1284662bf0a0ecb87c996b0729c162687aa5e)


morriscode:
Considering adding high_trust domains to the same filter block, all the FP's I've found are either org_domains or high_trust domains.

zoomequipd (approved):

PR #2266, opened 2025-01-06
Update impersonation_amazon.yml

peterdj45:
# Description

Adding a domain.domain negation for Amazon SLDs

# Associated samples

- https://na-west.platform.sublime.security/messages/76b7b298b9069871df5ea86682a693e384b1fc59cc3deb2b00a6032ae2bcbfe8


zoomequipd:
Interesting edge case with `amazon.com.be` in which `com.be` is actually owned by [amazon](https://www.dnsbelgium.be/en/whois/info/com.be/details). 

We likely don't need to move all of them to use `domain.domain` but there is no harm in doing so.

PR #2332, opened 2025-01-24

peterdj45:
# Description

negating links to the sender's root website when they use a freemail address to deliver invoices

# Associated samples

- https://platform.sublime.security/messages/55cddf15508705ff9cf90500a8abe90f720ed88a40a351a70048c74892de08ce

## Associated hunts

- https://platform.sublime.security/hunts/01949679-c3e7-7f09-b38c-f62bbbbe7f35

### DIFF
--- a/detection-rules/impersonation_quickbooks.yml
+++ b/detection-rules/impersonation_quickbooks.yml
@@ -56,6 +56,12 @@ source: |
                     .href_url.domain.domain == sender.email.domain.domain
                     and (.href_url.path is null or .href_url.path == "/")
                   )
+                  // handle links to the root website when the sender uses a freemail address to send invoices
+                  or (
+                    .href_url.domain.sld == sender.email.local_part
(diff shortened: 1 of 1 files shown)


peterdj45:
not sure if this is the best way to go about resolving this FP...seems like an edge case too. definitively open to ideas here

zoomequipd:
Just so I'm clear, is this because the sender email address, which obviously includes the freemail domain, is somewhere in the body and it gets picked up as a link?

peterdj45:
> Just so I'm clear, is this because the sender email address, which obviously includes the freemail domain, is somewhere in the body and it gets picked up as a link?

It's actually a link to the company's actual website (xyzcorp.com) in the body which is firing since they use a freemail address (xyzcorp@gmail.com) to send invoices to clients.

zoomequipd:
ahhh, I see, thank you for that clarification.

zoomequipd (approved):

PR #2335, opened 2025-01-24
Create abuse_google_shortlinks_services.yml

zoomequipd:
# Description

Add coverage for abuse of subdomains of `g.co` to send messages from google which lend credibility to actors.

### DIFF
--- a/detection-rules/abuse_google_shortlinks_services.yml
+++ b/detection-rules/abuse_google_shortlinks_services.yml
@@ -1,0 +1,29 @@
+name: "Google Services Using G.co Shortlinks"
+description: "Identifies messages from authenticated Google domains containing g.co shortened URLs with a subdomain in either the message body links or thread text."
+references: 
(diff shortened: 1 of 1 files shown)


zoomequipd:
No telemetry data available. 

I'm still fine getting this rule deployed. If it does match a message, it's very likely abuse of some sort. Though if we see FPs for some reason, I wouldn't be afraid to disable the rule outright based on Google taking action to ensure the path of abuse is eliminated.

https://www.theregister.com/2025/01/27/google_confirms_action_taken_to/

> A Google spokesperson told The Register: "We've suspended the account behind this scam, which abused an unverified Workspace account to send these misleading emails. 
> 
> "We have not seen evidence that this is a wide-scale tactic, but we are hardening our defenses against abusers leveraging g.co references at sign-up to further protect users."

morriscode (approved):

PR #2336, opened 2025-01-24
Update impersonation_sharefile.yml

zoomequipd:
# Description
Add new coverage for new lure 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/0c066528bc79e61279eee7199de674b228b10b1a02ae074f09143cba6dfda3eb)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/hunts/01949a79-b140-79ac-9b2e-1d5000a490c5/)


zoomequipd:
telemetry looking good, not many net new samples based on updated logic, but AS distribution is great.

PR #2341, opened 2025-01-26
Create open_redirect_predictiveresponse.yml

zoomequipd:
# Description
Add coverage for observed open redirect

This does appear to be used legit as well, but the negation on header.domains and the list-unsub header addressed the observed benign use. 
## Associated hunts

- [Hunt 1](https://platform.sublime.security/hunts/0194a4a0-afe6-7020-a487-655026e72f0a)


zoomequipd:
telemetry data looks good. ready for review

morriscode (approved):

PR #2391, opened 2025-02-07

aslefhewqiwbepqwefbpqsciwueh:
Suggesting to remove the word report due to a number of false positives.

# Description

Just a very small change to see about removing the word "report" from the regex in the suspicious subjects check. This has caused a few false positives for us. 

Many thanks for considering.


aslefhewqiwbepqwefbpqsciwueh:
Thanks for checking! 🙏 I'll fork the rule.

PR #2393, opened 2025-02-07
Update attachment_svg_embedded_js.yml

morriscode:
# Description
bumping severity to high for tony


tonysleva:
For documentation purposes, Tony was asking if there were cases where an unsolicited email that contained a SVG with embedded JS wouldn't be dangerous.

aidenmitchell (approved):

PR #2407, opened 2025-02-13
Update impersonation_schwab.yml

robino2020:
Added schwab-marketing.com that looks similar but is owned for +24 years by a German Marketing Agency.


zoomequipd (approved):

zoomequipd:
Thanks, this is great!

PR #2412, opened 2025-02-19
Update link_secure_sharepoint_file.yml

peterdj45:
# Description

Adding additional logic to catch unsolicited reply-to addresses

# Associated samples

- https://platform.sublime.security/messages/5fb05fab86319cdbeb23ed353c10f6816793d5dd7428948a219701fc9d3cbdf5


zoomequipd:
might consider reusing this logic? 
https://github.com/sublime-security/sublime-rules/blob/main/detection-rules/abuse_docusign_unsolicited_reply-to.yml#L36C1-L49C4

PR #2429, opened 2025-02-24
Update attachment_adobe_image_lure_qr_code.yml

zoomequipd:
# Description

1. add scan_base64 logic and fix logic error related to undisclosed recipients
1. simplified implementation of #2417 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/4d266508762f50d4a780a1b9da4c3b24eb9a5f3ba2301b4496fe62690560a6de?preview_id=0195235b-1118-78f2-8f6e-97d296e50fa8)

PR #2430, opened 2025-02-24
Update attachment_adobe_image_lure_qr_code.yml

zoomequipd:
# Description

change logic to require a QR code or a mention of QR code along with other suspicious indicators

> [!WARNING]
> This should only be placed in test-rules after #2429 , a full review of test-rules results should be conducted before being merged


zoomequipd:
results in FNs

PR #2433, opened 2025-02-25
Upgrade all GHA from Ubuntu 20.04 to latest.

ab:
https://www.notion.so/sublimesecurity/GitHub-Actions-Ubuntu-20-04-sunset-1a004655fc9d8018b776c64b47f12cf1?pvs=4

### DIFF
--- a/.github/workflows/clear-old-test-rules.yml
+++ b/.github/workflows/clear-old-test-rules.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   remove-stale:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     permissions:
       contents: write
       pull-requests: read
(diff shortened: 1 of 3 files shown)


aidenmitchell (commented):
`latest` is probably fine, but do we want to pin it to `24.04` just in case?

ab:
Yeah I was unsure about that. My pro/con thinking was:

Pros of `latest`:
- No longer need to update this every 2 years
- If we don't have anything that depends on a specific Ubuntu release, this will be simpler and easier
- Will automatically use a more up-to-date OS image sooner

Pros of `24.04`:
- If we *do* have things that depend on a specific Ubuntu release, we may be surprised with breakages when `latest` starts pointing to the new release (GitHub does gradual rollouts of this over a month or two)
- More stable runner env, more explicit about what OS version is in use

Given that: 1. we've gone with `latest` on all the other repos, and 2. I haven't seen anything in the other repos break with this upgrade (which suggests that we don't have anything major pinned to the specific Ubuntu release); I'm inclined to keep it with `latest` here as well. What do you think?

PR #2453, opened 2025-03-03
Add detection for Google Drive auto-download links

aidenmitchell:
## Summary
- Added detection for Google Drive direct download links using the `export=download` parameter
- Enhanced existing `link_download_suspicious_file.yml` rule to detect Google Drive auto-downloads of archives
- Created a dedicated rule `link_google_drive_direct_download_unsolicited.yml` for broader detection
- Both approaches target malicious distribution of files via Google Drive's direct download mechanism

## Test plan
- Tested against sample fc601411c8e1d83df90d63c5af128c89392112a0dfa1af45355476e516e4a133
- Validated rule syntax and confirmed detection

### DIFF
--- a/detection-rules/link_download_suspicious_file.yml
+++ b/detection-rules/link_download_suspicious_file.yml
@@ -21,7 +21,7 @@ source: |
             strings.icontains(.href_url.url, "drive.google.com/uc") 
             and strings.icontains(.href_url.url, "export=download")
             and any(ml.link_analysis(., mode="aggressive").files_downloaded,
-                .file_extension in ("zip", "rar", "7z")  // Archive files
+                .file_extension in $file_extensions_common_archives
             )
           )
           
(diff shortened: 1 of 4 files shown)


detection-rules/link_download_suspicious_file.yml
```diff
@@ -14,7 +16,16 @@ severity: "medium"
 source: |
   type.inbound
   and any(body.links,
-          any(ml.link_analysis(.).files_downloaded,
+          // Detect suspicious direct Google Drive downloads
+          (
+            strings.icontains(.href_url.url, "drive.google.com/uc") 
+            and strings.icontains(.href_url.url, "export=download")
+            and any(ml.link_analysis(., mode="aggressive").files_downloaded,
+                .file_extension in $file_extensions_common_archives
```
zoomequipd, line 24: Do you want to add any file.explode to that downloaded archive to inspect the file extension of the filed within the arcive?

aidenmitchell: Unless we see FPs, not really. A direct download link to an archive is weird enough IMO

morriscode (approved):

PR #2458, opened 2025-03-05
Update suspicious_request_for_quote_or_purchase.yml

aidenmitchell:
# Description

Fixing logic error where a sample with no RFP/RFQ language could flag.
Instead of simply requiring 2 of any indicators, this change ensures that:
1. At least one specific RFQ/RFP language pattern must be present
2. At least two indicators total must be present
3. If at least two language patterns are present, the rule will fire
4. Non-language indicators alone won't trigger the rule

This maintains detection effectiveness while preventing false positives from non-RFQ/RFP content.

Screenshot 2025-03-05 at 12 47 59 PM

# Associated samples
- https://platform.sublime.security/messages/25f70ab736f292245ba505e615986da0a67ff1fa0cabf69d448e62dee4afce5d?preview_id=01953dc2-f213-7eef-8998-e90f6612bf27

### DIFF
--- a/detection-rules/suspicious_request_for_quote_or_purchase.yml
+++ b/detection-rules/suspicious_request_for_quote_or_purchase.yml
@@ -37,60 +37,52 @@ source: |
     )
   )
   and (
-    3 of (
-      (
-        regex.icontains(body.current_thread.text,
-                        '(discuss.{0,15}purchas(e|ing))'
+    // Group the language patterns that specifically indicate RFQ/RFP
+    (
+      1 of (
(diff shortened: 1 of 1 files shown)


aidenmitchell:
/mql-mimic-exempt 458920,464777,608919,608923,608946,683538,683760,701538,709546,754842,754843,754844,754870,771789,774416,774422,774677,774678,774681,784403,804218,825275,846598,909441,918182,926136,935515,946518,957771,957869,957936,964086

None of these are RFP related.

zoomequipd (approved):

PR #2468, opened 2025-03-07
Update impersonation_linkedin.yml

peterdj45:
# Description

Negating DocuSign notifications originating from linkedin.com reply-to addresses

# Associated samples

- https://platform.sublime.security/messages/b84451f11ad89dd9ca41fe65b87340e308d00d33c95801c678344133cd4bb68a?preview_id=019571f1-f45b-75fc-935b-fcce30e818be


zoomequipd:
Do you think we should just add a high_trust_sender_root_domain negation?

morriscode (approved):

PR #2529, opened 2025-03-25
Allow a single rule to be modified by multiple PRs in test-rules

zoomequipd:
# Description

Update the `update-test-rules` GH action to allow the same file to be modified by multiple PRs. 

This is accomplished by the following changes, which are specific to the test-rules branch:
1) Prepending the PR number to the output filename
2) Replacing the `id` calculated by `generate-rule-ids/main.py` (and added by the `[Run Rule Validation]`(https://github.com/sublime-security/sublime-rules/blob/main/.github/workflows/rule-validate.yml#L92C15-L97)` step) with a deterministic UUID based on the updated filename. 

The existing `Remove Stale from test-rules Branch` action keys in on the `testing_pr` key within the file which continue to function as it currently does. 

# Testing
This has been tested here within a fork's branch via https://github.com/zoomequipd/sublime-rules/pull/2.  You can observe the [action](https://github.com/zoomequipd/sublime-rules/actions/runs/14071099381/job/39405267516?pr=2) and the associated [commit](https://github.com/zoomequipd/sublime-rules/commit/891f9dbbde45151cd631a53081b5728baa5f81e1) to the test-rules branch. 

The cleanup GH action [could not be tested](https://github.com/zoomequipd/sublime-rules/actions/runs/14071893491/job/39407672046#step:4:49) due to logic preventing it from running on forked repos. 
However, the cleanup action is unchanged.

### DIFF
--- a/.github/workflows/update-test-rules.yml
+++ b/.github/workflows/update-test-rules.yml
@@ -93,6 +93,22 @@ jobs:
         run: |
           export pr_num=${{ github.event.issue.number }}
           
+          # Function to generate a deterministic UUID using the DNS namespace and a seed string
+          # This simulates the Python uuid.uuid5() function using native tools
+          generate_uuid() {
+            # Use the standard DNS namespace UUID as defined in RFC 4122
+            local DNS_NAMESPACE="6ba7b810-9dad-11d1-80b4-00c04fd430c8"
+            local seed="$1"
+            
+            # Combine namespace and seed (similar to how Python's uuid5 implementation works)
+            # Generate a deterministic hash using openssl (similar concept to UUID5)
+            local hash=$(echo -n "$DNS_NAMESPACE$seed" | openssl dgst -sha1 | awk '{print $2}')
+            
+            # Format as UUID version 5
+            # First 8 chars, then 4 chars, then 4 chars (with first nibble set to 5 for UUID v5), etc.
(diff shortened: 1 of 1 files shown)


.github/workflows/update-test-rules.yml
```diff
@@ -130,9 +148,33 @@ jobs:
...
+            cp "source/$file" "destination/$dir_path/${prefixed_filename}"
+            
+            # Generate a deterministic UUID based on the prefixed filename
+            # test-rules requires each rule to have a unique, but deterministic
+            # id.  This will be different than the auto generated one with the main branch
+            new_uuid=$(generate_uuid "${prefixed_filename}")
+            echo "Generated UUID for ${prefixed_filename}: ${new_uuid}"
+            
+            # Update the ID in the YAML file using yq
+            yq -i '.id = "'$new_uuid'"' "destination/$dir_path/${prefixed_filename}"
```
cameron-dunn-sublime, line 171: Maybe there's some value in including the original `id` as another field like `original_id`. That would just make investigations a little easier if the name is being changed at the same time.

cameron-dunn-sublime (approved):

.github/workflows/update-test-rules.yml
zoomequipd: Good suggestion! Got this added!

cameron-dunn-sublime (approved):

PR #2554, opened 2025-04-04
Use github event sha to validate branch

cameron-dunn-sublime:
# Description

Prevent an issue where checks run with a later commit on the branch (e.g. a merge from `main`).

I validated by opening a PR against this branch and ensuring that basic operations succeed (adding a rule ID). CI succeeded on that PR, but I did not try and simulate the race condition.

### DIFF
--- a/.github/workflows/rule-validate.yml
+++ b/.github/workflows/rule-validate.yml
@@ -53,7 +53,7 @@ jobs:
           fi
           
           echo "##[set-output name=head_ref;]$head_ref"
-          echo "##[set-output name=repo;]$repo"
+          echo "##[set-output name=repo;]$repo"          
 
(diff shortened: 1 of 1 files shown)


.github/workflows/rule-validate.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           fi
           
           echo "##[set-output name=head_ref;]$head_ref"
-          echo "##[set-output name=repo;]$repo"
+          echo "##[set-output name=repo;]$repo"
```
cameron-dunn-sublime, line 56:
```suggestion
          echo "##[set-output name=repo;]$repo"
```

zoomequipd (approved):

PR #2563, opened 2025-04-08
Create generic_service_abuse_reply_to.yml

aidenmitchell:
# Description

Detects messages from services that write the true sender to the reply-to field, where the sender has no prior legitimate message history and is newly registered. Indicative of service abuse.

# Associated samples

- https://platform.sublime.security/messages/hunt?huntId=01961678-7e78-70a6-8e94-f3cce55d4975

### DIFF
--- a/detection-rules/generic_service_abuse_reply_to.yml
+++ b/detection-rules/generic_service_abuse_reply_to.yml
@@ -24,3 +24,4 @@ detection_methods:
   - "Header analysis"
   - "Sender analysis"
   - "Whois"
+id: "0937b4c5-72d8-5efd-834f-c80ca8336f25"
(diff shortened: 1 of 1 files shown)


detection-rules/generic_service_abuse_reply_to.yml
```diff
@@ -0,0 +1,27 @@
...
+  type.inbound
+  and any(headers.reply_to, network.whois(.email.domain).days_old < 30)
+  and sender.email.domain.domain in $replyto_service_domains
+  // 
+  // This rule makes use of a beta feature and is subject to change without notice
+  // using the beta feature in custom rules is not suggested until it has been formally released
+  // 
+  and not beta.profile.by_reply_to().solicited
+  and not beta.profile.by_reply_to().any_messages_benign
+
```
zoomequipd, line 15: thoughts on using reply_to_email()? 
```suggestion
  // 
  and not beta.profile.by_reply_to_email().solicited
  and not beta.profile.by_reply_to_email().any_messages_benign

```

zoomequipd: Probably not needed here because the logic of the rule is on the domain of the reply_to address.

zoomequipd (approved):

PR #2608, opened 2025-04-16
ta4903_PDF_Phish

brycampbell:
Ta4903 Phish from Proofpoint  https://www.proofpoint.com/uk/blog/threat-insight/ta4903-actor-spoofs-us-government-small-businesses-phishing-bec-bids


ta4903_PDF_Phish
```diff
@@ -0,0 +1,43 @@
+name: "Suspicious Attachment: PDF files"
```
zoomequipd, line 1: I would suggest something more specific than `Suspicious PDF Files`
```suggestion
name: "Attachment: Suspicious PDF files"
```

PR #2626, opened 2025-04-17
Brycampbell patch 2

brycampbell:
# Description

This rule is based on the Proofpoint blog regarding bid spoofing. 
https://www.proofpoint.com/uk/blog/threat-insight/ta4903-actor-spoofs-us-government-small-businesses-phishing-bec-bids


brycampbell:
Reviewed test rules, no telemetry so far.

zoomequipd (changes requested):
everything in "source" should be indented with twice spaces.  There is only one currently

zoomequipd (dismissed):
everything in "source" should be indented with twice spaces.  There is only one currently

aidenmitchell (approved):

PR #2627, opened 2025-04-17
Update paypal_invoice_abuse.yml

morriscode:
Expanding seller note match to include "your note"  dropping 4 of to a 3 of.

# Description

Expanding seller note match to include "your note"  dropping 4 of to a 3 of.
If it's a new rule or insight, explain what the rule is designed to catch/what the insight should fire on.

# Associated samples

- 29a2db991439cd522045605c043ccf50d2e9612908df705b6a12669e664baea4

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019644dc-12e7-7297-a843-4953fbc3583e)


morriscode:
Modified to ensure we aren't flagging on paypals disclaimer

zoomequipd:
thoughts on closing this in favor of https://github.com/sublime-security/sublime-rules/pull/2684?

PR #2644, opened 2025-04-23
Update impersonation_chase.yml

peterdj45:
# Description

adding `Chase Privacy Operations` as a suspicious message body keyword. editing display name negation to handle messages with a null sender display name.

# Associated samples

- https://platform.sublime.security/messages/672d4a4e3634f023965f4b4477b788f9cd1c4ce87b03423acc13df6b2aa2e1fb
- https://platform.sublime.security/messages/d3cd75975a081ffa103b5405fce7c6b8bb9b2b61ff3c1c0e6161d199878e12c4
- https://platform.sublime.security/messages/63684696d207dd8ad2ca7cb9340617c152f8be8a9e987852b49464b1493bf074

## Associated hunts

- https://platform.sublime.security/hunts/0196601d-4324-7cc2-b990-a4f7563ffb72

### DIFF
--- a/detection-rules/impersonation_chase.yml
+++ b/detection-rules/impersonation_chase.yml
@@ -37,6 +37,11 @@ source: |
         )
       )
     )
+    or strings.icontains(body.current_thread.text, 'Chase Privacy Operations')
+  )
+  and not (
+    sender.display_name is not null
+    and sender.display_name in~ ("chaser", "case")
   )
   and sender.display_name not in~ ("chaser", "case")
   and sender.email.domain.root_domain not in~ ('chase.com', 'united.com', 'transunion.com', 'shopping-chase.com')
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_chase.yml
```diff
@@ -37,6 +37,11 @@ source: |
...
       )
     )
+    or strings.icontains(body.current_thread.text, 'Chase Privacy Operations')
+  )
+  and not (
+    sender.display_name is not null
+    and sender.display_name in~ ("chaser", "case")
   )
   and sender.display_name not in~ ("chaser", "case")
   and sender.email.domain.root_domain not in~ ('chase.com', 'united.com', 'transunion.com', 'shopping-chase.com')
```
zoomequipd, line 47: The logic you added is perfect.  When considering any negations, we need to consider if the value can be null, so good work spotting that. But we should be able to collapse it into the existing condition without creating another one. 

```suggestion
    or strings.icontains(body.current_thread.text, 'Chase Privacy Operations')
  )
  and not (
    sender.display_name is not null and sender.display_name in~ ("chaser", "case")
  )
  and sender.email.domain.root_domain not in~ ('chase.com', 'united.com', 'transunion.com', 'shopping-chase.com')
```

morriscode (approved):

PR #2646, opened 2025-04-23
Update hr_impersonation_docusign_comment.yml

morriscode:
# Description

Adding beta reply-to to avoid firing on docusign


lvh (approved):
This fixed a thing for me and we worked on it together on Zoom

zoomequipd (approved):

PR #2655, opened 2025-04-28
Update fake_thread.yml

morriscode:
# Description

Updating to mirror the logic we use in rules to detect fake threads, specifically checking for the presence of email thread headers such as: From:, To:, Sent:, Date:, Cc:, and Subject:.

# Associated samples

- 9724f3af07b3f27677f68b818fd9db0c8cc2a22994383fe1e9150b5600f0e0a7


Rurik:
This is a really cool idea. Thank you!

zoomequipd (approved):

PR #2660, opened 2025-04-29
Create link_screenconnect.yml

aidenmitchell:
# Description
Detects when a link leads to a ConnectWise ScreenConnect installer and references a relay domain that doesn't match sender or organizational domains.

# Associated samples
- https://platform.sublime.security/messages/917b2fa5197a275b4075f93dad31661bd690776c2099c5c2e4fb4c1d22e2d420?preview_id=0196820a-b03b-74bc-9124-735b45032b31

### DIFF
--- a/detection-rules/link_screenconnect.yml
+++ b/detection-rules/link_screenconnect.yml
@@ -1,0 +1,43 @@
+name: "Link: ScreenConnect Installer With Suspicious Relay Domain"
+description: "Detects when a link leads to a ConnectWise ScreenConnect installer and references a relay domain that doesn't match sender or organizational domains."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(body.links,
+          strings.ends_with(.href_url.url, ".exe")
+          and any(ml.link_analysis(., mode="aggressive").files_downloaded,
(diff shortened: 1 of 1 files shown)


detection-rules/link_screenconnect.yml
```diff
@@ -0,0 +1,44 @@
+name: "Link: ScreenConnect Installer With Suspicious Relay Domain"
+description: "Detects when a link leads to a ConnectWise ScreenConnect installer and references a relay domain that doesn't match sender or organizational domains."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and any(body.links,
+          strings.ends_with(.href_url.url, ".exe")
```
brycampbell, line 8: screenconnect also packages `.msi` and `.pkg` files for macOS so might be useful to cover those.

(https://www.screenconnect.com/download/)

aidenmitchell: Agreed, but we haven't seen those come up (yet)
This rule also targets the specific strings in the `.exe`, so we'd need actual `.msi` and `.pkg` files to see how they embed the relay information.

zoomequipd (approved):

PR #2665, opened 2025-04-30
Create asr_new_sender_or_reply_to_with_new_linked_dom.yml

morriscode:
# Description

New ASR Rule combining sender or reply-to newness with a new link domain. No sender profile usage.

# Associated samples

- a0a7bed1e66926b8d240e83daaafae81ff7d62bfa12a6459697dbb8bafe7ebb1

## Associated hunts

- [Hunt 1](https://platform.sublime.security/hunts/019688ab-e5e4-7e3a-bd83-2c7d7bd6db7d)


detection-rules/asr_new_sender_or_reply_to_with_new_linked_dom.yml
```diff
@@ -0,0 +1,32 @@
+name: "New Sender or Reply-To with New Linked Domain"
```
zoomequipd, line 1: "New Sender" is ambiguous here, is it prevalence new or newly registered
```suggestion
name: "Newly Registered Sender or Reply-To Domain with Newly Registered Linked Domain"
```
I don't like this either, but it's less ambiguous

IndiaAce (approved):

PR #2670, opened 2025-05-02
apple_state-sponsored_attack_warning.yml

brycampbell:
This rule notifies Sublime users of advanced threats from Apple about attacks, it relies on logic contains in the email body and ensures the email is from Apple.

# Description

This rule notifies Sublime users of advanced threats from Apple about attacks, it relies on logic contains in the email body and ensures the email is from Apple.
## Associated hunts

https://platform.sublime.security/messages/hunt?huntId=01969205-70fa-7c7d-b5af-a76bc2b125d0

Example email notification:

013Q6uY0XB8UsAYE9WT5ZTG-2 fit_lim size_768x


zoomequipd (changes requested):
rule filename should be in lowercase and end in `.yml`

discovery-rules/Apple_State-Sponsored_Attack_Warning
```diff
@@ -0,0 +1,22 @@
+Rule name
+Apple State-Sponsored Attack Warning
```
zoomequipd, line 2: should be a single line
```suggestion
Rule name: "Apple State-Sponsored Attack Warning"
```

discovery-rules/Apple_State-Sponsored_Attack_Warning
```diff
@@ -0,0 +1,22 @@
...
+name: "Apple State-Sponsored Attack Warning"
+description: "Detects legitimate Apple threat notifications warning users about potential state-sponsored mercenary attacks targeting their iPhone. These notifications contain specific language about targeted attacks and Apple's confidence level in the warning."
+type: "rule"
+severity: "medium"
+source: | https://support.apple.com/en-us/102174
```
zoomequipd, line 7: add references in their own yaml array element
```suggestion
references:
  - "https://support.apple.com/en-us/102174"
source: | 
```

zoomequipd (approved):

PR #2674, opened 2025-05-05
Create sender_freemail_replyto_different_freemail.yml

zoomequipd:
# Description

Add coverage for a freemail sender with a different freemail provider as a reply-to

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=0196a0f1-4352-7173-99da-130010e330a3)


zoomequipd:
ready for review.  converted to ASR, added negations for commonly observed benign uses. has plenty to TP matches

PR #2684, opened 2025-05-08
Update paypal_invoice_abuse.yml

zoomequipd:
# Description

Complete rewrite of the rule to support use of xpath which allows for elements within the paypal template to be inspected individually.  This reduces FP significantly and allows for the removal of subject based "gating"

# Associated samples

- Sample 1
- Sample 2

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=0196adae-e21c-75c6-a7e4-b3f3cdf7b90c)

### DIFF
--- a/detection-rules/paypal_invoice_abuse.yml
+++ b/detection-rules/paypal_invoice_abuse.yml
@@ -17,153 +17,106 @@ source: |
     "paypal.com.ar",
     "paypal.co.uk"
   )
-  and (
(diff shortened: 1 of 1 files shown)


zoomequipd:
/mql-mimic-exempt: 835744, 850953, 941732

These are not malicious

zoomequipd:
looking great

PR #2686, opened 2025-05-08
Update attachment_adobe_image_lure.yml

zoomequipd:
# Description

rewrite this portion of the logic to limit file.explode calling from occurring per brand to per attachment

# Associated samples
N/A


zoomequipd:
testing looks to result in no loss of efficacy.

aidenmitchell (approved):

PR #2689, opened 2025-05-09
Update impersonation_docusign.yml

zoomequipd:
# Description

Test removing of sender profile elements after rework of the rule


zoomequipd:
/mql-mimc-except: 490065

docusign lure is not actionable

zoomequipd:
/mql-mimic-exempt: 490065

docusign lure is not actionable

morriscode (approved):

PR #2699, opened 2025-05-13
remove sender profile from open redirect rules

zoomequipd:
# Description

remove sender profile from open redirect rules


detection-rules/open_redirect_ticketmaster.yml
zoomequipd, line 1: ticketmaster seems to have resolved this

aidenmitchell (approved):

PR #2702, opened 2025-05-14
remove sender profiles from a subset of rules

zoomequipd:
# Description

remove sender profiles from a subset of rules


zoomequipd:
done testing

PR #2711, opened 2025-05-16
Create recipient_sld_html_class.yml

peterdj45:
# Description

Detects when the recipient's domain name is concealed within HTML class attributes. The message comes from either an unauthenticated trusted sender or an untrusted source.

# Associated samples

- https://platform.sublime.security/messages/6e75e06cfd0866e59e8c90b9299424ce099d12ddc9e92a61a9fceebd18d1cddd?preview_id=0196ca67-d7eb-71ce-8ab5-ab3048e5ce0e

## Associated hunts

- https://platform.sublime.security/messages/hunt?huntId=0196da56-2ac0-785a-89e6-d93da6f1980c

### DIFF
--- a/detection-rules/recipient_sld_html_class.yml
+++ b/detection-rules/recipient_sld_html_class.yml
@@ -1,0 +1,31 @@
+name: "Body HTML: Recipient SLD In HTML Class"
+description: "Detects when the recipient's domain name is concealed within HTML class attributes. The message comes from either an unauthenticated trusted sender or an untrusted source."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and any(recipients.to,
+          any(distinct(html.xpath(body.html, '//*[@class]/@class').nodes, .raw),
+              strings.icontains(.raw, ..email.domain.sld)
+          )
+          and .email.domain.root_domain not in $free_email_providers
+          and length(.email.domain.sld) > 4
+  )
+  and (
+    (
+      sender.email.domain.root_domain in $high_trust_sender_root_domains
+      and not headers.auth_summary.dmarc.pass
+    )
+    or sender.email.domain.root_domain not in $high_trust_sender_root_domains
+  )
+  and not profile.by_sender().solicited
+
+attack_types:
+  - "Credential Phishing"
+tactics_and_techniques:
+  - "Evasion"
+  - "Social engineering"
+detection_methods:
+  - "HTML analysis"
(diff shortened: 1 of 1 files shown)


MSAdministrator (approved):
These hunts look good to me.

peterdj45:
> These hunts look good to me.

I found a lot of false positives with this PR in test_rules (messages that have the recipient CC'ed or coming from an org_domain), but my most recent commit should address that. I'll give it a few more days in test_rules just to make sure though :)

zoomequipd:
most recent change solved the FNs, but looks like the activity isn't being observed recently. marking ready for review. 

This hunt had to have the $org_domain logic removed, so some FPs are present, but those would not be present in actual environments due to rule logic. 
[Hunt 1](https://platform.sublime.security/hunts/01996775-eae6-7981-8b34-4b1d2c3adbbc)

detection-rules/recipient_sld_html_class.yml
```diff
@@ -0,0 +1,64 @@
+name: "Body HTML: Recipient SLD In HTML Class"
```
IndiaAce: nit - sentence case this
```suggestion
name: "Body HTML: Recipient SLD in HTML class"
```

detection-rules/recipient_sld_html_class.yml
```diff
@@ -0,0 +1,64 @@
...
+description: "Detects when there is a a single recipient within $org_domains where the domain SLD is concealed within HTML class attributes. The message comes from either an unauthenticated trusted sender or an untrusted source."
```
IndiaAce:
```suggestion
description: "Detects when there is a single recipient within $org_domains where the domain SLD is concealed within HTML class attributes. The message comes from either an unauthenticated trusted sender or an untrusted source."
```
nit

detection-rules/recipient_sld_html_class.yml
```diff
@@ -0,0 +1,64 @@
...
+    and coalesce(headers.auth_summary.dmarc.pass, false)
+  )
+  
+  // a single recipient within the org_domains
+  and (
+    length(recipients.to) == 1
+    and all(recipients.to, .email.domain.domain in $org_domains)
+  )
+  
+  // there are more than 30 class attributes containinging the recipient's SLD
```
IndiaAce:
```suggestion
  // there are more than 30 class attributes containing the recipient's SLD
```
nit

PR #2712, opened 2025-05-16
Create open_redirect_asemailmgmteu.yml

peterdj45:
# Description

Message contains use of the asemailmgmteu.com open redirect. This has been exploited in the wild.

# Associated samples

- https://platform.sublime.security/messages/a9d38f9c3be006f2d18557761956a9dd48acc4304aee0864c3c9306d22e6f54a


detection-rules/open_redirect_asemailmgmteu.yml
```diff
@@ -0,0 +1,41 @@
...
+                                  'redirect=[^&]*asemailmgmteu\.com'
+          )
+  )
+  and (
+    not profile.by_sender().solicited
+    or (
+      profile.by_sender().any_messages_malicious_or_spam
+      and not profile.by_sender().any_false_positives
+    )
+  )
```
zoomequipd, line 21: trying to avoid using sender profile in open redirect rules
```suggestion
```

peterdj45: whoops, I think I copied an older open redirect template. will keep that in mind!

peterdj45:
the redirect is used legitimately in certain corporate newsletters, which is causing tons of FPs. however there only seems to be a couple of sender domains responsible for those, so a negation should do (they send tons of graymail so I don't think we should add those to high trust).

detection-rules/open_redirect_asemailmgmteu.yml
```diff
@@ -0,0 +1,38 @@
...
+          and not regex.icontains(.href_url.query_params,
+                                  'redirect=[^&]*asemailmgmteu\.com'
+          )
+  )
+  // negate legitimate use of the redirect in newsletters
+  and sender.email.domain.root_domain not in~ (
+      "wizznews.com",
+      "tomtom.com"
+  )
+
```
zoomequipd: we can make use of the stirngs.parse_url and query_params_decoded to better inspect the redirect element, and limit the total body links as well. 

https://platform.sublime.security/messages/hunt?huntId=0198f925-88fb-773a-b758-34bceae732a6
hunt looks great

```suggestion
  type.inbound
  and 0 < length(body.links) < 10
  and any(body.links,
          .href_url.domain.root_domain == "asemailmgmteu.com"
          and strings.icontains(.href_url.query_params, 'redirect=')
          // negate use of the redirect by asemailmgmteu.com
          and not any(.href_url.query_params_decoded["redirect"],
                      strings.parse_url(.).domain.root_domain == sender.email.domain.root_domain
                      and not strings.parse_url(.).domain.root_domain == "asemailmgmteu.com"
          )
  )
```

zoomequipd (dismissed):
looks like this method is still in use, minor tweaks should get this out the door.

detection-rules/open_redirect_asemailmgmteu.yml
peterdj45: looks good to me!

detection-rules/open_redirect_asemailmgmteu.yml
```diff
@@ -0,0 +1,34 @@
...
+description: "Message contains use of the asemailmgmteu.com open redirect. This has been exploited in the wild."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and 0 < length(body.links) < 10
+  and any(body.links,
+          .href_url.domain.root_domain == "asemailmgmteu.com"
+          and strings.icontains(.href_url.query_params, 'redirect=')
+          // negate use of the redirect by asemailmgmteu.com
```
MSAdministrator, line 12:
```suggestion
           and .display_text != "unsubscribe"
          // negate use of the redirect by asemailmgmteu.com
```

MSAdministrator (changes requested):

MSAdministrator:
Updated [Hunt](https://platform.sublime.security/messages/hunt?huntId=01994e54-9919-72c0-ac8b-1f42c89c8cfd)

MSAdministrator (approved):

PR #2722, opened 2025-05-20
Create encrypted_msaccess_database.yar

brycampbell:
This rule covers detection for the following file, AES-256-CBC with HMAC-SHA-256 VBScript implementation inside `accdb` file types. This YARA rule will be called from within an upcoming rule i have that relies on YARA to detect the logic contained in the `encrypt` & `decrypt` functions, the additional YARA coverage for Crypto library fragments are also required.

### DIFF
--- a/detection-rules/attachment_encrypted_msft_access_database.yml
+++ b/detection-rules/attachment_encrypted_msft_access_database.yml
@@ -13,7 +13,7 @@ source: |
             or .file_type == "msaccess"
           )
           and any(file.explode(.),
-                  any(.flavors.yara, . == 'encrypted_msaccess_database')
+                  any(.scan.yara.matches, .name == 'encrypted_msaccess_database')
           )
   )
   // Negating high-trust sender domains unless they fail DMARC authentication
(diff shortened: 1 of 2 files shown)


zoomequipd (approved):

detection-rules/attachment_encrypted_msft_access_database.yml
```diff
@@ -0,0 +1,39 @@
...
+  and any(attachments,
+          (
+            strings.iends_with(.file_name, ".accdb")
+            or .content_type == "application/msaccess"
+            or .content_type == "application/x-msaccess"
+            or .content_type == "application/vnd.ms-access"
+            or .file_type == "msaccess"
+          )
+          and any(file.explode(.),
+                  any(.flavors.yara, . == 'encrypted_msaccess_database')
```
zoomequipd, line 16:
```suggestion
                  any(.scan.yara.matches, .name == 'encrypted_msaccess_database')
```

zoomequipd (approved):

PR #2733, opened 2025-05-22
Mark VIP rules as strict, and create new near-match rules

aidenmitchell:
# Description

Adding a new VIP impersonation rule using `ilevenshtein`, and renaming the existing rule to note that it is a strict match.


zoomequipd (approved):
i like that you made this one an ASR rule

PR #2744, opened 2025-05-27
Update govdelivery_compromise.yml

zoomequipd:
# Description

handle cases where govdelivery.com isn't the sender but is the delivery infra and where dmarc is missing.

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/82c5912f8bc011dd0c8957954daafb7342b491e43c46720e68afd42e200eadcf?preview_id=01970f27-f639-742f-8c37-d430d1774887)


zoomequipd:
rule uses LA, not in test-rules,

aidenmitchell (approved):

PR #2748, opened 2025-05-28
Create open_redirect_bangkoksync.yml

brycampbell:
# Description

This rule is designed to detect the open redirect abuse used to forward to some observed click-fix domains.

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=0197172b-7232-7701-aa59-db0399170f80)


brycampbell:
Low volume 100% true positive results.

PR #2766, opened 2025-06-02
update test-rules process 

zoomequipd:


.github/workflows/update-test-rules.yml
```diff
@@ -1,198 +1,151 @@
...
+      uses: actions/checkout@v4
 
-      - name: Fail if not organization member
-        if: |
-          steps.is_organization_member.outputs.result != 'true'
-        run: exit 1
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.x
```
MSAdministrator: I think we should add this value as a env var up top and just reference anywhere we need. Makes it easier when changing the version in multiple places.

.github/workflows/update-test-rules.yml
```diff
@@ -1,198 +1,151 @@
...
+        # Disable adding tags that aren't useful
+        ADD_RULE_STATUS_TAG: 'false'
+        ADD_PR_REFERENCE: 'false'
+        INCLUDE_PR_IN_NAME: 'false'
+        ADD_AUTHOR_TAG: 'false'
+        CREATE_OPEN_PR_TAG: 'false'
 
-      - name: Install yq
-        uses: mikefarah/yq@v4.33.3
+      run: python scripts/download_detection_rules.py
```
MSAdministrator: Did you mean `sync_detection_rules.py` here ?

scripts/sync_detection_rules.py
```diff
@@ -0,0 +1,866 @@
...
+    # Create a namespace UUID (using the DNS namespace as a standard practice)
+    namespace = uuid.UUID('6ba7b810-9dad-11d1-80b4-00c04fd430c8')
+
+    # Create a UUID using the namespace and the seed string
+    deterministic_uuid = uuid.uuid5(namespace, seed_string)
+
+    return str(deterministic_uuid)
+
+
+def add_id_to_yaml(content, filename):
```
MSAdministrator, line 268: I'm curious why not just load the yaml using `pyyaml` and then update and write the new yaml file back to disk instead of using regex? I'm sure you have a reason just wondering.

MSAdministrator (commented):
Some initial questions

scripts/sync_detection_rules.py
zoomequipd: pyyaml doesn't output the source in the same format and it becomes very messy

.github/workflows/update-test-rules.yml
zoomequipd: I sure did

.github/workflows/update-test-rules.yml
zoomequipd: I'm not fully understanding what you're suggestion is there, any chance you could make the proposed change?

scripts/sync_detection_rules.py
```diff
@@ -0,0 +1,866 @@
...
+# Common configuration
+GITHUB_TOKEN = os.getenv('GITHUB_TOKEN')
+SUBLIME_API_TOKEN = os.getenv('SUBLIME_API_TOKEN')
+REPO_OWNER = os.getenv('REPO_OWNER', 'sublime-security')
+REPO_NAME = os.getenv('REPO_NAME', 'sublime-rules')
+OUTPUT_FOLDER = os.getenv('OUTPUT_FOLDER', 'detection-rules')
+
+# Script mode selection (default to 'standard' if not specified)
+# Possible values: 'standard', 'test-rules'
+SCRIPT_MODE = os.getenv('SCRIPT_MODE', 'standard')
```
cameron-dunn-sublime, line 19: What's the purpose/use case for `standard`?

zoomequipd: It creates the `Rules from open PRs` feed used by shared samples, which follows the same general workflow, but has different requirements.  see https://github.com/zoomequipd/sublime-rule-prs/blob/main/.github/workflows/sync_detection_rules.yml

cameron-dunn-sublime (approved):

.github/workflows/update-test-rules.yml
```diff
@@ -1,198 +1,152 @@
...
-      - name: Get PR branch
-        uses: alessbell/pull-request-comment-branch@ef3408c9757d05f89cb525036383033a313758a0  # v2.1.0
-        id: comment_branch
+    - name: Get latest script from main branch
+      run: |
+        # Checkout the scripts directory from main branch without switching branches
+        git checkout origin/main -- scripts/
+        
+        # Unstage the script changes - we only want to commit detection rule changes
+        git reset HEAD scripts/ || true
```
rw-access, line 45: Do you actually want to check out main or do you want to _merge_ main for this file? The reason for the latter is that it means you can choose to edit the script in a branch and use it before it's merged, but with this approach, you have to wait til it's merged.

Given the committed changes and all, you may not want to do exactly that, but it might be worth supporting the case of editing the script in a branch and using it. Plenty of other ways to do that with git

zoomequipd: Yup, I had considered this "issue" but the easiest thing to do is fork and test in your own repo.   This allows ones to test the same workflow file as well (otherwise even a "test" branch would be writing to the "prod" test-rules branch)

Will be something that gets addressed in a later phase.

PR #2783, opened 2025-06-04
Adding social landing check where URL shortener checks exist

aidenmitchell:
# Description

We have started to see abuse of social landing pages (link-in-bio) services, in a similar manner to URL shorteners. This PR adda a check for those services where we have been checking against the URL shortener list.


detection-rules/brand_impersonation_ms_planner.yml
```diff
@@ -12,6 +12,7 @@ source: |
               or .href_url.domain.root_domain in $free_file_hosts
               or .href_url.domain.root_domain in $free_subdomain_hosts
               or .href_url.domain.domain in $url_shorteners
+              or .href_url.domain.domain in $social_landing_hosts
```
MSAdministrator, line 15: Should this also include ?

`or .href_url.domain.root_domain in $social_landing_hosts`

aidenmitchell: Yep, good catch!

PR #2784, opened 2025-06-04
Update attachment_callback_phish_with_pdf.yml

aidenmitchell:
# Description

Negating Quickbooks invoice PDFs
# Associated samples
- https://platform.sublime.security/messages/6d43ca45d8ea7cdcfbd01bff1f62acad0b0fd13531e443e6326676758d5a633a?preview_id=0196cfef-eb3b-75e1-bfab-e5d810be62e5


MSAdministrator:
Just a question regarding why call out the speciific creator as I would be suprised that this wouldn't change occasionally? or there may be another reason as well?

aidenmitchell:
I ran a hunt with it and it seems that QB invoices always have that creator, I figured the combo of producer and creator should be good enough to avoid negating other attachments.

PR #2807, opened 2025-06-09
Update impersonation_google_drive_file_share

andrew-kline:
# Description

Adding a flexible "shared a * with you" to the strings ilike conditions. We've had phishing emails come in that this rule should've fired on but the ilike conditions didn't include this condition.


brycampbell:
[Hunt-1](https://platform.sublime.security/messages/hunt?huntId=01977d2f-4130-7618-9301-f24b63056d75)

PR #2808, opened 2025-06-09
Checking current thread text mentions follow up

MSAdministrator:
# Description

Extending rule to check if mentioning if a follow up from a previous email / thread. 

# Associated samples

- [Sample 1](https://na-east-2.platform.sublime.security/messages/31c655cc53af7b7c920b84af0a92d41cdc299849f98a8f450c00a2e13dd3f7ca?preview_id=0196fce7-b1f5-7052-87f5-e8cecc20f0a1)
- Sample 2

## Associated hunts

- Hunt 1

### DIFF
--- a/detection-rules/spam_attendee_list_solicitation.yml
+++ b/detection-rules/spam_attendee_list_solicitation.yml
@@ -39,6 +39,7 @@ source: |
       and not regex.icontains(body.current_thread.text,
                               "(?:debit card|transaction.{0,20}processed|receipt)"
       )
+      and regex.icontains(body.current_thread.text, 'follow(?:ing)?(-| )up')
     )
     // if there are indicators of a previous thread, also inspect the previous thread
     or (


detection-rules/spam_attendee_list_solicitation.yml
```diff
@@ -39,6 +39,7 @@ source: |
       and not regex.icontains(body.current_thread.text,
                               "(?:debit card|transaction.{0,20}processed|receipt)"
       )
+      and regex.icontains(body.current_thread.text, 'follow(?:ing)?(-| )up')
```
IndiaAce: It looks like this is breaking a unit test that checks against a couple of sample detections. I also noticed that this regex is already in the alert just a little further down in line 65, it is in an `any of 2` stanza so that might require some further tuning. 

The sample that you included for tuning also doesn't contain a "follow(ing) up" in the `body.current_thread.text`, so we might need to re-approach tuning this alert. 

Also, as a side, in your PRs if you could remove the "na-east-2" from your links it will redirect to Shared Samples so we don't have to jump into client-environments to check.

Thanks!

IndiaAce (changes requested):

detection-rules/spam_attendee_list_solicitation.yml
MSAdministrator: Thanks for this! I'll update based on your feedback!

detection-rules/spam_attendee_list_solicitation.yml
```diff
@@ -37,7 +37,7 @@ source: |
                           "(?:interest(s|ed)|accessing|purchas|obtain|acuir|sample|provide.{0,10}samples|counts|pricing)"
       )
       and not regex.icontains(body.current_thread.text,
-                              "(?:debit card|transaction.{0,20}processed|receipt)"
+                              "(?:debit card|transaction.{0,20}processed|receipt|purchasing?)"
```
IndiaAce: `|purchasing?` here is breaking a few of the unit tests. I managed to get it to match some of the random unit-tests by simply removing this stanza. Also, removing this stanza still allows the rule to match on your 2 samples so it might not be necessary.

IndiaAce (changes requested):

detection-rules/spam_attendee_list_solicitation.yml
MSAdministrator: Thanks! Updated!

IndiaAce (changes requested):
There are merge conflicts here that need to be resolved

MSAdministrator:
Okay, just tested this again and this doesn't seem to pass anymore so I don't think it's needed at this time. Closing this one out.

PR #2835, opened 2025-06-13
Update body_business_email_compromise_new_sender.yml

peterdj45:
# Description

adding logic to flag fake message threads that are nested within legit replies

# Associated samples

- https://platform.sublime.security/messages/aaa2ce92a60f427650a69b0bdfcc2f33e3f0b07a8358c6f703d5699e43b14df7

### DIFF
--- a/detection-rules/body_business_email_compromise_new_sender.yml
+++ b/detection-rules/body_business_email_compromise_new_sender.yml
@@ -15,11 +15,34 @@ source: |
       // out of office auto-reply
       // the NLU model will handle these better natively soon
       or strings.istarts_with(subject.subject, "Automatic reply:")
-      or regex.imatch(subject.subject, '(\[[^\]]+\]\s?){0,3}(re|fwd?|automat.*)\s?:.*')
+      or regex.imatch(subject.subject,
+                      '(\[[^\]]+\]\s?){0,3}(re|fwd?|automat.*)\s?:.*'
+      )
     )
     and (
-      length(headers.references) > 0
-      or any(headers.hops, any(.fields, strings.ilike(.name, "In-Reply-To")))
+      (
+        length(headers.references) > 0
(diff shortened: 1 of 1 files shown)


detection-rules/body_business_email_compromise_new_sender.yml
```diff
@@ -15,11 +15,34 @@ source: |
...
+              strings.icontains(., "date:"),
+              strings.icontains(., "cc:"),
+              strings.icontains(., "subject:")
+            )
+        )
+        and (
+          // newly registered sender domain
+          network.whois(sender.email.domain).days_old <= 30
+          // or suspiciously long HTML div id
+          or regex.icontains(body.html.raw, '<div\sid="(.{200,})".*>')
```
zoomequipd, line 43: matching on "not a `"` repeating at least 200 times, ensures that `"` is not matched within the `.{200,}` logic. 
```suggestion
          or regex.icontains(body.html.raw, '<div\sid=\"[^\"]{200,}\".*>')
```

zoomequipd, line 43: matching on "not a `"` repeating at least 200 times, ensures that `"` is not matched within the `.{200,}` logic. 
```suggestion
          or regex.icontains(body.html.raw, '<div\sid=\"[^\"]{200,}\".*>')
```

zoomequipd, line 43: Just so i'm understanding this, this provides a mechanism to override the previous thread negation based on the sender domain again and a div length? 

If that's correct, maybe a new rule that specifically handles that case is warranted instead? 

This hunt reflects that logic, @peterdj45  let me know what you think.  https://platform.sublime.security/messages/hunt?huntId=01990118-5c0f-7d33-9997-6e33ed867e71

zoomequipd (changes requested):

PR #2849, opened 2025-06-20
Creating new rule for suspect cloud service business documents

IndiaAce:
# Description
Leading to S3 buckets and using a suspicious domain (in this case dub.sh)
# Associated samples

- [Sample 1](https://na-east-2.platform.sublime.security/messages/7e3645be3cf1e55b3445a5bf6cacbc66114348fc30d864e0f317e7640ce2bf8f?preview_id=01976ac7-0b91-70c7-8026-3e59493d5863)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=01978dd7-7a7e-7b90-a5fb-2969bce38577)

### DIFF
--- a/detection-rules/suspicious_cloud_service_business_document.yml
+++ b/detection-rules/suspicious_cloud_service_business_document.yml
@@ -37,3 +37,4 @@ detection_methods:
   - "Header analysis"
   - "URL analysis"
   - "Sender analysis"
+id: "82226954-6865-50b5-a5bc-f3c7b0f38b32"


IndiaAce:
https://github.com/sublime-security/static-files/pull/472 

pre-req for this PR

IndiaAce:
Test results are looking great, ready for review

detection-rules/suspicious_cloud_service_business_document.yml
```diff
@@ -0,0 +1,40 @@
+name: "Suspicious URL Shortener: Cloud Storage and Business Documents"
+description: "Detects links combining cloud storage services and URL shorteners, particularly in business contexts like invoices and estimates, or in threads with financial approval language from untrusted senders."
+type: "rule"
+severity: "low"
+source: |
+  type.inbound
+  and (
+    // High-confidence: Suspicious cloud storage + URL shortener combo in business context
+    (
+      any(body.links, strings.icontains(.href_url.domain.domain, "tellwise-sites.s3.amazonaws.com"))
```
aidenmitchell, line 10: Is this a specific indicator we've seen? Do we want to just sub for any free subdomain?

detection-rules/suspicious_cloud_service_business_document.yml
```diff
@@ -0,0 +1,40 @@
...
+      and any(body.links, .href_url.domain.domain in~ $url_shorteners)
+      and any(["estimate", "invoice", "proposal", "repair", "service"], strings.icontains(subject.subject, .))
+      and not sender.email.domain.domain in~ $high_trust_sender_root_domains
+    )
+    or
+    // Medium-confidence: Multiple suspicious indicators together
+    (
+      (strings.istarts_with(subject.subject, "RE:") or strings.istarts_with(subject.subject, "FW:"))
+      and any(["approved", "please schedule", "NTE $"], strings.icontains(body.plain.raw, .))
+      and any(body.links, strings.icontains(.href_url.url, "dub.sh"))
```
aidenmitchell, line 20: Same here, we can use the `$url_shorteners` list perhaps?

IndiaAce: Good catch - it looks like that was my intention as I had added dub.sh to the url shorteners. It's likely this was just a hunt I was doing that I copied over instead of using the `$url_shorteners` I'll get this change in.

detection-rules/suspicious_cloud_service_business_document.yml
IndiaAce: This is a specific indicator we've seen. But I'm impartial to this decision, so we can certainly sub for the free subdomains. I'll re-hunt with these suggested changes and make some updates.

PR #2889, opened 2025-07-04
Create link_educated_manticore

brycampbell:
# Description

This rule is designed as an attempt to identify the phishing kits used by IGRC aligned activities as cited [here](https://research.checkpoint.com/2025/iranian-educated-manticore-targets-leading-tech-academics/) The phishing kit used by Educated Manticore is implemented as a Single Page Application (SPA) built with React, the rule is designed to identify this content as well as the TLD preferences mentioned.

I have implemented this as a ASR rule to attempt to justify the low confidence in this activity.

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=0197d4ad-86db-7229-8f1e-00adb6f2c9d9)


brycampbell:
This rule will be tested against additional live infrastructure once it becomes available.

aidenmitchell (approved):

PR #2896, opened 2025-07-04
Negate direct-to-VIP messages

aidenmitchell:
# Description

Negating messages that are directly sent to the VIP being "impersonated" in the subject.

# Associated samples
- https://platform.sublime.security/messages/672726abc80fdaf78338e1538c474becd85dd8ddbdecc9c11e6fca8846abc7b3?preview_id=0197d738-d8fe-7b4d-96c3-e939159c141c


lvh:
I don't think I can be a reviewer for this but I was the submitter than this looks right to me :)

PR #2913, opened 2025-07-09
Adding coverage for QR codes leading to CF CAPTCHA

aidenmitchell:
# Description
Adding coverage for QR codes leading to CF CAPTCHA
# Associated samples
- https://platform.sublime.security/messages/cea71867c0dad4d807bcb78755c0af772e9a4be88eed59f06739dcd71cd98ca8?preview_id=01948a74-ddfa-7e6d-bddf-e37cad68e906

### DIFF
--- a/detection-rules/link_suspicious_message_unscannable_cloudflare.yml
+++ b/detection-rules/link_suspicious_message_unscannable_cloudflare.yml
@@ -4,15 +4,18 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  and ( 
-    // few links
-    0 < length(body.links) < 20
-    // fewer unique root domain links
-    and length(distinct(body.links, .href_url.domain.root_domain)) < 10
-    // sender domain matches no body domains
-    and all(body.links,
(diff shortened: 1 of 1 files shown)


detection-rules/link_suspicious_message_unscannable_cloudflare.yml
```diff
@@ -231,48 +234,88 @@ source: |
...
-              ) < 200
+  and (
+    any(body.links,
+        (
+          strings.icontains(ml.link_analysis(., mode="aggressive").final_dom.display_text,
+                            "cloudflare"
+          )
+          // includes the turnstile CAPTCHA
+          or (
+            strings.icontains(ml.link_analysis(., mode="aggressive").final_dom.raw,
```
zoomequipd, line 245: you might also be able to use the `unique_urls_accessed` and look for URls going to cloudflare with captcha resources?

aidenmitchell: Yeah, this is likely the same effect, but good call

PR #2939, opened 2025-07-11

markmsublime:
# Description

Additional coverage to account for non-plural brand name. 

[- Sample 1](https://platform.sublime.security/messages/48b8e62e61e109eac0c11b471403fbad270c30b04fb841f17b73acb6ce8fef83?preview_id=01971767-c295-7064-9c59-f976cb5bf195)

### DIFF
--- a/detection-rules/impersonation_quickbooks.yml
+++ b/detection-rules/impersonation_quickbooks.yml
@@ -6,7 +6,7 @@ source: |
   type.inbound
   and (
     (
-      strings.ilike(sender.display_name, '*quickbooks*', 'quickbook*')
+      strings.ilike(sender.display_name, 'quickbook*')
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_quickbooks.yml
```diff
@@ -6,7 +6,7 @@ source: |
   type.inbound
   and (
     (
-      strings.ilike(sender.display_name, '*quickbooks*')
+      strings.ilike(sender.display_name, '*quickbooks*', 'quickbook*')
```
IndiaAce: `strings.ilike(sender.display_name, '*quickbook*')`
This should accomplish the same thing... not sure if you ran hunts with this logic but I think this is a more "efficient" way as well.

PR #2954, opened 2025-07-16
feat: Adding IOK converter

MSAdministrator:
# IOK Converter

This is a new converter which can download and convert [IOK Rules](https://github.com/phish-report/IOK) (Indicator of Kit) to a sublime rule (e.g. MQL format).

You can download only or download and automatically convert the rules using this python package.

This converter transforms website-focused IOK phishing kit indicators into email-focused Sublime Security rules by:

- **🔗 Using Link Analysis**: Converts website HTML patterns to `ml.link_analysis()` for analyzing linked websites
- **🏷️ Brand-Specific Detection**: Intelligent email subject and sender domain validation
- **📧 Email Context**: Adapts website indicators to email security use cases


IndiaAce:
Curious if you think we should overhaul the CI tests before working our way through some of these failed builds. Or if we should make an IOK directory? I'm not married to either of those ideas and maybe there's a third option.

MSAdministrator:
> Curious if you think we should overhaul the CI tests before working our way through some of these failed builds. Or if we should make an IOK directory? I'm not married to either of those ideas and maybe there's a third option.

Yeah, I'm adding them to a new directory called `iok-rules` and will update CI to exclude that or incorporate it -- haven't decided to be honest. I'm starting to look into that now.

MSAdministrator:
This currently being closed as we have decided to instead integrate these rules somewhere else.

PR #3010, opened 2025-07-29
Update body_extortion.yml

brycampbell:
# Description

Added in more words from the extortion attempt, that we currently do not identify

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/4f49edf2e9051d02921cf807bd3d618313f973a87dd89654c56164211a4d2b20?preview_id=019852a2-1672-7cf2-9b9d-8c08694d1f4b)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=01985603-49f5-7059-9811-de4ffe72c711)

# Screenshot (insights)

### DIFF
--- a/detection-rules/abuse_google_classroom.yml
+++ b/detection-rules/abuse_google_classroom.yml
@@ -1,0 +1,48 @@
+name: "Google Classroom Spoofing With WhatsApp Contact Information"
+description: "Detects messages impersonating Google Classroom notifications that contain WhatsApp contact information, likely attempting to redirect victims to out-of-band communication channels for social engineering attacks."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and sender.email.email == "no-reply@classroom.google.com"
+  and (
+    // check for a WhatsApp invitation in the currend_thread
+    (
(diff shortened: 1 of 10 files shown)


detection-rules/body_extortion.yml
```diff
@@ -34,14 +34,14 @@ source: |
     or 3 of (
       // malware terms
       regex.icontains(strings.replace_confusables(body.current_thread.text),
-                      "((spy|mal)ware|t[rŗ]ojan|remote control|infiltrat(ed|ion))"
+                      "((spy|mal)ware|t[rŗ]ojan|remote control|backdoor|intruder|comprom)"
```
zoomequipd: looks like this one got removed? `infiltrat(ed|ion)`

zoomequipd:
made a pretty large commit here 
Orignal exclusion matches by the change
https://platform.sublime.security/hunts/0198fbc2-5bc4-7a62-b82a-d1211ff5cd88

after my commit
https://platform.sublime.security/messages/hunt?huntId=0198fc26-1136-751f-a5b2-6e54f64e3f85

will leave in testing

zoomequipd:
looks like only an increase in TP hits, some FPs but the existing rule also matches them.

PR #3011, opened 2025-07-29
Update attachment_qr_code_suspicious_components.yml

vector-sec:
# Description

In situations where an email contains an attachment with a QR code, but the recipients.to is only "undisclosed-recipients" the sections like

```
any(recipients.to,
    strings.icontains(..scan.qr.url.url, .email.email)
)
```

always evaluate to true. I tried setting the filter for where email isn't null, but that seemed to not work either, so I went with where length of email.email > 0.

After this change, the false positive that I have with "undisclosed-recipients" and a benign URL bearing QR code in an attachment is no longer flagged by this message.

# Associated samples

The Sublime team may find a sample I shared in the platform with the comment `PR - Attachment: QR code with credential phishing indicators`

## Associated hunts

- Hunt 1

# Screenshot (insights)

### DIFF
--- a/detection-rules/attachment_qr_code_suspicious_components.yml
+++ b/detection-rules/attachment_qr_code_suspicious_components.yml
@@ -29,7 +29,7 @@ source: |
           ) > 3
         )
         or any(filter(attachments, .file_extension in ('pdf', 'docx')),
-               any(recipients.to,
+               any(filter(recipients.to, length(.email.email) > 0),
                    strings.icontains(..file_name, .email.domain.sld)
                )
         )
@@ -90,7 +90,7 @@ source: |
                                  )
                       )
                       or (
-                        any(recipients.to,
+                        any(filter(recipients.to, length(.email.email) > 0),
                             strings.icontains(..scan.qr.url.url, .email.email)
                         )
                       )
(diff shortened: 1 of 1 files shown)


IndiaAce:
Just a few in-flight observations here @zoomequipd - the sample EMLs that are no longer matching also contain `Undisclosed recipients` as the recipients. I'm not 100% sure we should be anchoring on that as a negation. Unless we did some more `or` logic. Another thing to note is that this rule seems to have pre-existing issues with nested and/or logic. See stanzas 72 or 122 for example - also present in the current logic. 
 
I think this stanza is the cause of the headaches: 
`                  any(file.explode(.),
              .scan.qr.type == "url"
              and not .scan.qr.url.domain.domain == "geico.app.link"
              and (
                    any(recipients.to,
                        strings.icontains(..scan.qr.url.url, .email.email)
                        or any(beta.scan_base64(..scan.qr.url.url,
                                                ignore_padding=true
                               ),
                               strings.icontains(., ..email.email)
                        )
                        or any(beta.scan_base64(..scan.qr.url.fragment,
                                                ignore_padding=true
                               ),
                               strings.icontains(., ..email.email)
                        )
                  )
              )
          )
`
We're checking were the `file.explode` urls contain the email. But when it's empty it looks like it's matching every url in the file explode. But if we were to make a change there, I think we might need to exclude some of these test emls. 

* Check out this canonical that contains a recipient and matches the above stanza: `4f424a3ed42fe0e12a88107c8f0b0263a5a721a94568e61c06e273c61b6c80a3` 

vs. this canonical that has no recipients and still matches: `4f4ef5d4fa1f28aba05019c46c994979bfb0350a84e91d10cda6fe71331ed5a2`

Sorry I couldn't fully resolve this, but that's the extent of what I could figure out... I'm keeping a close eye on this I'm curious to see the resolution. Feel free to ignore all of this and pursue a different conclusion.

PR #3016, opened 2025-07-30
Update scam_piano.yml

missingn0pe:
# Description

Forgot to add key phrasing to the attachment parse section & added type.inbound to the rule logic.


missingn0pe (commented):
Adding and argument.

PR #3027, opened 2025-07-31
Create link_multiple_http_protocols_in_single_url.yml

morriscode:
# Description
New ASR coverage for heavy link obfuscation through multiple redirection/security services.

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=01985e37-998e-79d4-a4ac-c812ad1b5d0c)

### DIFF
--- a/detection-rules/link_multiple_http_protocols_in_single_url.yml
+++ b/detection-rules/link_multiple_http_protocols_in_single_url.yml
@@ -1,0 +1,18 @@
+name: "Link: Multiple HTTP protocols in single URL"
+description: "Detects messages containing links with 5 or more HTTP protocol declarations within a single URL, indicating potential URL manipulation or obfuscation techniques."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and 0 < length(body.current_thread.links) < 10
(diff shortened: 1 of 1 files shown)


zoomequipd:
Looking through test results, i had a couple ideas to improve this idea.  
Note: your reference hunt no longer matches the included rule due to being parsed out via url decoders

After noticing the test-rules results matched when many different URL query params each contained a single URL, the idea was that the URLs should be within a single URL query param.  I had to reconstruct some of the functionality provided by query_params_decoded, but I think this worked well. 

Once I was able to narrow the matches down to a single query param i relized it's really _different_ domains of the URLs that was interesting.  I was able to reduce FPs by extracting the domain out of the "embedded" url and then ensure they were different and _not_ the same as the sender domain, then count the number of unique domains embedded in the query param. 

Hunt and proposed MQL:  https://platform.sublime.security/messages/hunt?huntId=01990742-61de-7a2f-a468-1ac9f785cf0a

zoomequipd:
looks like the addition of the unsubscribe negation had a very positive impact on this rule, will continue testing for a bit more

PR #3073, opened 2025-08-13
Update vip_impersonation_local_part.yml

mrobertsonris:
Salesforce FP across several Orgs.

# Description

Shared with Sublime as `SalesforceFP`

### DIFF
--- a/detection-rules/vip_impersonation_local_part.yml
+++ b/detection-rules/vip_impersonation_local_part.yml
@@ -45,6 +45,8 @@ source: |
     )
     or sender.email.domain.root_domain not in $high_trust_sender_root_domains
   )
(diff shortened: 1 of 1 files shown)


zoomequipd:
@mrobertsonris  sorry for the delay here, it looks like this PR got lost for a bit.  

I was able to find the reference samples, but I have concerns about in effect disabling this rule for all messages originating from salesforce given the amount of salesforce abuse being observed.  The concern is elevated for those that don't have/use salesforce.

I have some specific questions about the samples provided, I'll reach out via DM.

zoomequipd:
spoke with @mrobertsonris  and we have opted to close this PR.

PR #3156, opened 2025-08-29
Update attachment_qr_code_suspicious_components.yml

markmsublime:
Re-opened from #3011 -
# Description

In situations where an email contains an attachment with a QR code, but the recipients.to is only "undisclosed-recipients" the sections like

```
any(recipients.to,
    strings.icontains(..scan.qr.url.url, .email.email)
)
```

always evaluate to true. I tried setting the filter for where email isn't null, but that seemed to not work either, so I went with where length of email.email > 0.

After this change, the false positive that I have with "undisclosed-recipients" and a benign URL bearing QR code in an attachment is no longer flagged by this message.

# Associated samples

The Sublime team may find a sample I shared in the platform with the comment `PR - Attachment: QR code with credential phishing indicators`

### DIFF
--- a/detection-rules/attachment_qr_code_suspicious_components.yml
+++ b/detection-rules/attachment_qr_code_suspicious_components.yml
@@ -28,8 +28,9 @@ source: |
                  )
           ) > 3
         )
+        // the attachment name contains the SLD of a recipient
         or any(filter(attachments, .file_extension in ('pdf', 'docx')),
-               any(recipients.to,
+               any(filter(recipients.to, .email.domain.valid),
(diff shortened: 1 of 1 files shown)


markmsublime:
/mql-mimic-exempt: 471106, 529346, 715926, 716251, 716120, 716121, 716825, 716305, 716328, 716883, 726254, 722551, 726258, 726259, 726263, 726264, 726261, 726256, 776630, 776494, 780613, 780633, 780651, 780713, 782402, 776863, 880917, 928281

Reviewed all these samples and none of the QR codes had links with email addresses in them, which is what the logic is looking for and thus should have not fired in the first place.

PR #3158, opened 2025-08-29
Insight: Updating Links in Attachments to remove schema urls from common attachments

MSAdministrator:
# Description

It has been requested to remove schema URLs from common attachment types. These are mostly focused on Microsoft docs but other variations are included as well.

* purl.org - Sometimes used for Dublin Core metadata namespaces
* www.w3.org- For standard XML, XHTML, or MathML content
* purl.oclc.org - Alternative location for Dublin Core elements
* schemas.openxmlformats.org subdomains or variations

# Associated samples

- Sample 1
- Sample 2

## Associated hunts

- Hunt 1

# Screenshot (insights)

PR #3162, opened 2025-09-02
Insight: Filtering out common schema domains in Domains in Attachments

MSAdministrator:
# Description

This PR is related to #3151 and #3158. It is removing schema URls/domains from these insights to reduce cognitive overload.

Removing the following root_domains:
* "wps.cn"

Removing the following domains:

* "schemas.openxmlformats.org"
* "schemas.microsoft.com"
* "purl.org"
* "www.w3.org"
* "purl.oclc.org"
* "schemas.apple.com"

# Screenshot (insights)

484685984-b6537035-6b27-450e-9f18-369bd41b8ff1
484685988-14ec4f5e-4f95-482b-9dd4-83b5ea3e35a4
484685989-2cf97a81-c90a-4650-8f3e-21f233ca57e2

### DIFF
--- a/insights/attachments/links_domains.yml
+++ b/insights/attachments/links_domains.yml
@@ -4,7 +4,6 @@ source: |
   filter(map(attachments,
            map(file.explode(.),
                distinct(map(filter(.scan.url.urls,
-                                   .domain.root_domain not in $tranco_1m
                                    and .domain.root_domain not in $org_domains
                                    and .domain.root_domain not in (
                                      "sublimesecurity.com",
--- a/insights/attachments/links_domains.yml
+++ b/insights/attachments/links_domains.yml
@@ -1,12 +1,25 @@
 name: "Domains in attachments"
 type: "query"
 source: |
-  filter(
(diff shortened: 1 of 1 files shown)


insights/attachments/links_domains.yml
```diff
@@ -1,12 +1,30 @@
...
-        map(file.explode(.),
-            distinct(map(.scan.url.urls, .domain.domain), .)
-        )
-    ),
-    length(.) > 0
+  filter(map(attachments,
+           map(file.explode(.),
+               distinct(map(filter(.scan.url.urls,
+                                   .domain.root_domain not in $tranco_1m
+                                   and .domain.root_domain not in $org_domains
```
zoomequipd: filtering out tranco is not wanted on this one, that only applies to the low reputation insight
```suggestion
                                   .domain.root_domain not in $org_domains
```

zoomequipd (changes requested):

zoomequipd (approved):

PR #3166, opened 2025-09-03

zoomequipd:
# Change 1

Resolves FNs which were the result of replace_confusables not replacing all unicode letters with the "desired" letter (U+0399 -->  U+004C (Latin Capital Letter ) instead of U+0049 (Latin Capital Letter I)

1. better handle cases where confusables get replaced with similar but not the "desired" letters
2. sync up blocks between body.current_thread and screenshot (even though screenshot should not be "confused" by the confusables, keeping them in sync makes it easier to maintain)

- [Sample 1](https://platform.sublime.security/messages/4f6ddce54413bcb6023e6cd5837bcb5a1efa0a0f10a903529f4fd3f8e85d5491?from_canonical_id=4f6eeeaddd888fa59b82359d3a9762b057a57af194a0f1c070a7ca3b0a5bfcb1&preview_id=01991061-f490-73d2-943e-fa718ea27b51)

# Change 2

Add "preview" to logic of action words for display text links
- [Sample 1](https://platform.sublime.security/messages/4f67eea630ba10ed20ecb9887e43c94ba1f73fee12a5fc95047bfa9a993a2cba?preview_id=01991061-f89d-78e5-9a02-1523c9dd2fa8)

# Change 3
update references of `voice` to `v[o0][il1]ce`

### DIFF
--- a/detection-rules/link_credential_phishing_voicemail_language.yml
+++ b/detection-rules/link_credential_phishing_voicemail_language.yml
@@ -72,7 +72,7 @@ source: |
     or (
       length(attachments) > 0
       and (
-        all(attachments, .file_type in $file_types_images)
+        all(attachments, .file_type in $file_types_images and beta.parse_exif(.).image_height != 1)
         or (
           // there is a mix of fake audio attachments and images
           length(filter(filter(attachments,
@@ -86,7 +86,7 @@ source: |
                  )
           // the total # of fake audio attachments + the total # of image attachments = the total # of attachments
           // meaning, all attachments that are NOT fake audio attachments MUST be images
-          ) + length(filter(attachments, .file_type in $file_types_images)) == length(attachments
+          ) + length(filter(attachments, .file_type in $file_types_images and beta.parse_exif(.).image_height != 1)) == length(attachments
           )
         )
       )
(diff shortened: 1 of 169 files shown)


detection-rules/link_credential_phishing_voicemail_language.yml
```diff
@@ -90,36 +91,36 @@ source: |
...
-                )
-                and not regex.icontains(body.current_thread.text,
-                                        '(?:I(?:\sjust)?|just) left you a (?:\w+(\s\w+)?|[[:punct:]]+|\s+){0,3}(?:voice(?:mail)?|audio)(?: message)?'
-                )
-                and not regex.icontains(body.current_thread.text,
-                                        'you (?:have |received )my voice\s?(?:mail|audio|message)'
-                )
-              )
+      and (
+        regex.icontains(beta.ocr(beta.message_screenshot()).text,
```
IndiaAce:
```suggestion
        regex.icontains(beta.ocr(file.message_screenshot()).text,
```
There are a few more places in this rule that should have this swapped out for.

IndiaAce (changes requested):

PR #3214, opened 2025-09-11
Update attachment pdf comp review.yml

se-taylor:
# Description

Identified issues with the usage of the file.explode function for this rule. Due to the file.explode function treating content within a PDF attachment as separate attachments, the conditions for the PDF attachment, any QR code within it, and any text within it need to be handled separately. Update splits the existing conditions for the PDF attachment, QR code, and attachment text into separate file.explode loops.

# Associated Samples/Hunt

Testing of the new logic was carried out within my organisation's private Sublime tenant so I'm unable to share samples/hunts.

### DIFF
--- a/detection-rules/body_business_email_compromise_new_sender.yml
+++ b/detection-rules/body_business_email_compromise_new_sender.yml
@@ -11,15 +11,23 @@ source: |
   // negating legit replies
   and not (
     (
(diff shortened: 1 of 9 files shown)


IndiaAce (approved):
This is an excellent find and a great suggestion, thank you! Looks good to us.

PR #3215, opened 2025-09-11
Create brand_impersonation_gov_uk_img_src.yml

zoomequipd:
# Description

Add coverage for observed technique for SMS cred phish

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=01993934-56e5-7c2a-8064-6cc40a058976)


zoomequipd:
not a viable rule option - too many FPs due to auth failures and benign use of imgs src'ed from .gov.uk

PR #3237, opened 2025-09-16
Create google_appsheet_abuse_suspicious_url_display_name.yml

morriscode:
# Description

Google Appsheet abuse 

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=0199527a-13dc-7e55-b630-b48e7f028c5e)


detection-rules/google_appsheet_abuse_suspicious_url_display_name.yml
```diff
@@ -0,0 +1,38 @@
...
+  and sender.email.email == "noreply@appsheet.com"
+  and (
+    (
+      any($org_display_names, strings.contains(subject.base, .))
+      or any($org_slds, strings.contains(subject.base, .))
+    )
+    or (
+      any(body.links,
+          network.whois(.href_url.domain).days_old <= 10
+          or .href_url.domain.root_domain in $free_file_hosts
```
zoomequipd: appsheet itself is a free_file_host which causes this rule to FP on every appsheet email
```suggestion
          or (
            .href_url.domain.root_domain in $free_file_hosts
            and not .href_url.domain.root_domain == "appsheet.com"
          )
```

PR #3245, opened 2025-09-17
Create abuse_todo_list.yml

IndiaAce:
# Description
From a runner, testing coverage expansion for malicious todo list. My goal here is to just let this sit in test rules for a few days to see how it performs and iterate on the rule. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/4f7370d6dfd77d43d7c967fbebe938d69492cd448958655fe1aeb4b12449289c?preview_id=01992948-4a24-7877-8471-4f305c4f8cf7)

Edit one: expanding coverage preemptively to capture stuff like the following
- [Sample 2](https://platform.sublime.security/messages/4f80910eaab296f2085131d27cdd1aa69fd0b9365d241378a565503c1445b57a?preview_id=019971ed-027f-7b32-b3b9-bd24a8869fd8)

## Associated hunts
- [Hunt 1
](https://platform.sublime.security/messages/hunt?huntId=01995454-34ad-76db-ba6a-e60de047e53a)
- [Hunt 2 - goldmine](https://platform.sublime.security/messages/hunt?huntId=01997860-2f6b-795f-8c6e-951713b4b9fd)

### DIFF
--- a/detection-rules/abuse_todo_list.yml
+++ b/detection-rules/abuse_todo_list.yml
@@ -18,10 +18,7 @@ source: |
     )
   )
   // utilizing sendgrid infra
-  and (
-    strings.icontains(headers.return_path.domain.root_domain, "sendgrid")
-    or any(headers.domains, strings.icontains(.root_domain, "sendgrid"))
-  )
+  and headers.return_path.domain.domain == "sendgrid.net"
   // negate highly trusted sender domains unless they fail DMARC authentication
   and (
     (
--- a/detection-rules/abuse_todo_list.yml
+++ b/detection-rules/abuse_todo_list.yml
@@ -4,32 +4,57 @@ type: "rule"
 severity: "high"
 source: |
   type.inbound
-  //subject contains "todo list"
+  // subject contains "todo list"
   and (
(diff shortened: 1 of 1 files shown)


detection-rules/abuse_todo_list.yml
```diff
@@ -0,0 +1,51 @@
...
+      and any(ml.nlu_classifier(body.current_thread.text).intents,
+              .name == "cred_theft" and .confidence == "high"
+      )
+    )
+  )
+  // utilizing sendgrid infra
+  and (
+    strings.icontains(headers.return_path.domain.root_domain, "sendgrid")
+    or any(headers.domains, strings.icontains(.root_domain, "sendgrid"))
+  )
```
aidenmitchell: Other rules use `and headers.return_path.domain.domain == "sendgrid.net"`,  is there a specific reason for this wider net?

IndiaAce: I was working with two different samples at one point and I think I thought I needed it... but it matches both rules with your logic so I'm updating it to that.

aidenmitchell (approved):

PR #3298, opened 2025-09-24
Update impersonation_barracuda.yml

peterdj45:
# Description

negating `barracudashoes.it`

# Associated samples
- https://platform.sublime.security/messages/4f7fc4f012c5976848bf3386c388bce4d20d1eca76421c52670a09978073a402
(fired in customer env but not shared samples due to sender prevalence)

### DIFF
--- a/detection-rules/impersonation_barracuda.yml
+++ b/detection-rules/impersonation_barracuda.yml
@@ -27,7 +27,10 @@ source: |
     'worldspan.co.uk',
   
   // Barracudas Day Camps
-    'barracudas.co.uk'
+    'barracudas.co.uk',
+  
+  // BarracudaShoes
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_barracuda.yml
```diff
@@ -27,8 +27,12 @@ source: |
     'worldspan.co.uk',
   
   // Barracudas Day Camps
-    'barracudas.co.uk'
+    'barracudas.co.uk',
+  
+  // BarracudaShoes
+    'barracudashoes.it'
```
IndiaAce:
```suggestion
    // BarracudaShoes
    'barracudashoes.it'
```
small nit here - can we also re-align the comments on lines 26 and 29?

PR #3307, opened 2025-09-26
Fix ignoring testing_sha only test-rule updates

cameron-dunn-sublime:
# Description

This reintroduces changes made in #3279 which were reverted in #3292.

I refactored and cleaned up changes which originally started in #3189. The approach there was not great and it was confusing. [Here](https://github.com/sublime-security/sublime-rules/compare/df81636e218c4ca98e218e7beaa4feb8090eafce...cd.fix-testing-sha-only-changes#diff-46da79a1bb80815448d4b83787836733bedb89d169329dd3bf05b98474d1128d) you can see the diff between this change and the version of the workflow before I started messing with it (I can't link the file, and I've only changed the `Commit changes` step of `update-test-rules.yml`) -- basically these changes are simpler than what was originally done.

## Testing
Rather than local testing, I spun up a fork and got its `test-rule` syncing working and then debugged through all the cases I knew to care about:

1. New rule
2. Modifications to an existing rule
3. Closing a PR
4. Updating a branch w/o changing the rest of the rule

This seems to at last work. I've also left some debug printing which should be helpful to understand what all is going on in case of future issues.

### DIFF
--- a/.github/workflows/update-test-rules.yml
+++ b/.github/workflows/update-test-rules.yml
@@ -153,24 +153,17 @@ jobs:
           GIT_STATUS=$(git status --porcelain "$FILE" | cut -c1-2 | tr -d ' ')
           if [[ "$GIT_STATUS" == "A" ]]; then
             ADDED_COUNT=$((ADDED_COUNT+1))
+            STATUS="added"
           elif [[ "$GIT_STATUS" == "M" ]]; then
             MODIFIED_COUNT=$((MODIFIED_COUNT+1))
+            STATUS="modified"
           else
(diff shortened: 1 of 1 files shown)


.github/workflows/update-test-rules.yml
zoomequipd: i think this is duplicated now with lines 152-160 after pulling out the other logic.

cameron-dunn-sublime: Good catch: https://github.com/sublime-security/sublime-rules/pull/3307/commits/3b56d6ce3a03d9119c5373d78ae9a541f66d265f

zoomequipd (approved):

PR #3373, opened 2025-10-07
Create impersonation_vip_local_part_display_name.yml

zoomequipd:
# Description
Add coverage for when the local part and display name match a VIP but either the domain is different or authentication fails.


zoomequipd:
too much noise

PR #3374, opened 2025-10-07
Create link_gophish_rid.yml

zoomequipd:
# Description

coverage for default GoPhish URL format

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=0199bc25-642e-7af7-8f72-360f4d6d3098)


zoomequipd:
> A couple greymail hits, is that what you're aiming for?

No :-(

zoomequipd:
low telemetry, but should be good now.

PR #3391, opened 2025-10-08
Use checkout for sublime-security/sublime-rules for scripts

cameron-dunn-sublime:
# Description

Ensure that we cannot run a 3rd parties script in our own context. Tested in https://github.com/sublime-security/sublime-rules/pull/3392


zoomequipd (approved):

dlynch-sublime (approved):
lgtm

zoomequipd (approved):

PR #3491, opened 2025-11-07
Enhance impersonation detection for Microsoft Teams

peterdj45:
# Description

adding additional regex keywords to account for microsoft teams impersonation

also adding logic to override sender profile and explicit domain negations if the sender domain is not valid

# Associated samples
- https://platform.sublime.security/messages/4facb1b32712710d6216d0bdd455f09bbbc54066f759d13b74e1475d2ec20633


aidenmitchell (dismissed):

zoomequipd (changes requested):
there are many benign matches in telemetry data.

PR #3556, opened 2025-11-20
Enhance detection rules for SSA impersonation

peterdj45:
# Description

adding logic to flag samples referencing Login.gov in the subject or sender display name

expanding NLU topic check to include Government Services

merging regex conditions into a single `any` statement

# Associated samples
- https://platform.sublime.security/messages/4fbaab71e7cbfe4a08aae4dea254566c70a7649acda4cbcfbf493ff1f0acbf4c
- https://platform.sublime.security/messages/4fbb0775da5cf5deca31d87ac69fb21a0dbd525e274928d2619a10742ddddf68
- https://platform.sublime.security/messages/4fbb5c0abf1cc09f83bb2b40be01916aa6e9bb84ae694ff974e107f8661d4663

## Associated hunts

- https://platform.sublime.security/messages/hunt?huntId=019aa02f-4058-75bd-b95f-89e6bb8e681a - hunt for messages containing Government Services NLU topic


detection-rules/impersonation_social_security_admin.yml
```diff
@@ -39,7 +45,8 @@ source: |
   // Additional suspicious indicator
   and (
     any(ml.nlu_classifier(body.current_thread.text).topics,
-        .name == "Secure Message" and .confidence == "high"
+        .name in~ ("Secure Message", "Government Services")
```
zoomequipd: nit: given that topics are a enum, we shouldn't need the case insensitive option here.

zoomequipd (changes requested):
no telemetry yet.  would like this to either be backfilled with multiple hunts or let it sit in test rules a bit longer to make sure we're not introducing new FPs.

multiple hunts from different environment for just the "net new" matches would be helpful

PR #3789, opened 2026-01-15
Add detection rule for suspicious Google Meet invitations

peterdj45:
# Description

This rule detects suspicious Google Meet invitations that may be used for fraudulent activities, analyzing calendar invitations for specific indicators and content.

# Associated samples
- https://platform.sublime.security/messages/4fafbd98d85e045ce1651dc1d8d045f5c4071c9f3d9305e493dae5a926a88856
- https://platform.sublime.security/messages/8a2e4c908e9b8a9dcadf5e7c32516be8636a105ddf2f1ff2812b49723c549132

### DIFF
--- a/detection-rules/abuse_google_meet.yml
+++ b/detection-rules/abuse_google_meet.yml
@@ -45,3 +45,4 @@ detection_methods:
   - "Content analysis"
   - "HTML analysis"
   - "URL analysis"
+id: "92fa8397-4c0a-56e1-9990-37b15feeb95f"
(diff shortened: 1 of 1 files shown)


peterdj45:
Closing this out. The scaffolding (Invitation: + 2 ics attachments + meet.google.com Join link) heavily overlaps with the existing callback_phishing_google_meet.yml rule, and the BEC/partnership-offer payload here is better handled by extending existing Google Meet/Calendar coverage than as a standalone rule. Can revisit as an enhancement to those rules if the pattern warrants it.

PR #3815, opened 2026-01-21
Update attachment_fake_vm_pdf.yml

MSAdministrator:
# Description

During a hunt I noticed a sample that didn't match this rule but it only needed a small update. Hunts look good but will wait for mode results.

# Associated samples

- 4fbc5119500242cd314ee22862522f1a73d4f77a291eeb95490afbe77becff8f

## Associated hunts

- [Hunt negating existing rule over 60 days](https://platform.sublime.security/messages/hunt?huntId=019becfc-6254-7d00-a62b-48648cb01038) -- there's only a single match but it's something. ;)
- [Hunt negating existing rule - latest](https://platform.sublime.security/messages/hunt?huntId=019d6de3-9326-74bd-9be3-24f3c04170bd)


detection-rules/attachment_fake_vm_pdf.yml
```diff
@@ -8,18 +8,28 @@ source: |
...
+              .confidence == "high"
+              and .name == "Voicemail Call and Missed Call Notifications"
+      )
+    )
+    or (
+      // some sample have subjects that are not long enough
+      length(ml.nlu_classifier(subject.base).topics) == 0
+      // but the message has different content in plain vs current thread
+      and not strings.contains(body.plain.raw, body.current_thread.text)
+    )
```
zoomequipd: the more "direct" signal here is that there isn't enough data in the body.current_thread.text to show the `Voicemail Call and Missed Call Notifications` topic, in which case, I'd probably do that more direct.  Consider just leave this out and if let `length(ml.nlu_classifier(subject.base).topics) == 0` be the factor that determines that? 

Hunt - https://platform.sublime.security/messages/hunt?huntId=019bfd12-8ad0-7678-bcba-c1be09b81d29

```suggestion
      // some sample have subjects that are not long enough
      or length(ml.nlu_classifier(subject.base).topics) == 0
```

zoomequipd (changes requested):

detection-rules/attachment_fake_vm_pdf.yml
```diff
@@ -31,6 +41,7 @@ source: |
           // there is only a single page
           and any(file.explode(.), .depth == 0 and .scan.exiftool.page_count == 1)
   )
+  and not (sender.email.domain.root_domain == "zendesk.com")
```
IndiaAce:
```suggestion
  and not (
    sender.email.domain.root_domain == "zendesk.com"
    and coalesce(headers.auth_summary.dmarc.pass, false)
  )
```

IndiaAce (dismissed):
These changes look pretty good to me! Consideration for the zendesk matches would be to enforce auth passes on those domains... orrrrrr depending on the volume we could remove this line completely and allow "layer 2" to handle these sender exclusions.

detection-rules/attachment_fake_vm_pdf.yml
```diff
@@ -7,19 +7,29 @@ source: |
...
+              and .name == "Voicemail Call and Missed Call Notifications"
+      )
+    )
+    or (
+      ml.nlu_classifier(body.current_thread.text).language == "english"
+      and any(ml.nlu_classifier(body.current_thread.text).intents,
+              .confidence == "high" and .name == "bec"
+      )
+    )
   )
```
zoomequipd, line 20: i think this can be simplified like this? 
```suggestion
  and not strings.icontains(subject.base, 'fax')
  and ml.nlu_classifier(body.current_thread.text).language == "english"
  and (
    any(ml.nlu_classifier(body.current_thread.text).topics,
        .confidence == "high"
        and .name == "Voicemail Call and Missed Call Notifications"
    )
    or any(ml.nlu_classifier(body.current_thread.text).intents,
           .confidence == "high" and .name == "bec"
    )
  )
```

zoomequipd (changes requested):

zoomequipd (approved):

PR #3825, opened 2026-01-22
Enhance detection rule for SharePoint link sharing

peterdj45:
# Description

Adding condition to override solicited senders if more than 7 days have passed since last contact

# Associated samples
- https://platform.sublime.security/messages/4ffa0f8e41fb0eb72036134ab4acd7cca9520b4298094bc624b6682e3461bc88


detection-rules/link_sharepoint_sus_name.yml
```diff
@@ -126,6 +126,8 @@ source: |
   and (
     // and sender has never had email sent to them
     profile.by_sender().solicited == false
+    // more than 7 days since the sender was last contacted
+    or profile.by_sender_email().days_since.last_contact > 7
```
IndiaAce: I like this in concept, but I think my concern here is what happens to orgs are sending monthly invoices via sharepoint? That could get a little tricky... curious what your thoughts are on that. Gonna remove `review-needed` I like the results, just concerned about some FPs

IndiaAce (changes requested):

detection-rules/link_sharepoint_sus_name.yml
peterdj45: I think that's a valid concern, but likely could be resolved through a sender exclusion/attack score? 

Also wouldn't be opposed to bumping the condition up to over 30 days (the original sample was like 31 days since last contact so it'll still flag).

IndiaAce: > but likely could be resolved through a sender exclusion/attack score?

good point

peterdj45:
too many solicited sender FPs > 7 days but < 30. back to the drawing board

PR #3856, opened 2026-01-26
[senderprofile] Update link_google_presentation_open_redirect.yml

MSAdministrator:
# Description

Removing sender profile from top 10 rules.

This is related to task Exploration of removing FP gates from core feed rules when they have been overly suppressed


MSAdministrator:
We are closing this PR as the initiative is wrapping up.

PR #3893, opened 2026-01-28
Add detection rule for BAT file attachments

peterdj45:
# Description

This rule detects inbound messages with batch files as attachments or within compressed archives.

# Associated samples
- https://platform.sublime.security/messages/4ff835e44f7d1e0ead380a0d9212ec6c85a18d7f66f617f873c2a95a7424ce05?preview_id=019bd899-1d45-7c29-98ca-23339ebcc513&organization_id=dd9bdcae-c771-4263-9d48-71ea47cd49c2

## Associated hunts
- https://platform.sublime.security/messages/hunt?huntId=019c0640-7cd4-7283-ab73-69e48b2b98ec


peterdj45:
only seeing FPs in the wild, don't think this can be deployed even as an ASR rule.

PR #3958, opened 2026-02-04
Add CI check for missing file extensions

aidenmitchell:
## Summary
- Adds a new validation step to catch rule files that are missing the `.yml` extension
- Checks both `detection-rules` and `insights` directories
- Provides clear error output showing which files are missing extensions

## Test plan
- [x] Verified locally that the check passes when all files have extensions
- [x] Verified locally that the check catches files without extensions


cameron-dunn-sublime (approved):

.github/workflows/rule-validate.yml
```diff
@@ -246,6 +246,18 @@ jobs:
         run: |
           ! /bin/sh -c 'ls *.yml'
 
+      - name: Verify all rule files have .yml extension
+        if: github.event_name != 'issue_comment'
+        run: |
+          # Find files without extensions in rule directories
+          files_without_ext=$(find detection-rules insights -type f ! -name "*.*" 2>/dev/null || true)
```
zoomequipd: should we extend to DLP?

zoomequipd (approved):
non-blocking feedback

PR #4123, opened 2026-03-05
Add threat detection rules for acts of violence (batch 3)

zoomequipd:
# Description
This is 3 of 3 PRs to stay under the max rule size for test rules

Add multiple rules designed to detect acts of violence


zoomequipd:
no longer needed

PR #4211, opened 2026-03-17
Add detection rule for automobile assistance impersonation

peterdj45:
# Description

This rule detects messages impersonating automobile associations offering vehicle emergency kits or roadside assistance from untrusted senders.

# Associated samples
- https://platform.sublime.security/messages/502f373364d6a1c6c115a4188ed3deeb4caa66bbdc0ada168983bf66f6bc48cb

## Associated hunts
- https://platform.sublime.security/messages/hunt?huntId=019cfa70-04c0-743a-8767-912edc14688e


peterdj45:
Looks good in test rules!

L90D Shared EML: https://platform.sublime.security/messages/hunt?huntId=019db1cd-6ccc-7805-8336-f372c1ef2e88

detection-rules/brand_impersonation_aaa.yml
```diff
@@ -0,0 +1,50 @@
...
+                      '(car|vehicle|motor|driver|emergency|road.?side|break.?down|assist|save|discount|complimentary|free).{0,10}kit'
+  )
+  // and the sender is not from high trust sender root domains
+  and (
+    (
+      sender.email.domain.root_domain in $high_trust_sender_root_domains
+      and not headers.auth_summary.dmarc.pass
+    )
+    or sender.email.domain.root_domain not in $high_trust_sender_root_domains
+  )
```
IndiaAce:
```suggestion
  and not (
    sender.email.domain.root_domain in $high_trust_sender_root_domains
    and coalesce(headers.auth_summary.dmarc.pass, false)
  )
```
do we want to just use this "common snippet" for high trust sender domains that pass auth? [ran a hunt](https://hunt.limeseed.email/hunts/57e7f20d-fc80-4c49-9a85-711c56ebf759) to show the diff if we used the above logic... it returned 0 results so this should be fine.

detection-rules/brand_impersonation_aaa.yml
```diff
@@ -0,0 +1,50 @@
...
+          'RAC', // UK Royal Automobile Club
+          'RAA', // Australia Royal Automotive Assoc.
+          'CAA', // Canadian Automobile Assoc.
+          'BCAA', // BC Automobile Assoc.
+          'AMA', // Alberta Motor Assoc.
+        )
+    )
+  )
+  and regex.icontains(body.current_thread.text,
+                      '(car|vehicle|motor|driver|emergency|road.?side|break.?down|assist|save|discount|complimentary|free).{0,10}kit'
```
IndiaAce:
```suggestion
                      '(?:car|vehicle|motor|driver|emergency|road.?side|break.?down|assist|save|discount|complimentary|free\b).{0,10}kit'
```
noticed an fp here for "freedom kit" 
https://platform.sublime.security/messages/503974428448b3425551aabe94685925b4038d87cb6c668890904e5c691761eb?preview_id=019db69a-635e-799e-af57-db254f87bb9d

IndiaAce (changes requested):
Few notes of feedback, but I think overall this rule is awesome. One thing I noticed here [this is an FP](https://platform.sublime.security/messages/503b35745b6e13ef8e899d19a0b01b0ba530f0d89669ebb0dbb23a03cbc95083?preview_id=019db69a-3f53-781b-a160-6a030a62c9d1) but it's kinda tragic if you look we're incorrectly extracting an org from "Amapiano kit" as AMA which is... a bit lol but it's also got "Free afrobeat kits" selling samples. I think we can ignore this but figured I'd share. 

I'm not gonna remove from `review-needed` because if you're aligned with these changes we don't need to let this sit any longer, we should get this out the door! The telemetry is INSANE man... nice find.

detection-rules/brand_impersonation_aaa.yml
peterdj45: no objection to that!

detection-rules/brand_impersonation_aaa.yml
peterdj45: ahh good catch. def should be a non-capturing group, and I agree `free` should have a word boundary too

IndiaAce (approved):

PR #4236, opened 2026-03-20
Add detection rule for mismatched RAR file extensions

peterdj45:
This rule detects RAR files with mismatched extensions, indicating potential evasion tactics.

# Description

# Associated samples

- Sample 1
- Sample 2

## Associated hunts

- Hunt 1

# Screenshot (insights)


IndiaAce:

PR #4242, opened 2026-03-20
Create detection rule for ZoomInfo impersonation

cybher0808:
# Description
Detects messages impersonating ZoomInfo that contain financial payment requests, BEC language, or invoice viewing requests. The rule identifies both genuine new conversations and fake message threads from potentially suspicious senders.

# Associated samples
- [Sample 1](https://platform.sublime.security/messages/50324e4bf152a6ce5d0a54c07a8600ae0235ae7f6fb1d32b757b8edd6b317b54?preview_id=019d012a-5b37-7768-b36e-083162cafead)

# Associated hunts
- [Hunt 365D](https://platform.sublime.security/messages/hunt?huntId=01a0012e-4693-7199-a1b5-02590aa2b04a)


detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,57 @@
...
+  )
+  and regex.icontains(body.html.raw, 'zoominfo.com')
+  
+  // ml.nlu classifying emails based malicious contents
+  and (
+    any(ml.nlu_classifier(body.current_thread.text).entities,
+        .name == "financial" and .text == "payments"
+    )
+    or any(ml.nlu_classifier(body.current_thread.text).intents,
+           .name == "bec" and .confidence in ("medium")
```
peterdj45:
```suggestion
           .name == "bec" and .confidence in ("medium", "high")
```

just curious if there's a reason this is scoped to medium confidence matches? I think we can include high confidence as well

https://platform.sublime.security/messages/hunt?huntId=019d1d17-08d7-7ede-956c-9b0724545ee5

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,57 @@
+name: "Brand impersonation: ZoomInfo with financial request language"
+description: "Detects messages impersonating ZoomInfo that contain financial payment language or BEC indicators, combined with fake thread characteristics such as reply/forward subjects without proper email headers or no previous conversation history."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and length(attachments) >= 0
```
peterdj45:
```suggestion
  and length(attachments) > 0
```
in this condition, `>=` means greater than OR equal 0. So in effect, every message will be flagged regardless of attachment count.

I assume the intent was to require presence of an attachment? If so, couple of samples to consider without attachments:

- https://platform.sublime.security/messages/502b3db63824ea65a4d02ebfd6841ddd2ce9ba33563706f4443a82048a7ce5b5
- https://platform.sublime.security/messages/502be2ec1e66e3204fb921ab086a3e63e8a2657e9b17fd951286dd5e0088ae50

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,57 @@
...
+    or (
+      (length(headers.references) == 0 or headers.in_reply_to is null)
+      and (
+        subject.is_reply
+        or subject.is_forward
+        or length(body.previous_threads) > 0
+      )
+    )
+  )
+  and regex.icontains(body.html.raw, 'zoominfo.com')
```
peterdj45:
```suggestion
  and (
  strings.icontains(body.html.raw, 'zoominfo.com')
  or strings.icontains(sender.display_name, 'zoominfo')
  )
```

we could also check for `zoominfo` in the sender display name, would catch samples like this one:
- https://platform.sublime.security/messages/5038473456f084d4cb985a48dd66384301bc3f5a435bbdd38d63f5c218a14495

also can change `regex.icontains` to `strings.icontains` to help with performance

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,57 @@
...
+    length(body.previous_threads) == 0
+    // or is a fake thread
+    or (
+      (length(headers.references) == 0 or headers.in_reply_to is null)
+      and (
+        subject.is_reply
+        or subject.is_forward
+        or length(body.previous_threads) > 0
+      )
+    )
```
peterdj45, line 18:
```suggestion
    )
    // undisclosed or invalid recipients
    or any(recipients.to, .email.domain.valid == false)
```
the original sample was sent to undisclosed recipients with the victim BCC'ed, which is another signal we could lean into to help with edge cases, such as the threat actor replying to their own message.

FWIW we haven't seen that with this particular campaign yet, but it tripped up some of our BEC rules in the past

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,57 @@
...
+           .name == "bec" and .confidence in ("medium")
+    )
+    or any(ml.nlu_classifier(body.current_thread.text).topics,
+           .name == "Payment Information" and .confidence in ("high")
+    )
+    or any(ml.nlu_classifier(body.current_thread.text).topics,
+           .name == "Request to View Invoice"
+           and .confidence in ("medium", "high")
+    )
+  )
```
peterdj45, line 39:
```suggestion
   or any(ml.nlu_classifier(body.current_thread.text).topics,
          (.name == "Payment Information" and .confidence in ("high"))
          or (
            .name == "Request to View Invoice"
            and .confidence in ("medium", "high")
          )
   )
```
nit, but we can also merge both topic checks into one condition :)

peterdj45 (changes requested):
left some feedback on a few things, otherwise this looks fairly good overall. happy to jam on this if you'd like!

detection-rules/zoominfo_request_communication.yml
cybher0808: Honestly, I did not put high for confience which I should have since we can possibly gain further attraction for there are more matches. AGREE!

detection-rules/zoominfo_request_communication.yml
cybher0808: Smart choice to do to make it more simpler to read and it falls under the same match topic.

detection-rules/zoominfo_request_communication.yml
cybher0808: Okay, I had to look over this stanza that was added to the top portion to understand why it was placed. I had to read a few times to get it but now I see the correlation. I guess this is where we have to consider the current status of the email samples and eventually if the campaign takes a turn then we are able to shift our conditions to possibly meet what it may match. (my own thinking here - thinking out loud)

cybher0808:
Telem in test-rules shows a small amount of samples - opened to have this reviewed

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,64 @@
...
+        subject.is_reply
+        or subject.is_forward
+        or length(body.previous_threads) > 0
+      )
+    )
+    // undisclosed or invalid recipients
+    or any(recipients.to, .email.domain.valid == false)
+  )
+  and (
+    strings.icontains(body.html.raw, 'zoominfo.com')
```
IndiaAce: Is there a reason we're checking html.raw for a link instead of checking any body links? Just curious, did you run into FPs or something? I couldn't see anything specifically pointing to this in the Notion doc. I think we should spare the html.raw check as it's pretty expensive comparatively, especially if we're looking for a URL.

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,64 @@
...
+    or any(ml.nlu_classifier(body.current_thread.text).intents,
+           .name == "bec" and .confidence in ("medium", "high")
+    )
+    or any(ml.nlu_classifier(body.current_thread.text).topics,
+           .name == "Payment Information" and .confidence in ("high")
+           or (
+             .name == "Request to View Invoice"
+             and .confidence in ("medium", "high")
+           )
+    )
```
IndiaAce, line 38:
```suggestion
    or any(ml.nlu_classifier(body.current_thread.text).topics,
           .name in ("Payment Information", "Request to View Invoice")
           and .confidence != "low"
    )
```
I suggest consolidating this a bit further here for readability.

detection-rules/zoominfo_request_communication.yml
```diff
@@ -0,0 +1,64 @@
...
+  )
+  // negate when NLU confidently classifies as benign
+  and not any(ml.nlu_classifier(body.current_thread.text).intents,
+              .name == "benign" and .confidence == "high"
+  )
+  // negate highly trusted sender domains unless they fail DMARC authentication
+  and (
+    (
+      sender.email.domain.root_domain in $high_trust_sender_root_domains
+      and not headers.auth_summary.dmarc.pass
```
IndiaAce:
```suggestion
      and not coalesce(headers.auth_summary.dmarc.pass, false)
```
I suggesting enforcing auth on these!

IndiaAce (changes requested):
Couple of suggestions here, removing from R4R for the time being.

cybher0808:
Need to let this PR sit a bit longer before review. Sharing in mode shows only recent results, just readdressed this rule today 4/20. ‼️‼️

IndiaAce:
I'm working off assumptions here based on your message but I'm going to remove the `review-needed` label until you feel like you've gathered sufficient telemetry, feel free to add it back whenever!

cybher0808:
Reviewed [Mode](http://app.mode.com/sublimesecurity/reports/d3ba05360b64/runs/26b0911f1e4e)/[Multi-hunt](https://hunt.limeseed.email/hunts/fc1fc850-d562-463f-9e16-a5a437991345) for the results and re-ran hunt.

cybher0808:
Looking for a better option to swap around a few stanzas to make this rule stronger to catch malicious samples.

IndiaAce:

PR #4276, opened 2026-03-30
Negate SharePoint shares with Mimecast rewrites

peterdj45:
# Description
Add condition to negate legitimate SharePoint file shares with Mimecast rewrites.

# Associated samples
- https://platform.sublime.security/messages/5039f8d6156b22dd7df920bfc925fdef3d40ceaca2a31062853b53b673c57567

# Associated hunts
- https://platform.sublime.security/messages/hunt?huntId=019d9adf-956d-7896-b400-86de0ae33db5


peterdj45:
telemetry looks good: 

- L90D Shared EML hunt for net new negations: https://platform.sublime.security/messages/hunt?huntId=019d9adf-956d-7896-b400-86de0ae33db5

multi hunt results in ESC-9700

detection-rules/impersonation_microsoft_credential_theft.yml
```diff
@@ -114,6 +114,26 @@ source: |
...
+            ),
+            .href_url.domain.root_domain == "mimecastprotect.com"
+            and (
+              any(.href_url.query_params_decoded["domain"], . == "sharepoint.com")
+              or (
+                any(.href_url.query_params_decoded["domain"],
+                    strings.parse_domain(.).tld == "ms"
+                )
+              )
+            )
```
zoomequipd: you can avoid the second loop by calling strings.parse_domain a second time. 

also added in mimecast.com to match the logic in other rules

```suggestion
            .href_url.domain.root_domain in (
              "mimecastprotect.com",
              "mimecast.com"
            )
            and any(.href_url.query_params_decoded["domain"],
                    strings.parse_domain(.).tld == "ms"
                    or strings.parse_domain(.).root_domain == "sharepoint.com"
            )
```

zoomequipd (changes requested):

detection-rules/impersonation_microsoft_credential_theft.yml
```diff
@@ -114,6 +114,25 @@ source: |
     or sender.email.domain.root_domain not in $org_domains
   )
   
+  // negate sharepoint file shares with mimecast rewrites
```
IndiaAce, line 117: Just so I'm clear here; are we accepting that likely-malicious sharepoint shares will be negated? Important to note that we can't say for a certainty if these messages are malicious or not due to the nature of sharepoint, but we can make strong guesses. But we accept the FNs (for this rule) because they're matching the more specific sharepoint malicious rules? 
[Here's a hunt I ran](https://platform.sublime.security/messages/hunt?huntId=019daaf2-121d-744e-9c08-09e83065e3c6) showing all messages in sws that will be negated by this change. Just making sure we're aligned on this before reviewing.

peterdj45: Sorry, I should have clarified - Yes, these are legit SharePoint shares so they shouldn't be firing the `Brand Impersonation` rules. IMO, the more specific `Service Abuse` rules should fire on the FNs. I think the overall trade off is pretty minor here, since we're only negating if all the links lead to sharepoint.com or the various `.ms` domains, but I'm happy to discuss this further if you'd like!

IndiaAce: Thanks Peter, will review with this alignment.

IndiaAce (approved):

IndiaAce:
lgtm

zoomequipd (approved):

PR #4285, opened 2026-03-31
Create self_sender_display_name_email_in_subject.yml

D-Bolton:
# Description

Detects messages where the sender emails themselves with both their email address and display name present in the subject line, while the email address differs from the display name.

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/50394b86201416b8f57418f85c1ccf34c8b39202a83ac3a83ef6d2a286b2178f)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019d96a1-aca8-7046-838b-85fcca38da0d)
- Multi-hunts in ESC-9460


detection-rules/self_sender_display_name_email_in_subject.yml
```diff
@@ -0,0 +1,34 @@
...
+description: "Detects messages where the sender emails themselves with both their email address and display name present in the subject line, while the email address differs from the display name. This pattern is commonly used in social engineering attacks to create false legitimacy or test delivery mechanisms."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  // self sender
+  and (
+    length(recipients.to) == 1
+    and length(recipients.cc) == 0
+    and sender.email.email == recipients.to[0].email.email
```
peterdj45: nit, but we typically use map to achieve this in other rules

```suggestion
    and sender.email.email in map(recipients.to, .email.email)
```

peterdj45 (changes requested):
Found some FPs in test rules, primarily messages sent via Hearsay Social (social media marketing platform). Example: https://platform.sublime.security/messages/50471e1a18de024481af14a3b2e7f8c8fe9fff30c14456f355e24c6d3234ae4c

Could maybe negate these via presence of `X-HearsaySocial` headers? or the message ID, which is `hearsaysystems.com`. Happy to jam on this if you'd like!

D-Bolton:
> Found some FPs in test rules, primarily messages sent via Hearsay Social (social media marketing platform). Example: https://platform.sublime.security/messages/50471e1a18de024481af14a3b2e7f8c8fe9fff30c14456f355e24c6d3234ae4c
> 
> Could maybe negate these via presence of `X-HearsaySocial` headers? or the message ID, which is `hearsaysystems.com`. Happy to jam on this if you'd like!

I tightened the rule up more. I have multi-hunts in the ESC. Let me know what you think. Thanks!

detection-rules/self_sender_french_copypaste_instructions_suspicious_domains.yml
```diff
@@ -0,0 +1,35 @@
+name: "Self-sender with French copy/paste instructions and suspicious domains"
```
peterdj45:
```suggestion
name: "Self-sender with copy/paste instructions and suspicious domains (French/Français)"
```
another nit, but this seems to be the naming convention used in our other French-specific rules

detection-rules/self_sender_french_copypaste_instructions_suspicious_domains.yml
```diff
@@ -0,0 +1,35 @@
...
+source: |
+  type.inbound
+  // self sender
+  and (
+    length(recipients.to) == 1
+    and length(recipients.cc) == 0
+    and sender.email.email in map(recipients.to, .email.email)
+  )
+  and strings.contains(subject.subject, sender.email.email)
+  and strings.contains(subject.subject, sender.display_name)
```
peterdj45:
```suggestion
  and strings.icontains(subject.subject, sender.email.email)
  and strings.icontains(subject.subject, sender.display_name)
```
thoughts on making this case-insensitive?

detection-rules/self_sender_french_copypaste_instructions_suspicious_domains.yml
```diff
@@ -0,0 +1,35 @@
...
+  and sender.email.email != sender.display_name
+  // copy
+  and strings.contains(body.current_thread.text, 'copier')
+  // paste
+  and strings.contains(body.current_thread.text, 'coller')
```
peterdj45:
```suggestion
  and strings.icontains(body.current_thread.text, 'copier')
  // paste
  and strings.icontains(body.current_thread.text, 'coller')
```
same as above

peterdj45 (approved):
LGTM! left you some non-blocking feedback

detection-rules/self_sender_french_copypaste_instructions_suspicious_domains.yml
```diff
@@ -0,0 +1,34 @@
...
+  and strings.icontains(subject.subject, sender.display_name)
+  and sender.email.email != sender.display_name
+  // copy
+  and strings.icontains(body.current_thread.text, 'copier')
+  // paste
+  and strings.icontains(body.current_thread.text, 'coller')
+  and (
+    strings.contains(body.current_thread.text, '.pages.dev')
+    or strings.contains(body.current_thread.text, '.web.app')
+  )
```
IndiaAce, line 23: Non blocking, I wonder if we want to consider either swapping this to be a check on body links `*pages.dev` and then look for things like suspicious sld's or something. Just to consider for the future, this is the most "specific" part of the rule I feel like as the vector changes we might be introducing a foot-gun here. Food for thought!

IndiaAce (approved):
Approved with non-blocking feedback.

detection-rules/self_sender_french_copypaste_instructions_suspicious_domains.yml
D-Bolton: Good idea! I might create a broader rule for test rules to track this actor.

PR #4340, opened 2026-04-10
Create attachment_pdf_split_qr_code_yara

zoomequipd:
# Description

Add detection for an observed evasion technique in which the QR code is "split" and the two images are placed next to eachother to form a complete QR code. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/503b26fd09676b637784f45b68faa6c339ed2b635c8c5f5bd05d5b229ed86534?preview_id=019d3076-bf19-7a68-95f7-a90b5cbf2407)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019d78d5-389f-7b8a-9c43-93ca68a99c50)


zoomequipd:
low volume telemetry, looks like this technique is no longer being used

IndiaAce (approved):

PR #4349, opened 2026-04-14
Update generic_pdf.yar

keaton-sublime:
# Description
- Adding additional rule for the observed image lures in the PDFs associated with this CVE.
- Also updating the acro_js function rule to match one that we've seen used maliciously and in testing.


keaton-sublime:
moving this to review needed/ready for review. There's no matching MQL rules for these two so we don't really have anything to worry about yet.  I'll follow up with the MQL rule after we have some test data.

markmsublime (approved):

PR #4353, opened 2026-04-15
Update generic_pdf.yar

keaton-sublime:
# Description
Updating the yara here to include the three "eval" encodings we might see. 

This rule is trying to match base64 encoded data in objects, not necessarily inside the javascript object itself.  Renamed the rule name and the string var to make more sense.


markmsublime (approved):

keaton-sublime:
this one is closed in lieu of PR 4356

PR #4357, opened 2026-04-15
Add YARA rule compilation validation to CI

aidenmitchell:
## Summary
- Adds a CI step that compiles all YARA rules using `yarac` to catch syntax errors before merge
- Installs `yara` via apt and validates each `.yar` file in the `yara/` directory
- Uses GitHub Actions `::error` annotations for inline PR feedback on failures
- Validates all rules before failing so all errors are surfaced at once

## Test plan
- [x] Verified all 8 existing YARA rules compile successfully locally
- [ ] Confirm the new step runs in CI on this PR


.github/workflows/rule-validate.yml
```diff
@@ -215,6 +215,24 @@ jobs:
 
           echo "All batches submitted successfully!"
 
+      - name: Validate YARA rules
+        if: github.event_name != 'issue_comment'
+        run: |
+          sudo apt-get install -y -qq yara > /dev/null 2>&1
```
cameron-dunn-sublime: Req: pin `yara` version

.github/workflows/rule-validate.yml
```diff
@@ -215,6 +215,24 @@ jobs:
...
+          failed=0
+          for f in yara/*.yar; do
+            echo "Validating $f..."
+            if ! yarac "$f" /dev/null; then
```
cameron-dunn-sublime, line 227: This may produce different results than the parsing and validation we do within the platform. We can talk offline for more detail, but I know we've seen subtle differences between our validation in feeds vs. supported features in Strelka.

cameron-dunn-sublime (approved):

markmsublime (approved):

markmsublime (approved):

keaton-sublime (approved):

PR #4358, opened 2026-04-16
Test NLU Embedding Function

dlynch-sublime:
# Description
Creates a basic rule to exercise nlu_embedding at scale. This rule is guaranteed not to trigger (at most, we'll have four vectors), but will give us data on the embedding endpoint performance at scale before we begin relying on the function in attack score and stratum.


dlynch-sublime:
Closing PR, as the test this rule was for is completed.

PR #4382, opened 2026-04-22
[yara] cleanup split QR and add more

zoomequipd:
# Description

- Organize split QR code detection within the yara rules
- add three new QR code dimensions - 410,450,490

# Associated samples
Samples have been observed outside of Sublime's visiblity


keaton-sublime (approved):
"i like that", is what I said when reading this. lgtm

PR #4387, opened 2026-04-22
Add PCI credit card DLP rules

morriscode:
## Summary
- Adds DLP rules for Visa, Mastercard, American Express, and Discover credit card number detection in outbound email
- Adds a combined any-network rule (`dlp_pci_us_credit_card_any`) for broad PCI DLP coverage
- Covers card numbers in message body and text-parseable attachments (PDF, Office docs, CSV, etc.)
- No LUHN validation — regex-based pattern matching only

## Test plan
- [x] Backtest hunts run for all 5 rules over last 90 days
- [x] Review results for false positives before enabling block action


olayemioyebode-wq:
I tested all the codes and found only one that needs to be corrected: Mastercard 2-series BINs (2221–2720)

With some help from Claude, I think this is the right regex
\b(?:222[1-9]|22[3-9]\d|2[3-6]\d{2}|27[01]\d|2720)[\s-]?\d{4}[\s-]?\d{4}[\s-]?\d{4}\b

morriscode:
Thank you @olayemioyebode-wq 
Updating rules now

IndiaAce (approved):

PR #4422, opened 2026-04-29
Update link_romance_suspicious.yml

JFarina5:
# Description

Updating rule to tag links with a tld of `.ru`

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/505c30c50dd668b1c0f6091f7fbb268c4b5b6416b889dcab983076ad3b2d7b1b?preview_id=019dd7d4-184b-7974-8ff9-c3f66d53ffd8)

## Associated hunts

- [L30D Net New Hunt](https://platform.sublime.security/messages/hunt?huntId=019de364-c249-7b73-b031-0614ec8b33a8)
- [L7D Multi-Hunt](https://hunt.limeseed.email/hunts/d3ec56bd-feda-4436-ad83-2a25f4112e20)

PR #4430, opened 2026-05-01
Update link_romance_suspicious.yml

JFarina5:
# Description

Adding additional sus TLDs based on PR #4422 

## Associated hunts

- [L30D Net New Hunt](https://platform.sublime.security/messages/hunt?huntId=019df34a-5894-7201-a309-3b92fefcadd0)

### DIFF
--- a/detection-rules/link_romance_suspicious.yml
+++ b/detection-rules/link_romance_suspicious.yml
@@ -12,7 +12,7 @@ source: |
   and 1 of (
     any(body.links, network.whois(.href_url.domain).days_old < 30),
     any(body.links, .href_url.domain.root_domain in $url_shorteners),
-    any(body.links, .href_url.domain.tld in ('ru')),
+    any(body.links, .href_url.domain.tld in ('ru', 'app', 'digital', 'click')),
     any(headers.reply_to, network.whois(.email.domain).days_old < 30)
   )
   and 0 < length(distinct(body.links,


JFarina5:
Net new matches are looking good with the latest changes, seems to be grabbing a good amount of FNs as well. Hunt in description has been updated. Marking r4r.

detection-rules/link_romance_suspicious.yml
```diff
@@ -12,7 +12,7 @@ source: |
   and 1 of (
     any(body.links, network.whois(.href_url.domain).days_old < 30),
     any(body.links, .href_url.domain.root_domain in $url_shorteners),
-    any(body.links, .href_url.domain.tld in ('ru')),
+    any(body.links, .href_url.domain.tld in ('ru', 'app', 'digital', 'click')),
```
IndiaAce, line 15: These are all in https://github.com/sublime-security/static-files/blob/main/suspicious_tlds.txt thoughts on just swapping the check for sus tlds?

IndiaAce (changes requested):

detection-rules/link_romance_suspicious.yml
JFarina5: yeahhhh the problem I ran into was samples from `redditmail` that use `.link` and some yahoo samples that use `.me`

IndiaAce: alas

IndiaAce (approved):
Approving but we should talk about cleaving the "sexual" content out of the "romance" here... some of these matches should not be considered "romance" by any stretch of the imagination and I think it makes sense to upright this. This is outside the scope of your change so let's look at fixing it outside of this?

PR #4431, opened 2026-05-01
Update abused_payoneer_callback.yml

IndiaAce:
Hopefully resolving some Rule Validation failures with this? Unless comments like this are in a bunch of rules which they might be...


IndiaAce:
https://github.com/sublime-security/sublime-rules/pull/4432

PR #4457, opened 2026-05-07
Update callback_phishing_calendar_invite.yml

D-Bolton:
# Description
HTML can sometimes throw off NLU. This change allows us to parse the HTML in the calendar description to get the display_text for callback_scam intent. Will check `Request to View Invoice` in another PR to make things eaiser on the reviewer. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/4fad2f1e46ce8a25608936e6d1ecc19d4f9491a93f231275d0040a7923089e42)

## Associated hunts

Regression callback scam just html.parse | https://platform.sublime.security/messages/hunt?huntId=019e0346-2b68-7d8e-b2c6-999becdcfa3a
-- | --
Net new callback scam use html.parse | https://platform.sublime.security/messages/hunt?huntId=019e0347-72e9-776f-bf75-3704e5a69746
Net new callback scam use html.parse or raw | https://platform.sublime.security/messages/hunt?huntId=019e0359-a8e9-7948-8bd4-1f5c50293936
Regression callback scam use html.parse or raw | https://platform.sublime.security/messages/hunt?huntId=019e0365-18cc-7cf0-907d-9746db1c04a7

Net new multi-hunt: https://hunt.limeseed.email/hunts/ba33573b-ef1d-4141-8bb2-ac3651b1cce9

### DIFF
--- a/detection-rules/callback_phishing_calendar_invite.yml
+++ b/detection-rules/callback_phishing_calendar_invite.yml
@@ -15,8 +15,9 @@ source: |
                         any(ml.nlu_classifier(.description).intents,
                             .name == "callback_scam"
                         )
-                        or any(ml.nlu_classifier(strings.parse_html(.description).display_text).intents,
-                            .name == "callback_scam"
+                        or any(ml.nlu_classifier(strings.parse_html(.description).display_text
+                               ).intents,
+                               .name == "callback_scam"
                         )
                         or (
                           any(ml.nlu_classifier(.description).topics,
--- a/detection-rules/callback_phishing_calendar_invite.yml
+++ b/detection-rules/callback_phishing_calendar_invite.yml
@@ -15,6 +15,9 @@ source: |
                         any(ml.nlu_classifier(.description).intents,
                             .name == "callback_scam"
                         )
+                        or any(ml.nlu_classifier(strings.parse_html(.description).display_text).intents,
+                            .name == "callback_scam"
+                        )
                         or (
                           any(ml.nlu_classifier(.description).topics,
                               .name == "Request to View Invoice"


detection-rules/callback_phishing_calendar_invite.yml
```diff
@@ -15,6 +15,10 @@ source: |
                         any(ml.nlu_classifier(.description).intents,
                             .name == "callback_scam"
                         )
+                        or any(ml.nlu_classifier(strings.parse_html(.description).display_text
```
zoomequipd, line 18: i'm not too familiar with the `strings.parse_html` function, what does it return if the .description is not html?

D-Bolton: When there is no HTML it just passes through the description.

zoomequipd: in that case instead of an or statement, can we just replace the existing nlu call to include the `strings.parse_html` function?

zoomequipd (changes requested):

detection-rules/callback_phishing_calendar_invite.yml
D-Bolton: We still need the or statement. See this [hunt](https://platform.sublime.security/messages/hunt?huntId=019e0346-2b68-7d8e-b2c6-999becdcfa3a) for what we would miss. It passes the text through but doesn't pass new lines causing the nlu to fail. [See](https://platform.sublime.security/messages/4fb2e0fcc836af27e999d61407d35ef6607448789e6182577e4e6824d33b3c87?preview_id=019a6f21-27dc-7aea-8bf9-2027fcd90d9d) this example.

zoomequipd: ahhh, the rule as is works around these nuances.  we can follow up with the internal teams to address those behaviors.  the rule itself works as intended.

zoomequipd (approved):

PR #4513, opened 2026-05-20
Add first_name/last_name concat matching to org_vips sender rules

IndiaAce:
# Description
Add alternative name matching logic to org_vips sender-based rules to handle cases where
display_name is stored as "Lastname, Firstname" instead of "Firstname Lastname".
Uses `strings.concat(.first_name, " ", .last_name)` and `strings.concat(.last_name, ", ", .first_name)`
as additional `or` conditions inside existing `any($org_vips, ...)` blocks.

This is a test rule deployment to assess impact magnitude.

## Affected rules
- `impersonation_vip_bec_loose.yml`
- `vip_impersonation.yml`
- `impersonation_vip_urgent_request.yml`
- `impersonation_vip_invoicing_request.yml`
- `impersonation_vip_w2_request.yml`
- `sender_contains_org_vip.yml` (insight)

# Associated samples
N/A - validation only (no TP canonical available)

## Associated hunts
TBD - will be run after test rule deployment

### DIFF
--- a/insights/sender/sender_contains_org_vip.yml
+++ b/insights/sender/sender_contains_org_vip.yml
@@ -3,11 +3,7 @@ type: "query"
 source: |
   type.inbound
   and any($org_vips,
-          (
-            strings.icontains(sender.display_name, .display_name)
-            or strings.icontains(sender.display_name, strings.concat(.first_name, " ", .last_name))
-            or strings.icontains(sender.display_name, strings.concat(.last_name, ", ", .first_name))
-          )
+          strings.icontains(sender.display_name, .display_name)
           and not sender.email.domain.root_domain in $high_trust_sender_root_domains
           and not sender.email.domain.root_domain in $org_domains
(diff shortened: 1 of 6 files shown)


insights/sender/sender_contains_org_vip.yml
```diff
@@ -3,7 +3,11 @@ type: "query"
...
   and any($org_vips,
-          strings.icontains(sender.display_name, .display_name)
+          (
+            strings.icontains(sender.display_name, .display_name)
+            or strings.icontains(sender.display_name, strings.concat(.first_name, " ", .last_name))
+            or strings.icontains(sender.display_name, strings.concat(.last_name, ", ", .first_name))
+          )
           and not sender.email.domain.root_domain in $high_trust_sender_root_domains
           and not sender.email.domain.root_domain in $org_domains
           and headers.auth_summary.dmarc.pass
```
rw-access, line 9: Why are these inside the `any($org_vips)` check?

PR #4538, opened 2026-05-26
Update generic_pdf.yar

keaton-sublime:
# Description
Matching the image sizes within encrypted PDFs, which allows us to match the lure they're using without needed to decrypt anything. The stream content of the images is too variable to be useful in this case, so we use the sizes. Working well with these samples in my test data and in VT. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/500f618ef0458247e5d0ec2768d199b9aa620e8e7d94ba268120352c11967deb?preview_id=019c53b0-46c9-7e44-8722-0e8a4dfccb00)

### DIFF
--- a/yara/generic_pdf.yar
+++ b/yara/generic_pdf.yar
@@ -297,7 +297,7 @@ rule enc_pdf_image_sizes {
 	meta:
 		author      = "kyle eaton"
 		date        = "2026-05-26"
-		description = "PDF image sizes that correlate with the images used in an encrypted PDF lure"
+		description = "matches PDF image sizes that correlate with the images used in an encrypted PDF lure"
 	strings:
 		$header = { 25 50 44 46 2D 31 2E }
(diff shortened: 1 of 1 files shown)


keaton-sublime:
Marking as review needed/ready for review - this yara rule has no accompanying MQL rule.

markmsublime (approved):

yara/generic_pdf.yar
```diff
@@ -293,3 +293,17 @@ rule pdf_cve_2026_34621_observed_lures {
 		$header at 0 and any of ($img_*)
 }
 
+rule enc_pdf_image_sizes {
+	meta:
+		author      = "kyle eaton"
+		date        = "2026-05-26"
+		description = "PDF image sizes that correlate with the images used in an encrypted PDF lure"
```
IndiaAce:
```suggestion
		description = "matches PDF image sizes that correlate with the images used in an encrypted PDF lure"
```

IndiaAce (changes requested):

IndiaAce (approved):

PR #4539, opened 2026-05-26
Add detection rule for Square QR code abuse

peterdj45:
# Description

This rule detects messages from Square's marketing domain that contain suspicious QR codes redirecting to non-legitimate domains.

# Associated samples
- https://platform.sublime.security/messages/50751d482a8961255bc57c28d554ad774417a9ceae7f4377bddb8204d73ea980
- https://platform.sublime.security/messages/50754e8e71a3ba4352d56e4b05466dc1703c5da13f6eb105caf0a7544902bf35
- https://platform.sublime.security/messages/50755e47ddbbe68d36f8dedc67a867a3e2186e00bae4929ad6a0af1e5584f923

## Associated hunts
- https://platform.sublime.security/messages/hunt?huntId=019e6564-251d-7ecb-9d34-89cc5e2cef5b

### DIFF
--- a/detection-rules/service_abuse_square_QR_code.yml
+++ b/detection-rules/service_abuse_square_QR_code.yml
@@ -10,8 +10,14 @@ source: |
                  // ignore square's own free website hosting service
                  .url.domain.root_domain != "square.site"
           ),
-          .url.domain.root_domain in $self_service_creation_platform_domains
-          or .url.domain.root_domain in $free_file_hosts
+          (
+            .url.domain.root_domain in $self_service_creation_platform_domains
+            or .url.domain.domain in $self_service_creation_platform_domains
+          )
+          or (
+            .url.domain.root_domain in $free_file_hosts
+            or .url.domain.domain in $free_file_hosts
+          )
   )
 
 attack_types:
--- a/detection-rules/service_abuse_square_QR_code.yml
+++ b/detection-rules/service_abuse_square_QR_code.yml
@@ -1,0 +1,26 @@
+name: "Service abuse: Square marketing with suspicious QR code"
+description: "Detects messages from Square's marketing domain containing QR codes that redirect to self-service creation platforms, file sharing services, or image hosting services rather than legitimate Square domains."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and sender.email.domain.domain == "squaremktg.com"
+  and beta.scan_qr(file.message_screenshot()).found
+  and any(filter(beta.scan_qr(file.message_screenshot()).items,
+                 // ignore square's own free website hosting service
+                 .url.domain.root_domain != "square.site"
+          ),
+          .url.domain.root_domain in $self_service_creation_platform_domains
+          or .url.domain.root_domain in $free_file_hosts
+  )
+
+attack_types:
+  - "Credential Phishing"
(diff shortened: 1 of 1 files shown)


peterdj45:
hunts look good, active campaign observed across several environments.

multi-hunt results in ESC-14160

MSAdministrator:
@peterdj45 the samples you provided don't match the current rule but the inclusion of the `or .url.domain.root_domain == "image2url.com"` seems to be the match needed

detection-rules/service_abuse_square_QR_code.yml
```diff
@@ -0,0 +1,27 @@
...
+source: |
+  type.inbound
+  and sender.email.domain.domain == "squaremktg.com"
+  and beta.scan_qr(file.message_screenshot()).found
+  and any(filter(beta.scan_qr(file.message_screenshot()).items,
+                 // ignore square's own free website hosting service
+                 .url.domain.root_domain != "square.site"
+          ),
+          .url.domain.root_domain in $self_service_creation_platform_domains
+          or .url.domain.root_domain in $free_file_hosts
```
MSAdministrator: Add `or .url.domain.root_domain == "image2url.com"` to match on the samples provided.

peterdj45: @MSAdministrator just added `image2url.com` to free file hosts, but I don't think the list has synced over yet

https://github.com/sublime-security/static-files/pull/829

MSAdministrator: Ahh, sweet - thanks for the clarification!

MSAdministrator (approved):
LGTM

detection-rules/service_abuse_square_QR_code.yml
zoomequipd: do you also need to check `.href_url.domain.domain`?

zoomequipd (changes requested):

detection-rules/service_abuse_square_QR_code.yml
peterdj45: good catch, yes we should. i'll fix that

zoomequipd (approved):

detection-rules/service_abuse_square_QR_code.yml
```diff
@@ -0,0 +1,33 @@
+name: "Service abuse: Square marketing with suspicious QR code"
+description: "Detects messages from Square's marketing domain containing QR codes that redirect to self-service creation platforms, file sharing services, or image hosting services."
+type: "rule"
+severity: "high"
+source: |
+  type.inbound
+  and sender.email.domain.domain == "squaremktg.com"
+  and beta.scan_qr(file.message_screenshot()).found
+  and any(filter(beta.scan_qr(file.message_screenshot()).items,
```
zoomequipd, line 13: needs the beta disclaimer thing
```suggestion
  //
  // This rule makes use of a beta feature and is subject to change without notice
  // using the beta feature in custom rules is not suggested until it has been formally released
  //
  and any(filter(beta.scan_qr(file.message_screenshot()).items,
```

zoomequipd (changes requested):

zoomequipd (approved):

PR #4541, opened 2026-05-26
Update impersonation_dhl.yml

MSAdministrator:
# Description

During a runner ping, we were notified that a few FPs came through so negating the sender tld for `.dhl` gTLD in rule.

There was no official escalation created but this will resolve the issue reported.

# Associated samples

- [Sample](https://platform.sublime.security/messages/50738be0578883a99098e23f6d82aebc4657d8a46e3acaf86cb7eefb3c43e634?preview_id=019e51e9-a695-7d74-87a2-238a2613d693)

## Associated hunts

- [Hunt negation between rule differences](https://platform.sublime.security/messages/hunt?huntId=019e65ad-4e45-77c5-9fa1-64c7acff932f)

### DIFF
--- a/detection-rules/impersonation_dhl.yml
+++ b/detection-rules/impersonation_dhl.yml
@@ -122,8 +122,8 @@ source: |
     'dhlgpi.com', // DHL Australia
     'dhlfreight-news.com'
   )
-  and (
-    sender.email.domain.tld not in ('dhl')
+  and not (
+    sender.email.domain.tld in ('dhl')
     and coalesce(headers.auth_summary.dmarc.pass, false)
   )
   and (
--- a/detection-rules/impersonation_dhl.yml
+++ b/detection-rules/impersonation_dhl.yml
@@ -120,7 +120,7 @@ source: |
     'tdhlaw.com', // dhl in domain but unrelated
     'hapibenefits.com', // DHL rewards program
     'dhlgpi.com', // DHL Australia
-    'dhlfreight-news.com',
+    'dhlfreight-news.com'
   )
   and sender.email.domain.tld not in ('dhl')
   and (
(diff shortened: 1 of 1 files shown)


detection-rules/impersonation_dhl.yml
```diff
@@ -122,6 +122,7 @@ source: |
     'dhlgpi.com', // DHL Australia
     'dhlfreight-news.com'
   )
+  and sender.email.domain.tld not in ('dhl')
```
zoomequipd: auth check?

zoomequipd (changes requested):

detection-rules/impersonation_dhl.yml
MSAdministrator: I don't think it's needed but can add -- mostly because they own that gTLD so abuse is less likely ? That's my overall opinion though - thoughts?

zoomequipd: it can still be spoofed, i think we need an auth check.

MSAdministrator: Okay, updated!

detection-rules/impersonation_dhl.yml
```diff
@@ -122,6 +122,10 @@ source: |
     'dhlgpi.com', // DHL Australia
     'dhlfreight-news.com'
   )
+  and (
+    sender.email.domain.tld not in ('dhl')
+    and coalesce(headers.auth_summary.dmarc.pass, false)
+  )
```
zoomequipd:
```suggestion
  and not (
    sender.email.domain.tld in ('dhl')
    and coalesce(headers.auth_summary.dmarc.pass, false)
  )
```

zoomequipd (approved):
-

PR #4545, opened 2026-05-27
Update generic_pdf.yar

keaton-sublime:
# Description
This is matching a subset of w9 pdfs which use a few specific images and objects within their PDFs.  We have coverage for their w9 PDF, but they also use up to 2 additional PDFs which we do not typically have coverage for. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/506c61b03c5f210dbc5ed08847d214a1fd2d0aa66795242e1cb23fefb083a651?preview_id=019e5119-3363-74e0-add8-03f797430b47)
- [Sample 2](https://platform.sublime.security/messages/50780354709d81b95e0dfef65eaf5d0df4cf3e36367b2573fd1a8e39ee04f920?preview_id=019e66ff-dc1b-7cfc-a0fb-d115830af423)

# Associated Hunt

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019e69f8-7c38-7930-badc-9d02bde98e12)

### DIFF
--- a/yara/generic_pdf.yar
+++ b/yara/generic_pdf.yar
@@ -307,3 +307,31 @@ rule enc_pdf_image_sizes {
 		and all of ($img*)
 		and @img1 < @img2
 }
+
+rule w9_pdf_invoice_images {
+	meta:
+		author      = "kyle eaton"
+		date        = "2026-05-15"
+		description = "matching images within invoices in a subset of w9 pdfs"
+	strings:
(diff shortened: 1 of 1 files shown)


keaton-sublime:
Marking as review needed/ready for review - This is only a yara rule with no matching MQL.

markmsublime (approved):
✅ Both rules are syntactically correct!

PR #4566, opened 2026-05-29
Update recon_short_generic_greeting.yml

zoomequipd:
# Description
Objective: Reduce FPs

1. Address bug in sender profile logic as it related to dmarc
2. simplify profile logic
3. reduce FPs for employee --> work message by comparing sender display_name to mailbox first_name and mailbox.last_name (handle for when sender display_name/mailbox elements are null)
4. add negation for common senders

see ESC for details


markmsublime (approved):
we spoke about this offline, aligned with these changes. great solution!

PR #4575, opened 2026-06-01
Update generic_pdf.yar

keaton-sublime:
Removing w9_pdf_invoice_images.
This is a noisier rule than expected. It seems to be an artifact of the skia/chrome -> pdf pipeline on certain web pages. We could tighten up the rule with something like (any -> all on L322), but again, it's not an artifact of the actor's creation. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/50784e57e4ae972eab556086995481edb1942dfb7461f0f66ea27a7c1e7130ca?preview_id=019e8434-fad1-7351-96aa-1d872c96ea7e)

### DIFF
--- a/yara/generic_pdf.yar
+++ b/yara/generic_pdf.yar
@@ -308,20 +308,6 @@ rule enc_pdf_image_sizes {
 		and @img1 < @img2
 }
 
-rule w9_pdf_invoice_images {
-	meta:
-		author      = "kyle eaton"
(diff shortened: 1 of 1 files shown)


keaton-sublime:
Removing yara rule - there was no matching MQL for this particular rule in the first place, so this is really just cleaning up the .yar files.

markmsublime (approved):

PR #4599, opened 2026-06-03
Update generic_pdf.yar

keaton-sublime:
# Description
This rule is looking for PDFs which match a couple of specific optional conditions: any of two /rect sizes, or any of two text mappings. 
In my testing this is quite solid, however, if it ends up being too wide, we can combine these conditions, requiring any of 2 AND any of the other two. 

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/506f43cd1ef23bf029dd8a332000f1a6549973419b857afa9623c6a077b0ee16?preview_id=019e3d77-4be9-75a0-97a9-b63ced392d70)
- [Sample 2](https://platform.sublime.security/messages/507aa1bf6495b1079bf523365fb8c343a35d20af0e9bf2b12a5f74dba501b82b?preview_id=019e8353-8a7e-7838-aa22-6c40b948cdee)

### DIFF
--- a/yara/generic_pdf.yar
+++ b/yara/generic_pdf.yar
@@ -352,7 +352,7 @@ rule pdf_eCheckLure_format {
 	meta:
 		author      = "kyle eaton"
 		date        = "2026-06-03"
-		description = "matching patterns used in the eCheckRun lures, including link rects and text content."
+		description = "Matching patterns used in the eCheckRun lures, including link rects and text content."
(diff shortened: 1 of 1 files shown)


keaton-sublime:
Marking as ready for review. This doesn't have a matching MQL rule and will be used for hunting.

zoomequipd (approved):

PR #4613, opened 2026-06-04
Fuzzy Graymail rule

dlynch-sublime:
# Description
Adds a rule to the core feed for identifying graymail using `beta.fuzzy_attack_score`


dlynch-sublime:
The fuzzy graymail rule is now ready for review, to be added to the core feed. We've had it running in test rules for a few weeks now, and are looking for this to be in the core feed to support the Graymail GA.

detection-rules/fuzzy_graymail_rule.yml
```diff
@@ -0,0 +1,8 @@
+name: "Fuzzy Graymail"
+description: "Exercises the beta.fuzzy_attack_score() function. Flags on Graymail verdicts."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and beta.fuzzy_attack_score().verdict == "graymail"
```
zoomequipd, line 11: add in line disclaimer for beta fucntions
```suggestion
  //
  // This rule makes use of a beta feature and is subject to change without notice
  // using the beta feature in custom rules is not suggested until it has been formally released
  //
  and beta.fuzzy_attack_score().verdict == "graymail"
```

zoomequipd (changes requested):

detection-rules/fuzzy_graymail_rule.yml
```diff
@@ -0,0 +1,12 @@
+name: "Fuzzy Graymail"
```
zoomequipd:
```suggestion
name: "Fuzzy Attack Score: Advanced Graymail Detection"
```

zoomequipd (changes requested):

zoomequipd (approved):

PR #4632, opened 2026-06-08
Update generic_pdf.yar

keaton-sublime:
# Description
yara sig for a FN on a PDF with a fake invoice. There's not much to detect here but we do have a couple options, and when paired together it is pretty solid in testing. 

## Associated samples

- [Sample 1](https://platform.sublime.security/messages/507f5fd02facc9631c08225daabad546bedb578890fae610e24751dc787b93fa?preview_id=019e938d-b493-7bce-ad1d-817fdb933413)

## Associated hunts
best approximation to the yara rule we're creating:
- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019ea83e-48a0-7600-beb7-3159e26fd511)


keaton-sublime:
Marking as review-needed/ready for review - this is just a yara sig with no matching MQL.

zoomequipd (approved):

PR #4633, opened 2026-06-08
Refine name and description for UPS impersonation rule

cybher0808:
# Description
Adding `UPS-\w+` and `UPS.com` sender.display logic to the rule. 

# Associated samples
- [Sample 1](https://platform.sublime.security/messages/5078a039950381b2592cdf85c70567a009a55e72603d3ea8b7e3507b2a073fb5)

## Associated hunts
- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019eb811-2d6e-7368-9ac5-3beb4eabc555)
- [Multi-hunt](https://hunt.limeseed.email/hunts/574acc44-aea4-4cd0-9468-514eb5b450a1)

### DIFF
--- a/detection-rules/impersonation_ups.yml
+++ b/detection-rules/impersonation_ups.yml
@@ -1,6 +1,6 @@
-name: "Brand impersonation: UPS"
+name: "Brand Impersonation: UPS Delivery"
 description: |
-  Impersonation of United Parcel Service (UPS).
+  Detects messages impersonating UPS (United Parcel Service) through display name, email address patterns, subject content, or HTML styling that mimics UPS branding, while excluding legitimate UPS domains.
 references:
   - "https://www.bleepingcomputer.com/news/security/phishing-campaign-uses-upscom-xss-vuln-to-distribute-malware/"
   - "https://twitter.com/DanielGallagher/status/1429794038463479813"
@@ -12,6 +12,7 @@ source: |
   and sender.email.domain.root_domain not in ("ups.com", "upsemail.com")
   and (
     sender.display_name in~ ("UPS My Choice", "UPS Services")
+    or regex.icontains(sender.display_name, 'UPS-\w+')
     or strings.ilike(sender.email.local_part, "*united*parcel*service*")
     or strings.ilike(sender.email.domain.domain, '*united*parcel*service*')
     or strings.icontains(subject.subject, 'UPS delivery')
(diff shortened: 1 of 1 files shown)


cybher0808:
Most recent - update after sharing results from mode - [hunt - 3Days](https://platform.sublime.security/messages/hunt?huntId=019eb25d-bc7c-746c-8e1a-e6d355011e35). Submitting a R4R.

detection-rules/impersonation_ups.yml
```diff
@@ -1,6 +1,6 @@
-name: "Brand impersonation: UPS"
+name: "Brand Impersonation: UPS Delivery"
```
IndiaAce:
```suggestion
name: "Brand impersonation: UPS Delivery"
```
Is UPS Delivery the name of the org? Genuinely not sure, if it's just the word "delivery" lets lowercase that please and thank you!

detection-rules/impersonation_ups.yml
```diff
@@ -12,6 +12,7 @@ source: |
   and sender.email.domain.root_domain not in ("ups.com", "upsemail.com")
   and (
     sender.display_name in~ ("UPS My Choice", "UPS Services")
+    or regex.icontains(sender.display_name, 'UPS-\w+')
```
IndiaAce: Non blocking since it's alllll over this rule, but icontains is case insensitive so no need to all-caps that. Also  I found a sample that we *might* want to match on this but it's `Ups.com` so your "-" here causes it not to match, you don't have to hoover this up into your change and I'm aligned with approving this tbh but wanted to show you: https://platform.sublime.security/messages/507db04ef7f2d59883b680ea4827f37284d34299084e48bef04816fda4afdaab?preview_id=019e84b4-3c38-7661-8dfe-1663a5f2ed4f nice find on this!

IndiaAce (changes requested):

IndiaAce (changes requested):
Few nits!

cybher0808:
Made the changes after the review. I found this will be useful to be to add the `Ups.com` to sender display after researching that most legitimate emails have UPS or other names as the sender domain. This email you shared was the only one I came across so I just added this simple fix to flag on this sample.

Marking for R4R.

IndiaAce (approved):
I'm aligned here, nice work on this!

PR #4668, opened 2026-06-15
Refine detection rule for Canada Revenue Agency impersonation

cybher0808:
# Description
Updating this rule with additional logic for text. 

# Associated samples
- [Sample 1](https://platform.sublime.security/messages/5085165f52578b9f3a8f08640ba70c9850b2ec77cbc058ea334d72c106fdbabb?preview_id=019eacc7-e2a6-713d-902b-d29a92cb528a)
- [Sample 2](https://platform.sublime.security/messages/507e6447cb5030613fe3bb4270a313b1eb37885bd960b8363a2cbb125c54c1c4?preview_id=019e85bf-eb96-74bd-90da-21f60a9a6b7f)

## Associated hunts
- [Netnew - Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019fb682-51b3-72eb-97fa-b9a62f1af63e)
- [Multi-Hunt 30D](https://hunt.limeseed.email/hunts/54a7fd19-1e9e-4363-bf45-86cbdd9a5d81)

### DIFF
--- a/detection-rules/fake_canada_taxrevenue_T4.yml
+++ b/detection-rules/fake_canada_taxrevenue_T4.yml
@@ -4,6 +4,19 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
+  and (
+    // cred theft
+    any(ml.nlu_classifier(body.current_thread.text).intents,
+        .name == "cred_theft" and .confidence != 'low'
+    )
+    // contains a link not to the domain "canda.ca"
+    or any(body.current_thread.links,
+           .href_url.domain.root_domain not in ('canada.ca')
+    )
+    // contains any attachment that isn't an image
+    or any(attachments, .file_type not in $file_types_images)
+  )
+
   // sender claims to be CRA
   and (
     strings.icontains(sender.display_name, 'canada revenue agency')
@@ -16,25 +29,16 @@ source: |
       )
     )
   )
-  // nlu classification for the email's text
-  and (
-    any(ml.nlu_classifier(body.current_thread.text).intents,
-        .name == "cred_theft" and .confidence != 'low'
-    )
-    or any(ml.nlu_classifier(body.current_thread.text).entities,
-           .name in ("sender", "org")
-           and .text in ('Canada Revenue Agency', 'Canada Revenue Agency (CRA)')
-    )
-    or ml.nlu_classifier(body.current_thread.text).language == "french"
-  )
+
+  // not a high trust or cra-arc.gc.ca
   and not (
     (
       // negate highly trusted sender domains
       sender.email.domain.root_domain in $high_trust_sender_root_domains
-      // negate legit senders from merck
+      // negate legit senders from legitimate cra
       or sender.email.domain.root_domain == "cra-arc.gc.ca"
     )
-    // enforce auth
+    // enforce authentication
     and coalesce(headers.auth_summary.dmarc.pass, false)
   )
 attack_types:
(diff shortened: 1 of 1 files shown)


cybher0808:
Result look good, this is a quick add on from the last rule. Marking R4R.

detection-rules/fake_canada_taxrevenue_T4.yml
```diff
@@ -16,9 +16,16 @@ source: |
...
-          .name == "cred_theft" and .confidence != 'low'
+  // nlu classification for the email's text
+  and (
+    any(ml.nlu_classifier(body.current_thread.text).intents,
+        .name == "cred_theft" and .confidence != 'low'
+    )
+    or any(ml.nlu_classifier(body.current_thread.text).entities,
+           .name in ("sender", "org")
+           and .text in ('Canada Revenue Agency', 'Canada Revenue Agency (CRA)')
+    )
```
zoomequipd: This stanza appears to be designed to detect the "suspicious" part of an impersonation, the NLU sender/org logic seems to be a better match for the stanza above this one (`// sender claims to be CRA`)

detection-rules/fake_canada_taxrevenue_T4.yml
```diff
@@ -16,9 +16,16 @@ source: |
...
+    or ml.nlu_classifier(body.current_thread.text).language == "french"
```
zoomequipd: This appears to be a pretty broad "suspicious" element and given that french of spoken in some parts of Canada I suspect isn't a very strong indicator.

zoomequipd (changes requested):

zoomequipd:
you might have better luck "flipping' the logic of this rule around and instead of trying to find/define "suspicious", define what different from the benign/legit samples. 

for example, I noticed that VERY few of the legit emails that are from Canada Revenue Agency have links in the body.current_thread and when it does, they are to known/legit domains (canada.ca)

this hunt tests that (looks for messages from the high_trust/legit sender domain WITH links and returns 2 matches
https://platform.sublime.security/messages/hunt?huntId=019ef011-c22a-786f-a47a-270a05034b00

any message that appears to be from CRA and has links that don't go to canada.ca might be worth while. 
hunt testing that theory: https://platform.sublime.security/messages/hunt?huntId=019ef017-9d59-74b5-8de1-7399fff71538

---
The other sample includes a PDF attachment, I found that the normal CRA emails contain only image attachments. 
here is a hunt that looks for any attachments that are not images. 
https://platform.sublime.security/messages/hunt?huntId=019ef023-7247-7514-aecf-420733bf01fb


--- 
here is a search of messages from CRA that can be used as references
https://platform.sublime.security/messages/search?from=cra-arc.gc.ca&limit=100&lookBack=all-time

zoomequipd:
a net new hunt (over the current version of the rule) that implements some of my ideas
https://platform.sublime.security/messages/hunt?huntId=019ef030-35dd-7ddb-83d6-688de93405aa

cybher0808:
/mql-mimic-exempt: 875881

 875881 not in scope with rule

cybher0808:
Both @zoomequipd and I went over this rule live. The results look good, marking R4R.

detection-rules/fake_canada_taxrevenue_T4.yml
```diff
@@ -1,33 +1,46 @@
-name: "Brand impersonation: Canada Revenue Agency"
+name: "Brand impersonation: Canada Revenue Agency (CRA)"
```
zoomequipd: I would skip the rename only because it's a cosmetic change and doesn't reflect a change in the rule's logic. 

Because rule names can be used in custom automations, I generally limit the renames when the intention/scope/etc of the rule changes. 
```suggestion
name: "Brand impersonation: Canada Revenue Agency"
```

zoomequipd (changes requested):

detection-rules/fake_canada_taxrevenue_T4.yml
cybher0808: Okay, noted I will keep that in mind.

zoomequipd (approved):

PR #4705, opened 2026-06-22
excludes false positives from outlign.co

hilt86:
Outlign sends messages that probably match on task assigned messages from Outlign.

# Description

# Associated samples

### DIFF
--- a/detection-rules/brand_impersonation_ms_planner.yml
+++ b/detection-rules/brand_impersonation_ms_planner.yml
@@ -51,7 +51,8 @@ source: |
                "msftauthimages.net",
                "mimecastprotect.com",
                "office.com",
-               "microsoftproject.com"
+               "microsoftproject.com",
+               "outlign.co"
              )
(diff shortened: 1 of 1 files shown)


zoomequipd:
Hello @hilt86  Thanks for this submission!

There are currently two different locations for handle FPs within this rule. 
1. by negating the .href_url.domain.root_domain of the link (which you've implemented)
2. by negating on the sender root domain 

Why did you select option 1 instead of 2? 
Do you happen to have a sample that you can share (I'm on the Sublime Community slack if needed)? 

Thanks again!

hilt86:
@zoomequipd I think I chose #1 because there were examples in the rule using that method. I will send you an example on Slack

zoomequipd:
thank you for the sample! I've gone ahead and created https://github.com/sublime-security/sublime-rules/pull/4949 based on my review of the message, it include resolution of your FP. If you're ok with it, i'd like to close this PR and move forward with #4949

zoomequipd:
closing in favor of #4949

PR #4710, opened 2026-06-22
Update generic_zip.yar

keaton-sublime:
##Description
Two changes:
- with the older zipline_delivery rule I'm updating the metadata.
- new rule zip_smuggler_default - this is looking at the default zip smuggler (project) files that it creates with data embedded into the "extra data" space of the file but no extra data length set.

### DIFF
--- a/yara/generic_zip.yar
+++ b/yara/generic_zip.yar
@@ -2,7 +2,8 @@ rule zipline_delivery_telekom {
 	meta:
 		author      = "kyle eaton"
 		date        = "2026-06-18"
-		description = "https://github.security.telekom.com/2026/06/ZipLine-linked-spearphishing-campaign.html"
+		ref = "https://github.security.telekom.com/2026/06/ZipLine-linked-spearphishing-campaign.html"
+		description = "zip file with one double extension (docx.lnk) and two docx lures per the ref campaign"
 	strings:
 		$pklfh = { 50 4b 03 04 }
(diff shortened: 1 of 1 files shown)


keaton-sublime:
Yara rule with no matching MQL - this is for hunting at this time. I'll add a MQL rule if the rule looks solid in testing.

yara/generic_zip.yar
```diff
@@ -15,4 +16,27 @@ rule zipline_delivery_telekom {
...
+
+rule zip_smuggler_default {
+	meta:
+		author      = "kyle eaton"
+		ref      = "https://github.com/Octoberfest7/zip_smuggling"
+		date        = "2026-06-22"
+		description = "zip file with the default 'egghunt' value 0x55555555 after the file bytes of a PKLFH. s/o @ffforward"
+	strings:
+		$pklfh = { 50 4b 03 04 }
+		$ext   = ".lnk"
```
zoomequipd, line 29: does it make more sense to do the ext analysis/filter in MQL?

I suppose BOTH might be required given that all yara rules run on all files submitted to file.explode

zoomequipd (approved):

yara/generic_zip.yar
keaton-sublime: yeah i think this will need to be in yara just for the other conditions we're checking - but we could add a check in the MQL rule to only include zip files with lnk files inside.

PR #4722, opened 2026-06-23
Add CI check for rule files in sub-folders (DEOPS-294)

aidenmitchell:
Adds a CI step to `rule-validate.yml` that fails the build when a rule file is committed into a sub-folder instead of directly in its top-level rules directory.

## Why

Per DEOPS-294: a rule was recently committed at `detection-rules/html_cred_link/html_cred_link.yml` instead of `detection-rules/html_cred_link.yml`. The validation pipeline globs only one level deep (`*-rules/*.yml`, `exclusions/*.yml`), so a nested file silently slips past validation, MQL auto-formatting, and rule-ID assignment — producing confusing downstream failures rather than a clear error.

## What the check does

```
found=$(find *-rules exclusions -mindepth 2 -name '*.yml' 2>/dev/null)
```

- Scans every `*-rules` directory plus `exclusions` for `.yml` files nested more than one level deep.
- On a hit, emits a GitHub `::error file=...::` annotation per offending file with a suggested corrected flat path, e.g.:
  > `'detection-rules/html_cred_link/html_cred_link.yml' is in a sub-folder and will be skipped by validation. Did you mean 'detection-rules/html_cred_link.yml'?`
- `insights/` is intentionally exempt — it nests into category sub-folders by design.
- The `*-rules` glob mirrors the canonical directory list in the "Validate Rules" step, so new rule dirs are covered automatically.

Verified the logic flags nested files (with correct suggestions), ignores `insights`, and passes cleanly against current `main`.


keaton-sublime (approved):
lgtm

PR #4747, opened 2026-06-25
Add detection rule for suspicious copy-paste URLs

peterdj45:
This rule detects messages that instruct users to manually copy and paste URLs from suspicious domains, specifically targeting free subdomain hosting services. It includes support for non-English messages through translation.

# Description

# Associated samples
- https://platform.sublime.security/messages/508c338769d420b14be9f454668f911eac9788d672be05be8064090e0b235bfb?preview_id=019ef0ab-f379-7f18-93d0-cb9a960e560e

## Associated hunts
-


peterdj45:
actually, i'll update the existing french-language rule rather than making this standalone. #4768

PR #4773, opened 2026-06-30
Create link_storage_google_la_redirect.yml

zoomequipd:
# Description

Add ASR rule for observed technique used by a variety of different malicious message themes and intentions.

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/509af52be103fd125985af9fba7db3421b19c14c622f510522b8fd0925ccbe57?preview_id=019f1817-2a34-7f23-8dd3-59a6055299fc)
- [Sample 2](https://platform.sublime.security/messages/5097c6f82fce5563a315a4ed9c6f6855951b3d52f82921d8f385a6103abe5721?preview_id=019f1956-ecfb-7e6b-88c9-8fe9a91c3aa3)
- [Sample 3](https://platform.sublime.security/messages/509260d3668498dc93746430824c6a8f7e33aad47323dd3882f9f204677a38cf?preview_id=019f1956-e5fe-74dd-b940-cd086f6a81bb)
- [Sample 4](https://platform.sublime.security/messages/508eee1e46770a9e4c168d6d6370289b42f5c95c3def351965fc0a5476e0be07?preview_id=019f159d-1d12-756f-b94d-ed90eef0a2b8)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019f1961-db71-7f3c-94d9-cb4d6192b0f1)
- [L14D Multihunt](https://hunt.limeseed.email/hunts/7dd585f8-1dc0-410f-be9c-72bc9b565013)

### DIFF
--- a/detection-rules/link_storage_google_la_redirect.yml
+++ b/detection-rules/link_storage_google_la_redirect.yml
@@ -26,3 +26,4 @@ tactics_and_techniques:
 detection_methods:
   - "URL analysis"
   - "Threat intelligence"
+id: "c687f8bc-0e07-5c22-93d7-7dcf2d7986e8"
--- a/detection-rules/link_storage_google_la_redirect.yml
+++ b/detection-rules/link_storage_google_la_redirect.yml
@@ -1,0 +1,28 @@
+name: "Link: Google Cloud Storage redirect to external domain"
+description: "Messages contain links hosted on storage.googleapis.com that, when followed, redirect to a destination outside of googleapis.com. This technique abuses Google Cloud Storage's trusted reputation to bypass link reputation checks, with the actual payload hosted on an unrelated external domain. Observed lures include parcel delivery notifications impersonating shipping carriers (UPS, GLS), gambling bonus offers, health product promotions, and storage quota warnings."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and any(body.links,
+          .href_url.domain.domain == "storage.googleapis.com"
+          and (
+            ml.link_analysis(., mode="aggressive").effective_url.domain.root_domain != "googleapis.com"
+            and ml.link_analysis(., mode="aggressive").effective_url.domain.root_domain != ""
+          )
+  )
+tags:
+  - "Attack surface reduction"
+attack_types:
+  - "Credential Phishing"
+  - "Spam"
+  - "Malware/Ransomware"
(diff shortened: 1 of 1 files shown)


detection-rules/link_storage_google_la_redirect.yml
```diff
@@ -0,0 +1,29 @@
...
+description: "Messages contain links hosted on storage.googleapis.com that, when followed, redirect to a destination outside of googleapis.com. This technique abuses Google Cloud Storage's trusted reputation to bypass link reputation checks, with the actual payload hosted on an unrelated external domain. Observed lures include parcel delivery notifications impersonating shipping carriers (UPS, GLS), gambling bonus offers, health product promotions, and storage quota warnings."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and any(body.links,
+          .href_url.domain.domain == "storage.googleapis.com"
+          and (
+            ml.link_analysis(., mode="aggressive").effective_url.domain.root_domain != "googleapis.com"
+            and ml.link_analysis(., mode="aggressive").effective_url.domain.root_domain != ""
```
IndiaAce: are these ever going to be at odds? https://hunt.limeseed.email/hunts/fc3ede0e-7f1a-47ff-81c6-37c987bb2310

IndiaAce (changes requested):

detection-rules/link_storage_google_la_redirect.yml
zoomequipd: IIRC, the `root_domain != ""` logic was added  due to pages that were not submitted did not have a effective_url, well. they did, it was just empty.  This caused the first logic `.root_domain != "googleapis.com"` to be true and match on samples where the URL was never submitted to LA. 

This removed those matches.

zoomequipd: > You were right to push on the != "" guard. I thought it was catching unresolved/null effective_url links that would otherwise falsely match — but verifying it: null != "googleapis.com" evaluates to null (falsy, safe in MQL), and unresolved links come back null, never "", in every sample I checked. So the guard only defends against a literal empty-string case we've never actually observed. It's harmless to keep, but it didn't fix the earlier runaway (that was the .href_url vs. full-link-object bug) and no sample requires it.

IndiaAce (approved):

PR #4802, opened 2026-07-06
Hoist NLU classifier call out of body.links loop in link_mailto_placeholder

thunderkatz:
# Description

The classifier check on `body.current_thread.text` doesn't depend on link content, so nesting it inside `any(body.links, ...)` means the call site is reached once per qualifying link instead of once overall, and even though it's cached, it's not free. Hoisting the loop-independent check out of the loop saves that. Note that this requires splitting into two `any`s, since it was previously `any(P or (Q and C)`, so now it's `any(P) or (any(Q) and C)` (the scheme check is duplicated in both loops but is significantly cheaper than the cached nlu call).

# Associated samples
n/a - this is a no-op optimization

## Associated hunts
n/a

# Screenshot (insights)
n/a

### DIFF
--- a/detection-rules/link_mailto_placeholder.yml
+++ b/detection-rules/link_mailto_placeholder.yml
@@ -5,14 +5,13 @@ severity: "medium"
 source: |
   type.inbound
   and (
-    // @{domain} pattern is strong on its own
+    // @{domain} pattern is strong
     any(body.links,
         .href_url.scheme == "mailto"
         and regex.icontains(.href_url.url, '@\s*{\s*domain\s*}')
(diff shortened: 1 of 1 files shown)


zoomequipd (dismissed):
Approving without typical hunts due to performance concerns with running the hunts. 

@thunderkatz has validated this change should have no efficacy impact.

zoomequipd (approved):
@rw-access

thunderkatz:
Bigger issue is regex, not the repeated cache lookup, going to close to minimize unneeded churn.

PR #4810, opened 2026-07-06
Update zoom_events_abuse.yml

markmsublime:
# Description
broadening the xpath checks in this rule to cover more samples and fix the NLU check being in the wrong loop

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/509c733f1244d0d7da0a94e58e81dad409a54d9a8aba197f8991370dd296e8b1?preview_id=019f2448-a6e5-7db0-9727-3fc9aa9585f8)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019f4225-9693-73a9-b161-04e46b3324e1) - net new results

### DIFF
--- a/detection-rules/zoom_events_abuse.yml
+++ b/detection-rules/zoom_events_abuse.yml
@@ -17,10 +17,6 @@ source: |
               .name == "cred_theft" and .confidence != "low"
           )
   )
-  // negate legitimate messages where all links lead back to zoom
-  and not all(filter(body.links, .href_url.scheme != "mailto"),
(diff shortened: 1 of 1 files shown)


markmsublime:
tried negating the two unwanted matches, losing coverage for the large increase of matching samples. not easy to work around NLU in this rule. however, these are both correctly identified by layer 2 and I believe are acceptable for this rule. hunt updated after test rule share, marking review needed

IndiaAce (approved):

PR #4813, opened 2026-07-06
Create credential_phishing_remittance_lure.yml

markmsublime:
# Description
a common theme among cred phish campaigns I see is the use of remittance, trying this NLU based logic to see how much I can capture and if it makes a good feed rule or not 

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019f3935-c3be-7c06-9702-f43d46e2fbb8)


markmsublime:
this one is not it

PR #4817, opened 2026-07-07
Limit phishing_blue_button_links to anchors that actually have a link

thunderkatz:
# Description

If the anchor's href is to something like "#", there's no point in evaluating its style. Move the faster filters first.

# Associated samples
n/a

## Associated hunts
https://platform.sublime.security/hunts/019f3cd1-39e4-7254-bd45-22d6bae942ec
Found 7 message groups:
* One had `href="#"` and was still flagged by a medium severity rule
* All others had `href=""` surrounding no text, and were still flagged by multiple high severity rules

# Screenshot (insights)
n/a


thunderkatz:
Closing - not the right fix either.

PR #4829, opened 2026-07-07
Update generic_pdf.yar

keaton-sublime:
# Description
Matching PDFs which include a box shaped lure and an outbound link, along with embedded javascript.


keaton-sublime:
Marking review needed / ready for review. This is only a yara rule with no MQL.

markmsublime (approved):

PR #4845, opened 2026-07-09
Update generic_pdf.yar

keaton-sublime:
# Description
Yara rule for the PDF matching on the image size along with either the title or link size they add in.

# Associated samples
- [Sample 1](https://platform.sublime.security/messages/509cc98bdcfd6ee49bb91bc062a5188da7be2f41235a300a75e6b53add9f053f?preview_id=019f4707-dcd0-7a1a-98f0-206eba7ebcf6)

### DIFF
--- a/yara/generic_pdf.yar
+++ b/yara/generic_pdf.yar
@@ -522,3 +522,18 @@ rule pdf_rect_size_box_lure {
 	condition:
 		$header at 0 and $rect_size
 }
+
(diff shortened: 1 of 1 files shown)


keaton-sublime:
This is a yara rule with no matching MQL, marking ready for review/reveiw-needed so we can start testing this out.

markmsublime (approved):

PR #4857, opened 2026-07-09
Update brand_impersonation_gemini.yml

keaton-sublime:
# Description
A couple things:
1. We had one of the OR statements in there twice, removing that. 
2. We are using strings.icontains to check the body links, but in testing, we can get away with href_url.url instead. This should help with some perf issues for the rule. 

## Associated samples

- [Sample 1](https://platform.sublime.security/messages/508c2ea2aec45a7ff43f906d940062acf81a4ec2820e3bad43e4aa5320e4cb56?preview_id=019ed1a2-01ae-7935-b32d-b929030a5f54)
- [Sample 2](https://platform.sublime.security/messages/4fd8c31223a48f6d3c6eee15f14311304de49d22152a2cd4d55247fa78b5f16d?preview_id=019b324d-1f1d-754a-9653-34e443a09a55)

## Associated hunts

- [Current Logic](https://platform.sublime.security/messages/hunt?huntId=019f4872-e140-7dfb-b346-1b1fad6c13bd)
- [this PR logic](https://platform.sublime.security/messages/hunt?huntId=019f4881-c79b-705b-87d5-40c132414cab)
- [alternative idea for the fix](https://platform.sublime.security/messages/hunt?huntId=019f4861-b3dd-70a4-b0db-4a9d84a4d8c9)

The difference in these is coming from instances where multiple twitter links are present in the same message.

### DIFF
--- a/detection-rules/brand_impersonation_gemini.yml
+++ b/detection-rules/brand_impersonation_gemini.yml
@@ -6,64 +6,35 @@ source: |
   type.inbound
   and (
     // the address in the footer
-    (
-      regex.icontains(body.current_thread.text,
-                      "600 Third Avenue.{1,2}2nd Floor.{1,2}New York, NY"
-      )
-      and strings.icontains(body.current_thread.text, "Gemini Trust Company")
+    regex.icontains(body.html.display_text,
+                    "600 Third Avenue.{1,2}2nd Floor.{1,2}New York, NY"
     )
-    // logo detect combined with sender display name 
+    or strings.icontains(body.html.display_text, "Gemini Trust Company")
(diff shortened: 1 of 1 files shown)


keaton-sublime:
- [Hunt with old version](https://platform.sublime.security/messages/hunt?huntId=019fcd4d-aa3c-76aa-8c69-3b5af8b89749)
- [Hunt with updated change](https://platform.sublime.security/messages/hunt?huntId=019fcd50-08ed-7e13-99fb-79a4994fa456) 

That updated change shows that we no longer match the FPs from test rules.

PR #4860, opened 2026-07-10
Add daily auto-format MQL rules workflow

mpasquale-3:
## Description

Adds a daily GitHub Actions workflow that auto-formats all detection rules and insights using Sublime's MQL format API, opening a single rolling PR (`auto/mql-format`) for detection team review when formatting drifts.

## Why

The MQL formatter may evolve over time. Rather than requiring individual PRs to carry a paired sublime-rules formatting PR, a daily chore runs the formatter and surfaces any drift as a single reviewable PR. The PR branch is fixed (`auto/mql-format`) so subsequent runs amend the same PR rather than opening new ones.

## Test plan

- [ ] Trigger workflow manually via `workflow_dispatch` to verify it runs successfully
- [ ] Confirm `peter-evans/create-pull-request` creates/updates the `auto/mql-format` branch correctly

### DIFF
--- a/.github/workflows/mql-auto-format.yml
+++ b/.github/workflows/mql-auto-format.yml
@@ -1,0 +1,39 @@
+name: Auto-format MQL rules
+
+on:
+  schedule:
+    - cron: '0 6 * * *'  # daily at 06:00 UTC
+  workflow_dispatch:
+
+jobs:
+  format:
+    name: Format MQL rules
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out sublime-rules
+      uses: actions/checkout@v4
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+
+    - name: Install dependencies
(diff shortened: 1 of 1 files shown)


aidenmitchell (commented):
If we run auto-formatting on PRs in CI actions, what is this for? Just if it misses something?

aidenmitchell (approved):
addressed here: https://sublimesecurity.slack.com/archives/C0B7NQYKNKS/p1783963613578449

mpasquale-3:
> If we run auto-formatting on PRs in CI actions, what is this for? Just if it misses something?

You're right, rule-validate.yml does do a full sweep of rules. So mql-auto-format.yml is not strictly necessary, but it still has these helpful properties:

1. We'll get bot PRs for format changes even when there aren't active PRs to rules otherwise.
2. Since we'll catch format changes from the bot PRs, normal PRs will remain scoped to only the rules you intend to change.

PR #4885, opened 2026-07-14
Update generic_pdf.yar

keaton-sublime:
## Description
yara rule for PDF with 'open secure document' lure. These look at the image size and the link positioning, allowing us to catch this sample and another variation. 
## Associated samples
- [Sample 1](https://platform.sublime.security/messages/509bc054e6cde8e3f4646beadd3a0aa4fc973efa22b13598366aa43c4f6a3fd9?preview_id=019f1e70-c89b-7f0d-88be-908bd249d093)

### DIFF
--- a/yara/generic_pdf.yar
+++ b/yara/generic_pdf.yar
@@ -537,3 +537,27 @@ rule pdf_view_doc_here_title_box {
 		$header at 0 and $image_size and ($title or $link)
 }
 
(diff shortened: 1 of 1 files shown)


keaton-sublime:
marking ready for review and review needed, there's no matching MQL so this is just to gather telemetry for now.

markmsublime (approved):

PR #4877, opened 2026-07-13
Update service_abuse_calendly_callback.yml

markmsublime:
# Description
add second calendly sender address

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/50a7605bf9f980083cfad093a5d7d202463afb9a8da41085544f9aede686e63c?preview_id=019f5c5c-51a2-76a0-ba1e-3ad2aa5848ae)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019f5caa-99e4-78d4-8018-eb92611ad968) - new matches
- [multi](https://hunt.limeseed.email/hunts/72360dc9-5899-4167-8e64-dcd72a08d169)

PR #4886, opened 2026-07-14
Update service_abuse_calendly_callback.yml

markmsublime:
# Description
follow up PR based on feedback from #4877, seeing if we get better overall coverage using root domain instead of specific addresses

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019f6154-b95d-78cd-b0c5-df3f58f2ef55)


markmsublime:
too broad, keeping rule as is

PR #4889, opened 2026-07-14
Create link_javascript_obfuscation_redirect.yml

D-Bolton:
# Description
Detects inbound messages sent to self, invalid recipients, or undisclosed recipients containing links that, upon analysis, include a JavaScript setTimeout redirect using location.hash to obscure the final destination.

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/509486be52c48d7f1a6c71aea776748f6b8e4eeae5ee912177b297c63711863b?preview_id=019f42f4-4133-76bb-9383-6c00ba725a3c)

## Associated hunts

- [97 customer multi-hunt](https://hunt.limeseed.email/hunts/93dc92c5-b625-4f87-92ad-b93b3b5a4e9f)
- [SWS Hunt](https://platform.sublime.security/messages/hunt?huntId=019f6262-22ad-7e14-9bf8-73378bfc7c8b)

### DIFF
--- a/detection-rules/link_javascript_obfuscation_redirect.yml
+++ b/detection-rules/link_javascript_obfuscation_redirect.yml
@@ -1,0 +1,49 @@
+name: "Link: JavaScript redirect with hash-based obfuscation"
+description: "Detects inbound messages sent to self, invalid recipients, or undisclosed recipients containing links that, upon analysis, include a JavaScript setTimeout redirect using location.hash to obscure the final destination. This technique delays redirection and appends URL fragments to evade static link inspection."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and (
+    // self sender or invaild recipent domain or local parts match
+    (
+      length(recipients.to) == 1
+      and (
+        sender.email.email == recipients.to[0].email.email
+        or recipients.to[0].email.domain.valid == false
+        or sender.email.local_part == recipients.to[0].email.local_part
+      )
+    )
+    // no recipients defined
+    or (
+      (
+        length(recipients.to) == 0
+        or all(recipients.to, .display_name == "Undisclosed recipients")
+      )
+      and length(recipients.cc) == 0
+      and length(recipients.bcc) == 0
+    )
+  )
+  and 0 < length(body.links) < 10
+  and any(body.current_thread.links,
+          any(html.xpath(ml.link_analysis(., mode="aggressive").final_dom,
+                         "//script"
+              ).nodes,
+              strings.starts_with(.raw,
+                                  '<script>setTimeout(function(){window.location="'
+              )
+              and strings.ends_with(.raw, '"+location.hash},3000)</script>')
+          )
+  )
+  
+attack_types:
+  - "Credential Phishing"
+tactics_and_techniques:
+  - "Scripting"
+  - "Evasion"
+  - "Social engineering"
+detection_methods:
+  - "URL analysis"
+  - "Javascript analysis"
+  - "HTML analysis"
+  - "Sender analysis"
--- a/detection-rules/link_javascript_obfuscation_redirect.yml
+++ b/detection-rules/link_javascript_obfuscation_redirect.yml
@@ -47,3 +47,4 @@ detection_methods:
   - "Javascript analysis"
   - "HTML analysis"
   - "Sender analysis"
+id: "05c8c2d1-6018-5a3e-9e20-be1b41796e62"
--- a/detection-rules/link_javascript_obfuscation_redirect.yml
+++ b/detection-rules/link_javascript_obfuscation_redirect.yml
@@ -24,7 +24,7 @@ source: |
       and length(recipients.bcc) == 0
     )
   )
-  and 0 < length(body.links) < 10
+  and 0 < length(body.current_thread.links) < 10
   and any(body.current_thread.links,
           any(html.xpath(ml.link_analysis(., mode="aggressive").final_dom,
                          "//script"
--- a/detection-rules/link_javascript_obfuscation_redirect.yml
+++ b/detection-rules/link_javascript_obfuscation_redirect.yml
@@ -32,7 +32,7 @@ source: |
               strings.starts_with(.raw,
                                   '<script>setTimeout(function(){window.location="'
               )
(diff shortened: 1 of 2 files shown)


D-Bolton:
Hunt results look good. Tagging for review.

detection-rules/link_javascript_obfuscation_redirect.yml
```diff
@@ -0,0 +1,50 @@
...
+  )
+  and 0 < length(body.links) < 10
+  and any(body.current_thread.links,
+          any(html.xpath(ml.link_analysis(., mode="aggressive").final_dom,
+                         "//script"
+              ).nodes,
+              strings.starts_with(.raw,
+                                  '<script>setTimeout(function(){window.location="'
+              )
+              and strings.ends_with(.raw, '"+location.hash},3000)</script>')
```
zoomequipd: `,3000` is keying in on a very specific delay, is it worth expanding that to include other values?

detection-rules/link_javascript_obfuscation_redirect.yml
```diff
@@ -0,0 +1,50 @@
...
+    or (
+      (
+        length(recipients.to) == 0
+        or all(recipients.to, .display_name == "Undisclosed recipients")
+      )
+      and length(recipients.cc) == 0
+      and length(recipients.bcc) == 0
+    )
+  )
+  and 0 < length(body.links) < 10
```
zoomequipd: this uses body.links, consider using body.current_thread.links. 

Does this length of body links actually impact rule efficacy (resolve FPs)?

detection-rules/link_javascript_obfuscation_redirect.yml
```diff
@@ -0,0 +1,50 @@
...
+        sender.email.email == recipients.to[0].email.email
+        or recipients.to[0].email.domain.valid == false
+        or sender.email.local_part == recipients.to[0].email.local_part
+      )
+    )
+    // no recipients defined
+    or (
+      (
+        length(recipients.to) == 0
+        or all(recipients.to, .display_name == "Undisclosed recipients")
```
zoomequipd: There are a lot of variations in how this "undisclosed recipients" appears.  I've seen `Undisclosed recipients`, `Undisclosed recipients;`, `Undisclosed-recipients`, etc. 

My general approach is to use the .email.domain.valid field 

```suggestion
        or all(recipients.to, .email.domain.valid == false)
```

detection-rules/link_javascript_obfuscation_redirect.yml
```diff
@@ -0,0 +1,50 @@
+name: "Link: JavaScript redirect with hash-based obfuscation"
```
zoomequipd: name should reflect the major filtering occurring by recipient logic included in the rule

zoomequipd (changes requested):
minor feedback, name change the only blocking feedback.

detection-rules/link_javascript_obfuscation_redirect.yml
zoomequipd: Do we actually need this logic at all? given `all` matches when there are no recipients. i'd imagine this scenerio is actually already covered by your 
`length(recipients.to) == 0`  and `or recipients.to[0].email.domain.valid == false` logic. 

hunt looking for samples which match the `or all(recipients.to, .display_name == "Undisclosed recipients")` condition here: 
https://platform.sublime.security/messages/hunt?huntId=019f9600-2af2-7e55-b0ff-75c9d3908b07

detection-rules/link_javascript_obfuscation_redirect.yml
zoomequipd: TIL `location.hash` refers to the ~~anchor~~ fragment of the current location and in this case is being concatenated  with the next stage

https://developer.mozilla.org/en-US/docs/Web/API/Location/hash 

not really "obfuscation" it's just part of this landing page design.

detection-rules/link_javascript_obfuscation_redirect.yml
zoomequipd: case insensitive match prevents common FNs in general (not sure if this was present to avoid FPs though)
```suggestion
              and strings.iends_with(.raw, '"+location.hash},3000)</script>')
```

detection-rules/link_javascript_obfuscation_redirect.yml
```diff
@@ -0,0 +1,50 @@
...
+      and length(recipients.cc) == 0
+      and length(recipients.bcc) == 0
+    )
+  )
+  and 0 < length(body.links) < 10
+  and any(body.current_thread.links,
+          any(html.xpath(ml.link_analysis(., mode="aggressive").final_dom,
+                         "//script"
+              ).nodes,
+              strings.starts_with(.raw,
```
zoomequipd: consider istarts_with
```suggestion
              strings.istarts_with(.raw,
```

detection-rules/link_javascript_obfuscation_redirect.yml
D-Bolton: Updated the logic to use body.current_thread.links. 
This logic doesn't impact rule efficacy. It's a gate for LA.

detection-rules/link_javascript_obfuscation_redirect.yml
D-Bolton: Updated name and description.

D-Bolton:
[97 Customer multi-hunt](https://hunt.limeseed.email/hunts/3344baa8-9532-4ba9-bc8d-6ee5c794d76a)
[SWS Hunt](https://platform.sublime.security/messages/hunt?huntId=019fc900-c59f-7243-956c-1c6399ea69fa)

detection-rules/link_javascript_obfuscation_redirect.yml
D-Bolton: I didn't [find](https://platform.sublime.security/hunts/019fc8e8-0edb-7ea4-94a3-2913adac1699) any samples that had a different value.

D-Bolton:
Hunt results look good. Tagging for another review.

zoomequipd (approved):

PR #4892, opened 2026-07-15

cybher0808:
# Description
Adjust for length `ml.nlu_classifier` to this rule. 

# Associated samples
- [Sample 1](https://platform.sublime.security/messages/50a0c2367a898c4236f09aefa0648a665123a47653fe30827cddbf457c853b93)

## Associated hunts
- [Hunt - 90 Days](https://platform.sublime.security/messages/hunt?huntId=019f6caf-9ea1-7032-8950-abf6dcaee646)

### DIFF
--- a/detection-rules/impersonation_quickbooks.yml
+++ b/detection-rules/impersonation_quickbooks.yml
@@ -20,7 +20,7 @@ source: |
         length(filter(ml.nlu_classifier(body.current_thread.text).entities,
                       strings.icontains(.text, "quickbooks")
                )
-        ) > 2
+        ) >= 2
         and any(ml.nlu_classifier(body.current_thread.text).intents,
                 .name == "cred_theft"
         )
--- a/detection-rules/impersonation_quickbooks.yml
+++ b/detection-rules/impersonation_quickbooks.yml
@@ -10,7 +10,8 @@ source: |
                     'quickboo*',
                     'QuickB*',
                     'QBook*',
-                    'intuit*'
+                    'intuit*',
+                    'donotreply'
       )
       or strings.like(sender.display_name, "QB-*", "QB *")
       or strings.ilevenshtein(sender.display_name, 'quickbooks') <= 1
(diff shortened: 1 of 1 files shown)


missingn0pe:
This PR covers 50a0c2367a898c4236f09aefa0648a665123a47653fe30827cddbf457c853b93

cybher0808:
@missingn0pe Reviewed the changes and making the changes to the new merge commit. https://platform.sublime.security/messages/hunt?huntId=019f6bf3-468f-7d51-a2bf-faca14241808

cybher0808:
Results look good after running multi-hunt, the most recent hunt finds additional samples. Marking R4R.

detection-rules/impersonation_quickbooks.yml
```diff
@@ -10,7 +10,8 @@ source: |
                     'quickboo*',
                     'QuickB*',
                     'QBook*',
-                    'intuit*'
+                    'intuit*',
+                    'donotreply'
```
markmsublime:
```suggestion
```
do you need this here? I ran a [hunt](https://platform.sublime.security/messages/hunt?huntId=019f6c94-ffec-795f-b219-8455c64d1421) without it, just changing the other stanza seems to capture these as well. I don't think 'donotreply' is a strong signal for Quickbooks impersonation...there's a lot of meat on these samples: QB logo in the body, QB copyright in the footer, QB address in the footer, QB account info template, etc etc. I would focus on the stronger signals here to capture the impersonation holistically, or just run with the other change to this rule that also captures more samples

markmsublime (changes requested):

detection-rules/impersonation_quickbooks.yml
cybher0808: The link I could not open. I will be stick with the other change suggested instead of this above. It will still target on the samples that we are trying to capture. I will remove this stanza. @markmsublime

markmsublime (approved):

PR #4902, opened 2026-07-15
Add detection rule for suspicious toll-free numbers

peterdj45:
Detects messages containing the phone number 1 800-555-0199 in the subject or body, which is reserved for use by the entertainment industry. Commonly used as a placeholder phone number in LLM-generated campaigns.

# Associated samples
- https://platform.sublime.security/messages/4ffaa94cc935fe97a0200f15c4bc845944182e0420e2f0a9d3bf08a2445d63d3
- https://platform.sublime.security/messages/4fc2dd72eb2cac000772709f934ef984c7e2df335d96b8f778cf4c44af4449ff
- https://platform.sublime.security/messages/4fc2a075211b115b6b09fd736f9b1f06fd6a41e93bb886e9bdcbf15c711541ab

## Associated hunts
- https://platform.sublime.security/hunts/019bfbc3-af4a-76c1-8a07-e1239ae33492

### DIFF
--- a/detection-rules/sus_tollfree_number.yml
+++ b/detection-rules/sus_tollfree_number.yml
@@ -7,6 +7,8 @@ source: |
   and any([body.current_thread.text, subject.subject],
           regex.icontains(., '\b\+?(\d{1}.)?\(?8\d{2}?\)?.555.?0[12]99\b')
   )
+tags:
+ - "Attack surface reduction"
 attack_types:
   - "Spam"
 tactics_and_techniques:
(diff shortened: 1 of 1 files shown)


peterdj45:
multi-hunts look good! https://hunt.limeseed.email/hunts/8445d92e-794f-493a-81e1-de5041c0f50c

IndiaAce (changes requested):
Could we ASR this as well as keeping it a spam rule? There are some likely_benign matches that I'm sort of on the fence about calling "spam" but perhaps making it an ASR rule would help out here? I know we're shying away from using sender profile but might also be worth checking out solicited sender gates? Let me know your thoughts, but I'd suggest taking a peak at some of those likely benign matches first and seeing what you think!

peterdj45:
> Could we ASR this as well as keeping it a spam rule? There are some likely_benign matches that I'm sort of on the fence about calling "spam" but perhaps making it an ASR rule would help out here? I know we're shying away from using sender profile but might also be worth checking out solicited sender gates? Let me know your thoughts, but I'd suggest taking a peak at some of those likely benign matches first and seeing what you think!

yep, i'm good with making this an ASR rule! ran some hunts with sender profile added, but we're dropping some solicited and common prevalence TPs

IndiaAce (approved):

PR #4907, opened 2026-07-16
Update generic_zip.yar

keaton-sublime:
## Description
Matching zip files which contain a dll file and an exe that matches a known excel binary in both file size and crc32. 

## Associated samples
- [Sample 1](https://platform.sublime.security/messages/50a09a13cb77404ade8c2326eb8315f49e166d3bb419ce4a7cf19694443ddbbd?preview_id=019f61db-7dae-7eab-9626-7678ac244053)


keaton-sublime:
This is just a yara rule (no matching MQL) so I'm marking it for review at this time.

markmsublime (approved):

PR #4913, opened 2026-07-16
Add phishing simulation exclusions for common training vendors

MSAdministrator:
## Summary

Adds exclusion rules to remove known phishing-simulation / security-awareness-training messages from live analysis. Covers the following vendors:

- Adaptive Security
- Arctic Wolf MSA
- Breach Secure Now
- CanIPhish
- InfoSec IQ
- Material
- Mimecast
- Phish Notify
- Phished
- Phriendly Phishing
- Right Hand AI
- SoSafe

## Notes

Each rule matches on the vendor's known simulation sending domains and requires SPF to pass, so the simulation domains cannot be spoofed by a real attacker to bypass analysis.


MSAdministrator:
@zoomequipd Do you think this should be closed?

zoomequipd:
probably should.

PR #4914, opened 2026-07-16
Add phishing simulation header insights for common training vendors

MSAdministrator:
## Summary

Adds informational insight (`query`) rules under `insights/headers/` that surface phishing-simulation / security-awareness-training messages, tagged **Headers** and **Phishing simulation**. Covers the following vendors:

- Adaptive Security
- Arctic Wolf MSA
- Breach Secure Now
- BullPhish
- CanIPhish
- Fable
- Huntress
- InfoSec IQ
- Material
- Microsoft
- Mimecast
- NINJIO
- Phish Notify
- Phished
- Proofpoint
- Right Hand AI

## Notes

These are `informational` severity query rules that identify simulation senders by their known sending domains. Several complement the exclusion rules in #4913.


MSAdministrator:
### ⚠️ Note: `root_domain` list may need updating

CI's `root_domain` misuse check flagged 4 literals in `phishing_simulation_proofpoint.yml` that can never match a `sender.email.domain.root_domain` comparison. I removed them to get CI green, but this points at a coverage gap worth reviewing:

- `avoidphishing.securityeducation.com`
- `avoidphishing.eu.securityeducation.com`
- `avoidphishing.ap.securityeducation.com`

  These are **subdomains** of `securityeducation.com`. `root_domain` resolves them to `securityeducation.com`, which is already listed in each block — so removing them loses no coverage. **However**, if the intent was to match *only* the `avoidphishing.*` subdomains (and not all of `securityeducation.com`), the correct fix is to compare against `sender.email.domain.domain` instead of `root_domain`. Please confirm which behavior is desired.

- `corpinternal.org2`

  `.org2` is not a valid TLD, so `root_domain` resolves to `org2.` and this never matches. If a real domain was intended (e.g. `corpinternal.org`), it should be corrected rather than dropped.

More broadly: several of these simulation domain lists are long and may contain other subdomain/invalid-TLD entries that don't trip the linter but still won't match as intended. The full `root_domain` lists across the phishing-simulation insights may be worth an audit/refresh.

MSAdministrator:
@zoomequipd Do you think this should be closed?

zoomequipd:
probably should.

PR #4923, opened 2026-07-17
Update: body_fake_investment_outreach_suspicious_tld.yml with additional signals

missingn0pe:
## Description
Updates detection signals for fake investment/funding outreach spam from suspicious TLDs to catch more variants.

**Impact:** 396 net new detections

### Subject Detection
- Added support for "investment into" and "financing for" subject prefixes
- Added "family office" keyword detection
- Added mailbox first name personalization check

### Body Content Detection
- Enhanced with NLU topic classification:
  - "Financial Communications" (medium/high confidence)
  - "B2B Cold Outreach" (medium/high confidence) + financial keywords
- Simplified financial keywords: funding, capital, rates, amounts (e.g., "100k", "300k")

### Sender Signals
- Maintained suspicious TLD check
- Added explicit `.info` TLD (common in this campaign)

## Associated samples

[- Sample 1](https://platform.sublime.security/messages/50a174c424fe997de6f17e36ea3137206aef8728cea71497e50e04beb2e2b149?preview_id=019f62f6-3fb6-7358-8b8d-c4b99382d611)
[- Sample 2](https://platform.sublime.security/messages/50a4bb89ab6fbc8b558268cff3e878575f8a0deb387033ea9c9af8f97b65c776?preview_id=019f62f6-b23a-7471-8ceb-0b8eb826399d)
[- Sample 3](https://platform.sublime.security/messages/50a4ed31497776937d441e66db6362a7599568252510b4d467394d1a89450491?preview_id=019f62f7-3e8d-7dce-94dc-ed34638b6dfa)

## Associated hunts

[- Hunt 1 (Shared samples) - L90D - PR logic vs old logic](https://platform.sublime.security/messages/hunt?huntId=019f71e1-efc9-703b-94fe-dcc56310602e)
[- Hunt 2 (Shared samples) - L90D - PR logic](https://platform.sublime.security/messages/hunt?huntId=019f71e2-77a1-74d9-b308-00aaa9ca0dd6)
[- Hunt 3 (Multi-hunt) - L30D - PR logic](https://hunt.limeseed.email/hunts/9e98ca2c-02bb-4ce1-a412-8b90dc9bfd09)

PR #4943, opened 2026-07-21
Create bec_family_office_investment_solicitation.yml

JFarina5:
# Description

Observed this type of activity, frequently, in the FN queue.

Detects inbound messages from domains less than 180 days old that solicit investment interest from named organizations or individuals on behalf of a purported private family office. Senders use financial-sounding names and rotate first names and target company names while keeping the subject line structure nearly identical. The outreach is designed to appear as a legitimate investment inquiry to bait a response.

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/50a82575397a5aaf4c8915555221379cc021dde4892c96c837a018c32268496e?preview_id=019f7100-9f9e-742c-bb71-6249d95f63e7)

## Associated hunts

- [L30D Hunt](https://platform.sublime.security/messages/hunt?huntId=019fb327-c8cd-729d-84d8-a80616ca6aa8)


JFarina5:
Results are looking good. This rule may work better as a 'spam' rule given the results, but tagging malicious samples as well. Hunt in the description has been updated. Marking r4r.

JFarina5:
FN count here is sort of insane, but the rule is pretty specific and may work better as a spam rule. Hunt in the description is up to date, marking review ready.

detection-rules/bec_family_office_investment_solicitation.yml
```diff
@@ -0,0 +1,18 @@
+name: "BEC/Fraud: Fake family office investment solicitation from newly registered domains"
```
zoomequipd: while i understand this rule is matching heavy on this theme, the rule itself doesn't enforce it. Might be worth renaming the rule. 
```suggestion
name: "Spam: Large financial amount mention from newly registered sender domain"
```

detection-rules/bec_family_office_investment_solicitation.yml
```diff
@@ -0,0 +1,18 @@
...
+description: "Detects inbound messages from domains less than 180 days old that solicit investment interest from named organizations or individuals on behalf of a purported private family office. Senders use financial-sounding names and rotate first names and target company names while keeping the subject line structure nearly identical. The outreach is designed to appear as a legitimate investment inquiry to bait a response."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  and network.whois(sender.email.domain).days_old < 180
+  and regex.icontains(body.html.raw, '\$\d[-–—]\$\d{1,2}(?:m\b|million)')
```
zoomequipd: you could use the unicode class `Pd` which covers [all dashes](https://www.compart.com/en/unicode/category/Pd) 

I'm guessing the intention of body.html.raw was to capture both current_thread and previous_threads, switched over to body.html.display_text.  

```suggestion
  and regex.icontains(body.html.display_text, '\$\d\p{Pd}\$\d{1,2}(?:m\b|million)')
```

zoomequipd:
pointing out this this PR https://github.com/sublime-security/sublime-rules/pull/4923/changes#diff-81f37783333c2145bee0ca9e1860a0e2f21d27a1e51111e669cf4c81da30344fR14 is also attempting to loop these in.  

i don't see a problem with that, but just wanted to point it out.

zoomequipd (changes requested):

detection-rules/bec_family_office_investment_solicitation.yml
JFarina5: Very cool, nice idea!

detection-rules/bec_family_office_investment_solicitation.yml
JFarina5: Yep, I like it.

JFarina5:
Mode results look good for the latest change, looks like results are all TPs. Hunt in the description is up to date, marking review ready.

zoomequipd:
multihunt: https://hunt.limeseed.email/hunts/e7296001-8e23-493e-a8cf-cbb8cb21b861

zoomequipd (approved):

PR #4947, opened 2026-07-22
logic fix in attachment_credential_phishing_secure_message

sfc-gh-tpeska:
# Description
when body.links is an empty, mql returns true so `not all()` evaluates to false.  this means that the rule never fires for messages without body links.

# Associated samples

## Associated hunts

# Screenshot (insights)


sfc-gh-tpeska:
Testing, meant to push to forked repo, my bad. 😬

PR #4949, opened 2026-07-23
Update brand_impersonation_ms_planner.yml

zoomequipd:
# Description

After revewing #4705 found multiple opportunities for improvement. 

1. remove second loop of recipients.to
2. convert to beta.parse_exif instead of calling file.explode
3. add auth check for negated sender domains
4. add sender domain negation for outlign.co

# Associated samples
FP sample was provided privately


hilt86:
thanks

zoomequipd:
last commit resolved conflict, no changes to rule behavior.

IndiaAce (approved):

PR #4980, opened 2026-07-29
Update impersonation_microsoft_credential_theft.yml - Sender Negations

missingn0pe:
# Description

- Adding sender negation for legitimate B2B invites

# Associated samples

[- Sample 1](https://platform.sublime.security/messages/50a2db381eddefa8f5ca99a62ac6f9d740efb451d3f2ac458a3add6f25c4771b?preview_id=019f65e3-ab29-78ad-9f67-b9ff395a8854)

## Associated hunts

[- Hunt 1 (Shared samps) - Exclusive - L180D](https://platform.sublime.security/messages/hunt?huntId=019fafdd-5ee8-7659-a023-8b09f47849c0)
[- Hunt 2 (Multi) - Exclusive - L30D](https://hunt.limeseed.email/hunts/46dda704-08ef-4aa1-8201-8def05b20bb0)
[- Hunt 3 (Shared samps) - PR Logic - L30D](https://platform.sublime.security/messages/hunt?huntId=019fafdd-2672-7317-854e-902c991e36ef)
[- Hunt 4 (Multi) - PR Logic - L30D](https://hunt.limeseed.email/hunts/29522adf-94f6-48eb-9e75-2deb85876d51)

PR #4979, opened 2026-07-29
Update spam_onmicrosoft.yml - Sender negation

missingn0pe:
# Description

- Adding sender negation for legit MSFT b2b invites.
- Adding auth to org_domains
- General readability cleanup

# Associated samples

[- Sample 1](https://platform.sublime.security/messages/50a2db381eddefa8f5ca99a62ac6f9d740efb451d3f2ac458a3add6f25c4771b?preview_id=019f65e3-ab29-78ad-9f67-b9ff395a8854)

## Associated hunts

[- Hunt 1 (Shared samps) - Exclusive - L180D](https://platform.sublime.security/messages/hunt?huntId=019fafb4-b183-7be6-8d2b-e0787f255135)
[- Hunt 2 (Multi) - Exclusive - L30D](https://hunt.limeseed.email/hunts/b6ee503e-796f-4274-8181-e5147b010554)
[- Hunt 3 (Shared samps) - PR Logic - L30D](https://platform.sublime.security/messages/hunt?huntId=019fafb5-d485-7efa-8124-d5b715658441)
[- Hunt 4 (Multi) - PR Logic - L30D](https://hunt.limeseed.email/hunts/437c9f93-2e96-4e3e-94f4-be1e7fe20fa6)

PR #4963, opened 2026-07-27
Update: impersonation_microsoft.yml - FP - Sender negations

missingn0pe:
# Description
Adds exclusion for legitimate Microsoft B2B application invitations that were previously flagged as brand impersonation. The new logic validates messages using sender local part, domain pattern, display name, and infrastructure-validated message ID to eliminate false positives.

**Net-new coverage:** N/A (FP reduction - 135 messages excluded over 180 days)

# Associated samples

- [Sample 1](https://platform.sublimesecurity.com/messages/50a2db381eddefa8f5ca99a62ac6f9d740efb451d3f2ac458a3add6f25c4771b)

## Associated hunts

- [Hunt 1 (Shared samples) - L180D - Exclusive](https://platform.sublime.security/messages/hunt?huntId=019fa558-6f07-7337-b5c8-38905a4d279b)
- [Hunt 2 (Shared samples) - L180D - Exclusive without rule logic](https://platform.sublime.security/messages/hunt?huntId=019fa55c-0fca-79e5-bda0-2ba8317b5253)
- [Hunt 3 (Multi-hunt) - L30D - Exclusive](https://hunt.limeseed.email/hunts/61dc323a-a239-4015-bd00-2ac712f384c3)

PR #4990, opened 2026-07-30
Create: spam_onmicrosoft_applications.yml

missingn0pe:
## Description

Detects external Microsoft (onmicrosoft.com) invitations from non-Microsoft tenants that are unsolicited.

  **Impact:**
Original scope hits - 732
Reduced scope hits - 434
*FP reduction over L30D* - 298

  ## Concept

  ### Detection Logic
  - Matches standard Microsoft invitation structure: `invites@*.onmicrosoft.com` sender
  - Validates infrastructure markers: `pepf` in Message-ID (legitimate Microsoft invitation routing)
  - Requires at least one body link for invitation acceptance

  ### False Positive Reduction
  - Excludes legitimate Microsoft tenant: `microsoft.onmicrosoft.com` with SPF pass and microsoft.com links
  - Negates known recipient domains: Checks if tenant subdomain has prior communication history
  - Reply-to validation: Excludes if reply-to domain is both a known sender AND recipient (solicited)
  - High-trust domains: Excludes reply-to addresses from established high-trust domains

  ### Subject Exclusions
  - Ignores "You've left the [Organization]" termination notifications

  ## Related changed rules
[- Microsoft Impersonation](https://github.com/sublime-security/sublime-rules/pull/4963)
[- Microsoft Impersonation with embedded logo](https://github.com/sublime-security/sublime-rules/pull/4980)
[- Spam: Default Microsoft Exchange Online sender domain (onmicrosoft.com)](https://github.com/sublime-security/sublime-rules/pull/4979)

  ## Associated samples

[  - Sample 1](https://platform.sublime.security/messages/50a2db381eddefa8f5ca99a62ac6f9d740efb451d3f2ac458a3add6f25c4771b?preview_id=019f65e3-ab29-78ad-9f67-b9ff395a8854)

  ## Associated hunts

[  - Hunt 1 (Shared samples) - L180D - Net old](https://platform.sublime.security/messages/hunt?huntId=019fb4da-2a38-7c34-b1a3-45ef335d68b5)
[  - Hunt 2 (Shared samples) - L180D - PR logic](https://platform.sublime.security/messages/hunt?huntId=019fb4c4-1955-7f34-83e0-b84def36824d)
[  - Hunt 3 (Multi-hunt) - L30D - Net old](https://hunt.limeseed.email/hunts/8fed02bf-019d-4112-9e13-f99f10edc546)
[  - Hunt 4 (Multi-hunt) - L30D - PR logic](https://hunt.limeseed.email/hunts/0ae7bf93-75a0-40b0-8abc-0045020075a3)

### DIFF
--- a/detection-rules/spam_onmicrosoft_applications.yml
+++ b/detection-rules/spam_onmicrosoft_applications.yml
@@ -32,7 +32,7 @@ source: |
       length(headers.reply_to) > 0
       and (
         // domain has communicated with recipient
-        any(headers.reply_to, .email.domain.root_domain in $sender_domains)
+        any(headers.reply_to, .email.domain.domain in $sender_domains)
         // solicited sender
         and any(headers.reply_to, .email.domain.root_domain in $recipient_domains)
       )
--- a/detection-rules/spam_onmicrosoft_applications.yml
+++ b/detection-rules/spam_onmicrosoft_applications.yml
@@ -60,3 +60,4 @@ detection_methods:
   - "Header analysis"
   - "Sender analysis"
   - "URL analysis"
+id: "e991c7d7-5bbb-5521-a0bb-c34e00705d2e"
--- a/detection-rules/spam_onmicrosoft_applications.yml
+++ b/detection-rules/spam_onmicrosoft_applications.yml
@@ -34,7 +34,7 @@ source: |
         // domain has communicated with recipient
         any(headers.reply_to, .email.domain.domain in $sender_domains)
         // solicited sender
-        and any(headers.reply_to, .email.domain.root_domain in $recipient_domains)
+        and any(headers.reply_to, .email.domain.domain in $recipient_domains)
       )
     )
     // negate high trust sender root domains
--- a/detection-rules/spam_onmicrosoft_applications.yml
+++ b/detection-rules/spam_onmicrosoft_applications.yml
@@ -43,12 +43,6 @@ source: |
     )
   )
   
-  // negate microsoft's domain
-  and not (
-    sender.email.domain.domain == "microsoft.onmicrosoft.com"
-    and headers.auth_summary.spf.pass
-    and all(body.links, .href_url.domain.root_domain == "microsoft.com")
-  )
 tags:
   - "Attack surface reduction"
 attack_types:
--- a/detection-rules/spam_onmicrosoft_applications.yml
+++ b/detection-rules/spam_onmicrosoft_applications.yml
@@ -4,8 +4,8 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  // this rule matches any default microsoft onmicrosoft domains that are not msft themselves
-  // rule can be modified
+  // this rule matches any default microsoft onmicrosoft domains that are not microsoft themselves
+  // rule can be retrofitted locally or exclusions can be set on trusted senders
   
   // action link present
   and length(body.links) > 0
--- a/detection-rules/detection-rules/spam_onmicrosoft_applications.yml
+++ b/detection-rules/detection-rules/spam_onmicrosoft_applications.yml
@@ -1,0 +1,62 @@
+name: "Spam: Microsoft Applications sender default domain (onmicrosoft.com) invite from unknown sender"
+description: "Detects inbound messages sent via Microsoft's default OnMicrosoft domains using standard invite infrastructure (local part 'invites', onmicrosoft.com root domain, and 'pepf' in the message ID) that do not originate from known or trusted senders. False positives are reduced by excluding messages from known recipient domains, trusted reply-to domains, and legitimate Microsoft infrastructure where all links resolve to microsoft.com."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  // this rule matches any default microsoft onmicrosoft domains that are not msft themselves
+  // rule can be modified
+  
+  // action link present
+  and length(body.links) > 0
+  and (
+    (
+      // standard structure local part
+      sender.email.local_part == "invites"
+      // standard structure default domain
+      and sender.email.domain.root_domain == "onmicrosoft.com"
+      // infra validated message id
+      and strings.icontains(headers.message_id, "pepf")
+    )
+    // ignore terminated access notifications
+    and not strings.istarts_with(subject.base, 'you’ve left the')
+  )
+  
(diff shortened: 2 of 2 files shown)


detection-rules/spam_onmicrosoft_applications.yml
```diff
@@ -0,0 +1,63 @@
...
+  // action link present
+  and length(body.links) > 0
+  and (
+    (
+      // standard structure local part
+      sender.email.local_part == "invites"
+      // standard structure default domain
+      and sender.email.domain.root_domain == "onmicrosoft.com"
+      // infra validated message id
+      and strings.icontains(headers.message_id, "pepf")
```
zoomequipd, line 19: does it really matter if these include the header or not, so long as they "appear" to be from invites, it's probably sufficient?

detection-rules/spam_onmicrosoft_applications.yml
```diff
@@ -0,0 +1,63 @@
...
+           .email.domain.root_domain in $high_trust_sender_root_domains
+    )
+  )
+  
+  // negate microsoft's domain
+  and not (
+    sender.email.domain.domain == "microsoft.onmicrosoft.com"
+    and headers.auth_summary.spf.pass
+    and all(body.links, .href_url.domain.root_domain == "microsoft.com")
+  )
```
zoomequipd: This logic doesn't appear to be necessary based on the hunt results looking for this condition.

https://platform.sublime.security/messages/hunt?huntId=019fc8a8-0bda-70ee-afdc-db28b9e8dfda
```suggestion
```

zoomequipd (changes requested):

detection-rules/spam_onmicrosoft_applications.yml
```diff
@@ -0,0 +1,63 @@
...
+      and strings.icontains(headers.message_id, "pepf")
+    )
+    // ignore terminated access notifications
+    and not strings.istarts_with(subject.base, 'you’ve left the')
+  )
+  
+  // false positve reduction/negations
+  and not (
+    // extract the subdomain & tld, then check it against known recipients
+    strings.concat(sender.email.domain.subdomain, ".", sender.email.domain.tld) in $recipient_domains
```
zoomequipd, line 28: the tld will always be .com right? as part of `and sender.email.domain.root_domain == "onmicrosoft.com"`

so you can just static this to .com?

detection-rules/spam_onmicrosoft_applications.yml
```diff
@@ -0,0 +1,63 @@
...
+    or 
+    // or check if reply-to exists and is a solicited recipient
+    (
+      length(headers.reply_to) > 0
+      and (
+        // domain has communicated with recipient
+        any(headers.reply_to, .email.domain.root_domain in $sender_domains)
```
zoomequipd: the sender_domains list is the full domain, not root domain
```suggestion
        any(headers.reply_to, .email.domain.domain in $sender_domains)
```

detection-rules/spam_onmicrosoft_applications.yml
```diff
@@ -0,0 +1,63 @@
...
+        // solicited sender
+        and any(headers.reply_to, .email.domain.root_domain in $recipient_domains)
```
zoomequipd:
```suggestion
        and any(headers.reply_to, .email.domain.domain in $recipient_domains)
```

detection-rules/spam_onmicrosoft_applications.yml
```diff
@@ -0,0 +1,63 @@
...
+      )
```
zoomequipd: good use of `beta.profile.by_reply_to()`

detection-rules/spam_onmicrosoft_applications.yml
zoomequipd: We're baking an assumption into how subdomains of `onmicrosoft.com` relate to the orgainzation's (the sendering org) email domains.  they are def not a 1:1 mapping.  My experience tells me this assumption is going to be wrong in many cases.

zoomequipd (changes requested):

detection-rules/spam_onmicrosoft_applications.yml
missingn0pe: This was an approach to ensuring they were legitimately sent via the MSFT infra. Due to high volatility of authentication issues across a lot of samples, I thought it was a good idea to keep this in as this rule is designed to flag legitimate invites, not considering impersonations that would fall back on the 2 impersonation rules that were modified.

detection-rules/spam_onmicrosoft_applications.yml
missingn0pe: This was ripped from all the other rules that were updated, and I didn't test this logic. If no noise is present, happy to remove it!

detection-rules/spam_onmicrosoft_applications.yml
missingn0pe: Good call out. I pulled this logic from the current onmicro spam rule, so that will need to be updated as well. We could also account for a handful of other TLD's as well if needed (.net, .org, .gov, .edu etc)? Although, I observed some of these custom subdomains aren't 1:1 to the sender domains (which is exactly what happened here and was expressed to the SE when reporting back). This was added as a preliminary effort of FP control based on observed approaches that are live. Happy to discuss a better alternative for this approach as leven was originally considered, but was too volatile in distance.

detection-rules/spam_onmicrosoft_applications.yml
missingn0pe: Ah, my mistake here. Will update!

detection-rules/spam_onmicrosoft_applications.yml
missingn0pe: My approach to this was to try *not* to use profiles, and this appeared to be a good alternative strategy.

missingn0pe:
After review & consideration of telemetry, we cannot prove there is an attack present behind these automated emails. Given the noise it may produce in client environments, we're in favor of not approving this rule. However, this rule will be installed locally & monitored for any reported impact of spam/malice. In the event we can prove an attack is present, we can install this rule to core feed.

- [Final hunt](https://platform.sublime.security/messages/hunt?huntId=01a09129-2ada-721a-9cae-b1b4a7b8dfd3)

PR #5008, opened 2026-08-03
Update link_microsoft_low_reputation.yml - Sender negations

missingn0pe:
# Description

Adding another rule into the negation maelstrom for #4990. I thought I had them all but this makes 5.

# Associated samples

[- Sample 1](https://platform.sublime.security/messages/50a2db381eddefa8f5ca99a62ac6f9d740efb451d3f2ac458a3add6f25c4771b?preview_id=019f65e3-ab29-78ad-9f67-b9ff395a8854)

## Associated hunts

[- Hunt 1 (Shared samples) - L180D](https://platform.sublime.security/messages/hunt?huntId=019fc959-e915-770b-b227-02ec7127b1a3)
[- Hunt 2 (Multi) - L30D](https://hunt.limeseed.email/hunts/cd0c1722-ce0c-4be5-a5ec-8093e2b7794e)
[- Hunt 3 - Exclusive](https://platform.sublime.security/messages/hunt?huntId=019fc958-8e58-74be-8c67-a8d4dfdd65e0)

### DIFF
--- a/detection-rules/link_microsoft_low_reputation.yml
+++ b/detection-rules/link_microsoft_low_reputation.yml
@@ -385,6 +385,17 @@ source: |
    and subject.subject == strings.replace_confusables(subject.subject)
  )
  
+ // negate legitimate microsoft b2b applications invitations
+ and not (
+   length(body.links) > 0
+   and (
+     sender.email.local_part == "invites"
+     and sender.email.domain.root_domain == "onmicrosoft.com"
+     // infra validated message id
+     and strings.icontains(headers.message_id, "pepf")
+   )
+ )
+ 
  // negate org domains unless they fail DMARC authentication
  and (
    (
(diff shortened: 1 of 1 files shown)


detection-rules/link_microsoft_low_reputation.yml
```diff
@@ -385,6 +385,17 @@ source: |
    and subject.subject == strings.replace_confusables(subject.subject)
  )
  
+ // negate legitimate microsoft b2b applications invitations
+ and not (
+   length(body.links) > 0
```
IndiaAce: This seems like a redundant check to line 2:
if 
`and 0 < length(body.links) < 50` 
then
`length(body.links) > 0`
should never be false we can drop this to clean it up a bit, other than that I'm aligned to approve.

IndiaAce (changes requested):

detection-rules/link_microsoft_low_reputation.yml
missingn0pe: Valid point. I copied this stanza over from the other 4 rules & should have removed the link check for redundancy. Absorbing and sending it!

IndiaAce (approved):

PR #5019, opened 2026-08-04
Optimize three more rules slowing down one env

thunderkatz:
# Description
* `body_html_class_name_placeholder_rcpt.yml` gets a similar `{}` pre-filter like the one added in #5006. This is on HTML node attributes, not emails, but the same idea applies.
* `headers_anonymousfox.yml` and `impersonation_google_groups_suspicious.yml` have their `regex.icontains` turned into plain `strings.istarts_with` for header matching and `strings.icontains` for the rest.

# Associated samples
n/a

## Associated hunts
n/a

# Screenshot (insights)
n/a

### DIFF
--- a/detection-rules/headers_anonymousfox.yml
+++ b/detection-rules/headers_anonymousfox.yml
@@ -12,14 +12,10 @@ source: |
     any(headers.hops,
         any(.fields,
             strings.istarts_with(.name, "X-Authenticated-Sender", "X-Sender")
-            and (
-              strings.icontains(.value, "anonymousfox-")
-              or strings.icontains(.value, "smtpfox-")
-            )
+            and strings.icontains(.value, "anonymousfox-", "smtpfox-")
         )
     )
-    or strings.icontains(sender.email.email, "anonymousfox-")
-    or strings.icontains(sender.email.email, "smtpfox-")
+    or strings.icontains(sender.email.email, "anonymousfox-", "smtpfox-")
   )
 attack_types:
   - "BEC/Fraud"
--- a/detection-rules/headers_anonymousfox.yml
+++ b/detection-rules/headers_anonymousfox.yml
@@ -11,11 +11,14 @@ source: |
   and (
     any(headers.hops,
         any(.fields,
-            regex.icontains(.name, "^(X-Authenticated-Sender|X-Sender)")
-            and regex.icontains(.value, "(anonymous|smtp)fox-")
+            (strings.icontains(.name, "X-Authenticated-Sender")
+             or strings.icontains(.name, "X-Sender"))
+            and (strings.icontains(.value, "anonymousfox-")
+                 or strings.icontains(.value, "smtpfox-"))
         )
     )
-    or regex.icontains(sender.email.email, "(anonymous|smtp)fox-")
+    or strings.icontains(sender.email.email, "anonymousfox-")
+    or strings.icontains(sender.email.email, "smtpfox-")
   )
 attack_types:
   - "BEC/Fraud"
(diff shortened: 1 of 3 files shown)


detection-rules/headers_anonymousfox.yml
```diff
@@ -11,11 +11,18 @@ source: |
   and (
     any(headers.hops,
         any(.fields,
-            regex.icontains(.name, "^(X-Authenticated-Sender|X-Sender)")
-            and regex.icontains(.value, "(anonymous|smtp)fox-")
+            (
+              strings.istarts_with(.name, "X-Authenticated-Sender")
+              or strings.istarts_with(.name, "X-Sender")
+            )
```
hugh-sublime: `strings.istarts_with` can take in multiple arguments.

detection-rules/headers_anonymousfox.yml
```diff
@@ -11,11 +11,18 @@ source: |
...
+            and (
+              strings.icontains(.value, "anonymousfox-")
+              or strings.icontains(.value, "smtpfox-")
```
hugh-sublime: The new regex would actually have to do less work if you left it as is. Same with below.

detection-rules/impersonation_google_groups_suspicious.yml
```diff
@@ -16,9 +16,9 @@ source: |
   )
   and any(headers.hops,
           any(.fields,
-              regex.icontains(.name,
-                              "X-Authenticated-Sender|X-Sender|X-Original-Sender"
-              )
+              strings.istarts_with(.name, "X-Authenticated-Sender")
+              or strings.istarts_with(.name, "X-Sender")
+              or strings.istarts_with(.name, "X-Original-Sender")
```
hugh-sublime: You can combine this into a single `strings.istarts_with`.

detection-rules/headers_anonymousfox.yml
rw-access: These should also be consolidated into a single `icontains` call, it takes multiple args

detection-rules/headers_anonymousfox.yml
```diff
@@ -11,11 +11,15 @@ source: |
...
+            strings.istarts_with(.name, "X-Authenticated-Sender", "X-Sender")
+            and (
+              strings.icontains(.value, "anonymousfox-")
+              or strings.icontains(.value, "smtpfox-")
+            )
         )
     )
-    or regex.icontains(sender.email.email, "(anonymous|smtp)fox-")
+    or strings.icontains(sender.email.email, "anonymousfox-")
+    or strings.icontains(sender.email.email, "smtpfox-")
```
rw-access: should also be consolidated

detection-rules/impersonation_google_groups_suspicious.yml
thunderkatz: Done

detection-rules/headers_anonymousfox.yml
thunderkatz: Sorry, are you saying regexlite would do less work than the two contains?

detection-rules/headers_anonymousfox.yml
thunderkatz: Done.

detection-rules/headers_anonymousfox.yml
thunderkatz: Done.

detection-rules/headers_anonymousfox.yml
hugh-sublime: Ultimately, `regexlite` will do suffix consolidation on `fox-` but you could save that work by leaving it as it was. Not a big deal though either way. It will just be strictly less performant on compile once we're onto `regexlite`.

Same could be said for the other regex to string contains conversions.

thunderkatz: Sure, reverted to regex (those aren't showing up in the dashboard at the moment anyway).

zoomequipd (approved):

PR #5022, opened 2026-08-04
Fix null handling for nlu_classifier language checks

mpasquale-3:
# Description

Wrap `ml.nlu_classifier(body.current_thread.text).language` comparisons with `coalesce(..., true)` in two rules so they don't fall back to `null` when `body.current_thread.text` is `null` and the classifier returns no output. Using `coalesce` will make these rules behave the same as if `unknown` was returned by `ml.nlu_classifier`.

Rules updated:
- `credential_phishing_nifty.com_domain_abuse.yml` — `!= "japanese"` check
- `credential_phishing_reauthentication.yml` — `!= "english"` check


mpasquale-3:
This is not needed. The `nlu_classifer` returns `"unknown"`

PR #5039, opened 2026-08-06
Update link_content_credential_phishing.yml

markmsublime:
# Description
update regex to match literal dot, add non-capturing group, and update high trust negation to latest and greatest 

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=019fd79f-3bb9-7b52-b76b-c6f99f3c45a9) - showing no cases where the any character regex is needed

### DIFF
--- a/detection-rules/link_content_credential_phishing.yml
+++ b/detection-rules/link_content_credential_phishing.yml
@@ -69,16 +69,13 @@ source: |
   
   // negate docusign originated emails
   and not any(headers.hops,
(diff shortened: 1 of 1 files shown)


rw-access (approved):
Thanks! Also FYI - you don't have to sweat capturing vs non-capturing groups, MQL removes captures under the hood automatically anyway and tries a ton of regex optimizations, especially the low hanging fruit ones.

zoomequipd (approved):

PR #5040, opened 2026-08-06
Use regex.imatch for certain URL regexes

hugh-sublime:
# Description

These URL regexes should be applied to the entire URL instead of seeing if the URL contains this regex. This does have some performance improvements and is probably closer to the original intention of the rule.

### DIFF
--- a/detection-rules/link_numeric_ip_obfuscation.yml
+++ b/detection-rules/link_numeric_ip_obfuscation.yml
@@ -4,7 +4,7 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  and any(body.links, regex.icontains(.href_url.url, 'https?://[0-9]{7,12}/.+'))
+  and any(body.links, regex.imatch(.href_url.url, 'https?://[0-9]{7,12}/.+'))
 attack_types:
   - "Credential Phishing"
   - "Malware/Ransomware"
--- a/detection-rules/link_numeric_ip_obfuscation.yml
+++ b/detection-rules/link_numeric_ip_obfuscation.yml
@@ -4,7 +4,11 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  and any(body.links, regex.imatch(.href_url.url, 'https?://[0-9]{7,12}/.+'))
+  and any(body.links,
+          any(.href_url.ip.translation.encoders,
+              . in ('octal', 'decimal_integer')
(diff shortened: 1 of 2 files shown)


zoomequipd:
regression check hunts:
https://platform.sublime.security/messages/hunt?huntId=019fd815-7bf1-7886-af38-c92e89de2477
https://platform.sublime.security/messages/hunt?huntId=019fd816-db24-78ff-87fb-c38d056d015a

detection-rules/link_numeric_ip_obfuscation.yml
```diff
@@ -4,7 +4,7 @@ type: "rule"
 severity: "medium"
 source: |
   type.inbound
-  and any(body.links, regex.icontains(.href_url.url, 'https?://[0-9]{7,12}/.+'))
+  and any(body.links, regex.imatch(.href_url.url, 'https?://[0-9]{7,12}/.+'))
```
zoomequipd: ~i think we can do this now.~
confirmed, notes below are the path to confirmation
```suggestion 
  and any(body.links,
          any(.href_url.ip.translation.encoders,
              . in ('octal', 'decimal_integer')
          )
  )
```

zoomequipd (changes requested):

detection-rules/link_numeric_ip_obfuscation.yml
zoomequipd: regression hunt on octal encoding detection:
https://platform.sublime.security/messages/hunt?huntId=019fd819-f588-73d7-a049-b0c68f05e7e8

zoomequipd: regex matches decial_int as well
sample: https://platform.sublime.security/messages/50bf2e8ac91ca7a687f429df9a10fd05ea7c1cccdaf46fbfa2a30766d4dbd810?preview_id=019fd77a-aef8-7b87-873c-eb6f0b4fbe48

zoomequipd: regression check on octal OR decimal_int
https://platform.sublime.security/messages/hunt?huntId=019fd81e-5fa4-79cb-b5d9-f23c283d579e
(no backfill completed on the ip translation, so matches past feature are expected, i think we're good on dropping the regex at this point)

zoomequipd (approved):

PR #5074, opened 2026-08-10
Update: link_credential_phishing_secure_message

peterdj45:
# Description

adding `share` as an additional suspicious keyword

# Associated samples
- https://platform.sublime.security/messages/50b9592de249bb5fb5b8c03f0bc859cdc827a94074a01e5d91e0d3f053902d17

## Associated hunts
-


IndiaAce:

PR #5104, opened 2026-08-12
Fix BOM in YARA test fixture zip

yungbeaniebaby:
# Description

The `test.txt` inside the old `sublime_yara_test.zip` (linked from docs.sublime.security/docs/yara's "Testing YARA in Sublime" section) had a UTF-8 BOM before the string `Sublime-Standard-Test-String`. The `SublimeStandardTestString` rule's condition is `$test_string at 0`, so the BOM caused the doc's own sample fixture to fail its own sample rule for everyone following the guide.

This PR adds a corrected `sublime_yara_test.zip` under `yara/` (no BOM, string starts at byte offset 0, verified via hexdump). It's a static test asset, not a detection rule change.

Once merged, `docs.sublime.security/docs/yara` should be updated to link to this file's raw GitHub URL instead of the old GitHub file attachment.

# Associated samples

N/A — this change replaces a benign test fixture, not a detection rule.

## Associated hunts

N/A


zoomequipd:
diff visually here 
image

zoomequipd (approved):

PR #5106, opened 2026-08-12
Create adaptive_phishing_simulation.yml

zoomequipd:
# Description

New default global exclusion for adaptive security which covers both SMTP and DMI delivery methods. 

## Associated hunts

- [Hunt 1](https://hunt.limeseed.email/hunts/79ebcd8b-98a5-442e-a209-f1a2e1c1ef90)

### DIFF
--- a/exclusions/adaptive_phishing_simulation.yml
+++ b/exclusions/adaptive_phishing_simulation.yml
@@ -1,0 +1,31 @@
+name: "Adaptive Security phishing simulation"
+description: "Identifies phishing simulations sent by Adaptive and excludes the message from live analysis."
+type: "exclusion"
+source: |
+  type.inbound
+  and (
+    // smtp delivery
+    any(headers.ips,
+        //
+        // This rule makes use of a beta feature and is subject to change without notice
+        // using the beta feature in custom rules is not suggested until it has been formally released
+        //
+        beta.ip_in(.ip,
+                   '204.220.178.8',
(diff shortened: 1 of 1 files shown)


exclusions/adaptive_phishing_simulation.yml
```diff
@@ -0,0 +1,31 @@
...
+        //
+        beta.ip_in(.ip,
+                   '204.220.178.8',
+                   '204.220.178.36',
+                   '204.220.178.57',
+                   '131.162.232.0/22',
+        )
+    )
+    // dmi delivery
+    or length(headers.hops) <= 1
```
IndiaAce, line 21: are there instances where these known IPs have been false but the length of headers have been 1 or 0? I guess my only concern here is how expansive this *feels* as an or to checking for the 4 above IPs. If we do have results, could we grab those found IPs, confirm they're adaptives, and then just add them to the `beta.ip_in` check?

IndiaAce (changes requested):

IndiaAce (approved):

exclusions/adaptive_phishing_simulation.yml
randypargman: In the caes of DMI I looked at, there were NO IP ADDRESSES AT ALL in the hops - just directly inserted the message with no SMTP delivery and thus no IP to look at

PR #5112, opened 2026-08-13
disable spamhaus rules real quick

keaton-sublime:
## Description
Disabling the spamhaus rules.


keaton-sublime:
we've decided to manage this in a different way.

PR #5123, opened 2026-08-17
Update attachment_pdf_link_sus_lang.yml

zoomequipd:
# Description

Handle single-page PDFs with an invalid xref table.

Some malicious single-page PDFs ship with an invalid xref table, causing `beta.parse_exif(.).page_count` to return `0` instead of `1`, so single-page-crafted phish were slipping past the `page_count == 1` gate. We now also accept a PDF when `page_count == 0`, exiftool emits a `Warning` field with value `Invalid xref table`, and the PDF scanner rendered exactly one page (`pdf_2_png_*` exploded file count == 1). This preserves the "single page" intent using the scanner's render as the fallback source of truth.

Same fallback as sublime-security/sublime-rules#5122, applied to this rule.

# Associated samples

- [Sample 1](https://platform.sublime.security/messages/50bc69373f42295b3012519edd99623abbdaef2261c8f16411f71a94f831ad72?preview_id=019fd8e6-ad5c-7d7b-a112-6d34c15ac180)

## Associated hunts

- [Net-net multihunt](https://hunt.limeseed.email/hunts/ed227ff3-4934-443d-a19f-93ed9717563c)
- [Net-new hunt](https://platform.sublime.security/messages/hunt?huntId=01a0105c-f11d-744d-92bb-15d49c6e9539)

(claude wrote this desc, i made the rule change)


zoomequipd:
marked for review needed after bumping tests now that Github is working again

markmsublime (approved):

PR #5127, opened 2026-08-18
Update generic_zip.yar

keaton-sublime:
## Description
yara rule to match zip files which contain different file names in the same-index local file header and central directory records. Could be an indicator of modified or malformed files. 

## Associated samples

- [Sample 1](https://platform.sublime.security/messages/50ca0e589d58e21ad4531735463b2e9d144dde21e8b471f8e4d4463a59a73da9)


keaton-sublime:
This is a yara rule with no matching MQL. Marking ready for review/review needed so we can begin testing. 

My initial thoughts is that this will probably end up as an ASR rule.

markmsublime (approved):

PR #5132, opened 2026-08-19
Update generic_zip.yar

keaton-sublime:
## Description
Expanding the logic of this to include an observed win word binary as well as the excel binary. 

## Associated Samples
- [Sample](https://platform.sublime.security/messages/50cb6e645191291f663e92d613e89712b2daf46e377a727b262219838504b1b2?preview_id=01a013eb-af95-7929-bb54-b7952eea82e5)

### DIFF
--- a/yara/generic_zip.yar
+++ b/yara/generic_zip.yar
@@ -41,26 +41,30 @@ rule zip_smuggler_default {
 		)
 }
 
-rule zip_excel_dll {
+rule zip_office_bin_dll {
 	meta:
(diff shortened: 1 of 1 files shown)


keaton-sublime:
Updating the existing yara (including a name change) to expand the logic. The threat is the same as the previous version, but they are also using word instead of just excel. I've expanded that. 

Marking as ready for review/review needed - there is no matching MQL for this.

markmsublime (approved):

PR #5138, opened 2026-08-20
Add test rules for FAS spam/malicious verdicts

dlynch-sublime:
# Description
Adds two new test rules, similar to the core graymail rule:
* fuzzy_malicious_rule.yml defines a rule flagging on malicious verdicts from fuzzy attack score
* fuzzy_spam_rule.yml defines a rule flagging on spam verdicts from fuzzy attack score

### DIFF
--- a/detection-rules/fuzzy_malicious_rule.yml
+++ b/detection-rules/fuzzy_malicious_rule.yml
@@ -1,0 +1,12 @@
+name: "Fuzzy Attack Score: Malicious Detection"
+description: "Exercises the beta.fuzzy_attack_score() function. Flags on Malicious verdicts."
+type: "rule"
+severity: "medium"
+source: |
+  type.inbound
+  //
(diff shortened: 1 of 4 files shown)


detection-rules/fuzzy_malicious_rule.yml
```diff
@@ -0,0 +1,12 @@
+name: "Fuzzy Attack Score: Malicious Detection"
```
dlynch-sublime, line 1: If this needs to change in test, provide an update

detection-rules/fuzzy_spam_rule.yml
```diff
@@ -0,0 +1,12 @@
+name: "Fuzzy Attack Score: Advanced Spam Detection"
```
dlynch-sublime, line 1: If this needs to change in test, provide an update

PR #5146, opened 2026-08-21
Update generic_pdf.yar

keaton-sublime:
## Description
Matching PDFs from a known DOUBLOON DREDGER template.


keaton-sublime:
This is a yara rule with no matching MQL. Marking ready for review/review needed.

markmsublime (approved):

PR #5204, opened 2026-08-27
Remove noisy and redundant DLP rules

rw-access:
# Description

This removes some DLP rules that are either legacy, out of scope, noisy, or need multiple indicators to turn into high signal detections.

There was also one that needed to be migrated to `extract_sensitive_information`

# Associated samples
N/A

## Associated hunts
N/A

# Screenshot (insights)
N/A


Jummyberry (approved):
Ran hunts on all deleted DLP rules in shared samples over 30 days. Removals of noisy and redundant dlp rules are justified. Also reviewed surgical procedure dropped and Spain Passport added as a new rule. Approved

Jummyberry (approved):
Ran hunts on all deleted DLP rules in shared samples over 30 days. Removals of noisy and redundant dlp rules are justified. Also reviewed surgical procedure dropped and Spain Passport added as a new rule. Approved

PR #5211, opened 2026-08-28
Fix rule validation diagnostics: deprecated identifiers, regex hints, logic ambiguities

aidenmitchell:
## Summary

Resolves nearly all `warning`/`hint`-severity diagnostics found by validating every rule file against `https://play.sublime.security/v1/rules/bulk_validate` (see #5209 for the CI annotation work that surfaced these). No `error`-severity diagnostics existed before or after this PR -- all rules validated cleanly throughout.

Diagnostic count across the whole repo: **311 → 74** (all 74 remaining are either explicitly deferred below, or a noisy hint CI already filters out -- see "Not fixed").

### Mechanical fixes (5 commits)
- **Deprecated identifiers** (39 files): `beta.ml_topic`→`ml.nlu_classifier`, `beta.scan_base64`→`strings.scan_base64`, `any_false_positives`→`any_messages_benign`, `beta.linkanalysis`→`ml.link_analysis`, `beta.message_screenshot`→`file.message_screenshot`, bare `ilike(`→`strings.ilike(`.
- **Regex anchoring** (49 files, 66 call sites): `regex.icontains`/`regex.contains` → `regex.imatch`/`regex.match` at call sites where the pattern is already start-anchored, per the validator's own suggestion. **Correction:** the initial conversion missed that `match`/`imatch` require the pattern to account for the whole string, while `icontains`/`contains` only anchor what's given and allow anything to follow. A follow-up commit appends a trailing `.*` to the 52 (of 66) converted patterns that didn't already constrain the end of the string, to preserve the original matching behavior. Every one of the 41 affected files was re-validated after the correction.
- **Loop-invariant hoisting** (17 files, 23 sites): moved loop-invariant sub-expressions out of `any()`/`all()` calls for performance; used the exact `length(items) > 0 and invariant` rewrite where naive hoisting would mishandle the empty-collection edge case.
- **Duplicate set values** (2 files, 34 removed): de-duplicated literal list entries, keeping first occurrence; spot-checked removed values weren't near-duplicate typos.

Every touched file was re-validated individually against the endpoint after editing (no new errors, target diagnostic gone) before being committed.

### Logic-error fixes, human-reviewed (1 commit, 5 files)
The validator flags ambiguous unparenthesized `and`/`or` mixing. These aren't style nits -- 4 of 5 were real detection gaps:
- `impersonation_microsoft_via_html_table_quarantine.yml`: sender-domain exclusion and quarantine-keyword check would only gate the *last* OR-branch of the logo-detection chain, not all of them -- risked false positives on genuine Microsoft/Office/SharePoint mail.
- `link_coinbase_low_rep_or_shortened.yml`: the Google Maps exemption only applied to one of the two membership checks it was meant to guard.
- `impersonation_github_sawfish.yml`: the lookalike-domain-link path would no longer be scoped by `type.inbound` or the recipient-exclusion guard.
- `attachment_eml_svg_js.yml`: would match any `.svg`-extension attachment regardless of content, bypassing the malicious-JS check.
- `link_download_suspicious_file.yml`: lower-confidence/clarity-only change; behavior was likely already correct, parenthesized to make it explicit.

I reviewed each proposed fix against the rule's description/comments before applying, and re-validated afterward.

### Not fixed -- needs a human/vendor decision
- **`attachment_ics_file_non_rfc_compliant.yml`**: `VFREEBUSY` isn't a valid ICS component-type value, but the rule's own description explicitly names it as intended (`VTODO, VJOURNAL, VFREEBUSY, or VEVENT`). Valid neighbors are `VEVENT`/`VJOURNAL`/`VTODO`/`VALARM`/`VTIMEZONE`. Left as-is per request.
- **`attachment_malicious_onenote_commands.yml`**: `"onenote_file" in .flavors.yara` isn't a valid YARA flavor (checked `onenote`, `onestore`, `msonenote`, `one_file`, `cfbf_file`, `ole_file`, `msg_file`, and others -- none valid). This looks like the vendor's classifier has no OneNote flavor, so this branch of the rule likely never matches. Left as-is per request.
- **`beta_dlp_oauth_client_secret.yml`** and **`beta_dlp_clear_text_credentials_outbound.yml`** (8 occurrences): `oauth_client_secret` isn't a valid `beta.ml_extract_sensitive_information` type. Tried ~15 plausible alternatives (`oauth_secret`, `client_secret`, `oauth_token`, `api_key`, `access_token`, etc.) -- none are valid either. This looks like a real gap in the vendor's ML classifier, not a naming issue on our side. A separate, already-working `dlp_oauth_client_secret.yml` covers the same detection via regex instead. Left as-is per request; needs vendor/engineering follow-up.

### Also filtered, not a fix
64 remaining `hint`s are all "This field may be null for MDM retrieval or backtests/hunts..." -- already suppressed from CI annotations in #5209 as non-actionable noise.

### Also included: unrelated CI fix
This PR's own CI run surfaced a pre-existing bug in `.github/workflows/rule-validate.yml`, unrelated to rule content: `git add -- **/*.yml` is unquoted, so when the shell has no `globstar` (the Actions runner's default), `**` doesn't recurse and files nested two+ levels deep (`insights/<category>/*.yml`) silently never get staged. When the auto-formatter only touches `insights/` files, the commit step fails outright. Fixed by quoting the pathspec so git's own glob engine handles it (also applied to #5209, since `pull_request_target` always runs the workflow YAML from `main`, not the PR branch -- the fix only takes effect once it's on `main`).

## Test plan
- [x] Every file touched by the mechanical-fix commits was re-validated individually after editing (no new errors, target diagnostic resolved)
- [x] Every logic-error fix re-validated after editing
- [x] Full-repo re-scan (all 1587 files) before opening this PR: 0 errors, 74 remaining diagnostics -- all accounted for above
- [x] Regex trailing-wildcard correction: all 41 affected files re-validated, 0 errors
- [ ] CI run on this PR itself (blocked until the CI fix lands on `main` via #5209 -- see above)


detection-rules/abuse_cloudflare_workers_hosted_eviltokens_domain_structure.yml
```diff
@@ -10,8 +10,8 @@ source: |
           // eviltokens cloudflare workers domain struct
           (
             strings.iends_with(.href_url.domain.domain, '-account.workers.dev')
-            and regex.icontains(.href_url.domain.domain,
-                                '^(?:(?:page-)?adobe|calendar_invite|(?:page-)?docusign|fax|quarantine|onedrive|page-password|sharepoint|voicemail|index)-[a-z0-9]{3}\.[a-z0-9-]{3,}'
+            and regex.imatch(.href_url.domain.domain,
+                             '^(?:(?:page-)?adobe|calendar_invite|(?:page-)?docusign|fax|quarantine|onedrive|page-password|sharepoint|voicemail|index)-[a-z0-9]{3}\.[a-z0-9-]{3,}'
```
rw-access: Not quite right, if this switches to `imatch` it needs trailing wildcards. Or you can switch `^` to `\A` and keep it with contains

PR #5210, opened 2026-08-28
Create link_fake_video_new_domain.yml

keaton-sublime:
## Description
Matches messages with a recently registered domain sent from a fake "support" address, where the link that references a hosted video.

## Associated samples
- [Sample 1](https://platform.sublime.security/messages/50d122e11ba6963a55adbb3cf5c33fa555dcc70667abd2a3feed0b84a7b2e3af?preview_id=01a0447b-c090-72a2-9690-55ec4e44cf1a)
- [Sample 2](https://platform.sublime.security/messages/50cb6e645191291f663e92d613e89712b2daf46e377a727b262219838504b1b2?preview_id=01a013eb-af95-7929-bb54-b7952eea82e5)

## Associated hunts

- [Hunt 1](https://platform.sublime.security/messages/hunt?huntId=01a04970-fdc4-734d-b160-968e4d173a03)
- [Multi-hunt](https://hunt.limeseed.email/hunts/27294ac3-a2fa-4cd1-9c67-cebbf469c48e)
- [updated multi-hunt (aug 31)](https://hunt.limeseed.email/hunts/656b7e48-20e8-4972-8c70-354d869dd836)

PR #5222, opened 2026-08-31
Update link_fake_video_new_domain.yml

keaton-sublime:
## Description
The [non-blocking feedback](https://github.com/sublime-security/sublime-rules/pull/5210) was to try this without the check for local sender, so I figured I'd throw that into test rules and let it bake.


keaton-sublime:
Marking ready for review/review needed. This was a tip up from Bry. Looks like the activity has died off based on the telem, but last 3 days in test rules haven't shown any FPs.

zoomequipd (approved):

PR #5285, opened 2026-09-04
Remove old regex-style Slack token DLP rule

randypargman:
## Summary
- Removes `dlp_slack_token.yml`, the old regex-based Slack token detection rule
- The new recognizer-based rule for Slack access tokens supersedes this

## Test plan
- [ ] Confirm new Slack access token recognizer rule is active and covers the same detection surface


Jummyberry (approved):
Verified that beta_dlp_slack_access_token (ML recognizer-based rule) is active and covers the same detection surface as the old regex rule. Ran the old regex pattern as a hunt against shared samples 0 hits, confirming no active detection is being lost with this removal. Approving

rw-access (approved):

Jummyberry (approved):

PR #5322, opened 2026-09-10
Update attachment filtering in Docusign rule

randypargman:
Replace strict zero attachments logic with "zero attachments but ignore image warning banners" logic.

# Description

3rd party tools adding warning banners as images rather than text mess with rules that have "zero attachments" logic by adding an attached image warning banner that wasn't part of the original message.  This uses a new list of warning banner image SHA256 hashes to exclude those.


markmsublime:
shared results from test rules, net new hunt showing no telemetry - [hunt](https://platform.sublime.security/messages/hunt?huntId=01a0a642-36eb-7980-abaa-2e8126328611)

PR #5324, opened 2026-09-10
Update abuse_fliphtml5 rule replace zero attachment logic

randypargman:
# Description


markmsublime:
no results in test rules to share, net new hunt for posterity - [hunt](https://platform.sublime.security/messages/hunt?huntId=01a0a640-8d85-7d5c-8d91-281cd3eb03b9)

PR #5325, opened 2026-09-10
Update BEC rule to match intent of no attachments other than banners

randypargman:
Replace strict zero attachments logic with "zero attachments but ignore image warning banners" logic.

# Description

In a previous PR, @markmsublime  changed the logic from length(attachments) == 0 (which was causing FN) to checking for zero attachments other than images. This PR further refines that to only excluding known email warning banners.


markmsublime:
shared results from test rules, net new hunt showing no telemetry - [hunt](https://platform.sublime.security/messages/hunt?huntId=01a0a63f-dba4-7c07-b2aa-ced8911bacc6)

PR #5326, opened 2026-09-10
Refine attachment checks in mobile solicitation rule

randypargman:
Update attachment filtering logic to exclude known 3rd party email warning banners.

# Description

In a previous PR, @markmsublime changed it from length(attachments) == 0 (which was causing FN) to checking for zero attachments other than images. This PR further refines that to only excluding known email warning banners.


markmsublime:
shared results from test rules, net new hunt showing no telemetry - [hunt](https://platform.sublime.security/messages/hunt?huntId=01a0a63d-1b3b-7ac8-bc62-c886b90c6f19)

PR #5332, opened 2026-09-11
Rename Fuzzy Attack Score on Graymail

dlynch-sublime:
# Description
Renames the fuzzy attack score graymail rule to use the new function name over the deprecated function name.


markmsublime (approved):
lgtm!

PR #5335, opened 2026-09-11
Update uk dlp rules meta only

randypargman:
# Description

Changed metadata of all UK specific rules to be consistent with attack_types and tags of other DLP rules. 
- Replaced "DLP: PII" with "DLP: Personal data" in attack_types
- Moved "DLP: GDPR" from attack_types to "UK GDPR" in tags
- Added "UK" to tags

No change to any rule logic was made, metadata only.


randypargman:
I also changed one filename from `beta_dlp_nhs_number.yml` to `beta_dlp_uk_nhs_number.yml` for file naming consistency with other rules

Jummyberry (approved):
Approved

PR #5336, opened 2026-09-11
Added country and EU tag to each EU member state rule

randypargman:
# Description

Changed metadata only, for EU member state specific rules only.
- Added tag "EU" to each
- For those that lacked a country tag, added the country name consistently


Jummyberry (approved):
Approved

PR #5404, opened 2026-09-18
Metadata only: Consolidate Financial, Consolidate Healthcare

randypargman:
# Description

Changed metadata of DLP rules only:
We had tags for "Healthcare" and for "Healthcare data" -> Consolidated both to "Healthcare" tag
We had attack_types for "DLP: Healthcare" and "DLP: Healthcare Data" -> Consolidated both to "DLP: Healthcare"
Those that also had "DLP: HIPAA" kept that attack type in addition to "DLP: Healthcare" so they can be found both ways.

We had tags for "Financial" and "Financial data" -> Consolidated both to "Financial" tag
We had attack_types for "DLP: Financial" and "DLP: Financial Data" -> Consolidated both to "DLP: Financial"

For consistency in capitalization, changed every instance of "Personal data" to "Personal Data" across the board.

No changes to any detection logic were made. 

Removed one reference to "DLP: PII" and replaced with "DLP: Personal Data" for consistency.


Jummyberry (approved):
The commit title mentions only Financial and Healthcare but also includes Personal Data Capitalization aside that approving this PR

